PR #1, opened 2019-06-06
Solution by Kate

ekaterinakur:
[Preview] (https://ekaterinakur.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList.js
+++ b/src/components/TodoList.js
@@ -1,0 +1,84 @@
+import React from 'react';
+import TodoItem from './TodoItem';
+import './TodoList.css';
+
+class TodoList extends React.Component {
+  constructor(props) {
+    super(props);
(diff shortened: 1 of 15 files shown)


src/components/TodoList.js
```diff
@@ -0,0 +1,84 @@
...
+      })
+    }))
+  }
+
+  render() {
+    if (!this.state.requested) {
+      return <button onClick={this.loadItems}>Load</button>;
+    } else if (this.state.loaded) {
+      return (
+        <div className="todos-list">
```
gestap: why you need this `div`?

src/users.js
```diff
@@ -0,0 +1,232 @@
+export const users = [
```
gestap: this file and file above from another task

PR #70, opened 2019-09-23
solution

ArnyCorp:
link: https://arnycorporation.github.io/react_dynamic-list-of-todos/


irenhh:
sorting is not optional

PR #79, opened 2019-09-24
Vyacheslav-P react_dynamic-list-of-todos solution

Vyacheslav-P:
* [link](https://vyacheslav-p.github.io/react_dynamic-list-of-todos/)


irenhh:
sorted list has to be the same for each option no matter in what order I click sorting buttons

src/App.js
```diff
@@ -1,15 +1,135 @@
...
+      sortedTodosList: [...prevState.todosList]
+        .sort((todo1, todo2) => todo1.user.name.localeCompare(todo2.user.name)),
+    }));
+  };
+
+  sortCompleted = () => {
+    this.setState(prevState => ({
+      sortedTodosList: [
+        ...prevState.todosList.filter(todo => todo.completed),
+        ...prevState.todosList.filter(todo => !todo.completed),
```
irenhh: instead of filtering twice create a variable, assign filtering result and then assign the value of this variable to the keys

irenhh: take this into consideration for other cases in this project

src/App.js
```diff
@@ -1,15 +1,135 @@
...
+      isLoading,
+      buttonText,
+      isError,
+    } = this.state;
+
+    if (!isLoaded) {
+      return (
+        <div>
+          {isError
+            ? <p>No Data :( Try again</p>
```
irenhh: instead of ternary operator use logical operator

src/App.js
Vyacheslav-P: we discussed this issue

PR #83, opened 2019-09-25
My dynamo-react #1?

Volodsher:
https://volodsher.github.io/react_dynamic-list-of-todos/

Thanks for your help! and also, could you please help me with todos PropTypes in TodoList, just don't have any clue what to do with it.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,15 +1,76 @@
 import React from 'react';
 import './App.css';
+import { TodoList } from './components/TodoList/TodoList';
 
-import todos from './api/todos';
-import users from './api/users';
+const BASE_URL = 'https://jsonplaceholder.typicode.com';
 
-function App() {
-  return (
-    <div className="App">
-      <h1>Dynamic list of todos</h1>
-    </div>
-  );
+class App extends React.Component {
+  state = {
+    todos: [],
+    isLoading: false,
+    isLoaded: false,
+    error: null,
+  }
+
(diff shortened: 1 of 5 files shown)


Volodsher:
Thanks, I cleaned everything up! at least, I believe so:)

src/App.js
```diff
@@ -1,15 +1,71 @@
...
+      ]);
+      const todos = await todosResponse.json();
+      const users = await usersResponse.json();
+
+      const todosWithUsers = todos.map(({ userId, ...rest }) => ({
+        ...rest,
+        user: users.find(({ id }) => id === userId),
+      }));
+
+      console.log(todosWithUsers);
```
irenhh: it's still here

src/components/TodoList/TodoList.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './TodoList.css';
+import { TodoItem } from '../TodoItem/TodoItem';
+
+export const TodoList = (props) => {
+  const { todos } = props;
+  console.log(typeof todos);
```
irenhh: the same

src/App.js
Volodsher: got it

src/components/TodoList/TodoList.js
Volodsher: nailed that as well

PR #183, opened 2020-05-18
solution

IgorKulish:
[DEMO LINK](https://igorkulish.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,8 +1,75 @@
...
+          };
+        });
+
+        setIsLoaded(true);
+        setTodos(preparedTodosList);
+      });
+  };
+
+  const sortTitles = () => {
+    setTodos([...todos]
```
irenhh, line 31: don't save sorted todos to the initial list, do it on rendering

src/App.tsx
```diff
@@ -7,6 +7,7 @@ import { Todos } from './interfaces';
 
 const App = () => {
   const [todos, setTodos] = useState<Todos[]>([]);
+  const [initialTodos, setInitialTodos] = useState<Todos[]>([]);
```
irenhh, line 10: do sorting on rendering, it's too expensive to store two lists if they are very big

src/App.tsx
```diff
@@ -25,39 +24,40 @@ const App = () => {
...
   };
 
-  const sortTitles = () => {
-    setTodos([...initialTodos]
-      .sort((a, b) => a.title.localeCompare(b.title)));
-  };
+  const sortingMethod = (sort: string) => {
+    switch (sort) {
+      case 'title':
+        return setTodos([...todos]
```
irenhh, line 33: you do the same again, don't set sorted todos in the state, just return a new array and call it after `render(...`

PR #197, opened 2020-05-20
solution

cutiekate:
[DEMO LINK](https://cutiekate.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,115 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.css';
 
-const App = () => (
-  <h1>Dynamic list of TODOs</h1>
-);
+import { getUsers, getTodos, Todo } from './helpers/api';
+
+const App = () => {
+  const [initialTodos, setInitialTodos] = useState<Todo[]>([]);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleLoadClick = async () => {
+    setIsLoading(true);
+
+    const todosFromServer = await getTodos();
+    const usersFromServer = await getUsers();
+
+    const todosWithUsers = todosFromServer.map(todo => ({
+      ...todo,
+      user: usersFromServer.find(user => user.id === todo.userId),
+    }));
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,8 +1,115 @@
...
+    const todosFromServer = await getTodos();
+    const usersFromServer = await getUsers();
+
+    const todosWithUsers = todosFromServer.map(todo => ({
+      ...todo,
+      user: usersFromServer.find(user => user.id === todo.userId),
+    }));
+
+    setInitialTodos(todosWithUsers);
+    setTodos(todosWithUsers);
```
irenhh, line 23: store only the initial array in the state, the other one should be calculated just before rendering

src/App.tsx
```diff
@@ -1,8 +1,115 @@
...
+                  {todo.user ? todo.user.name : 'Unknown'}
+                </span>
+              </li>
+            ))}
+          </ul>
+        </>
+      )}
+
+
+    </div>
```
irenhh, line 111: remove extra empty lines

src/helpers/api.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface Todo {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
+  user?: User;
+}
+
+const getAll = <T>(url: string): Promise<T[]> => {
```
irenhh, line 16: get all what?

src/App.tsx
```diff
@@ -1,8 +1,85 @@
...
+    setSortBy(1);
+  };
+
+  const sortByTitle = () => {
+    setSortBy(2);
+  };
+
+  const sortByCompleted = () => {
+    setSortBy(3);
+  };
```
irenhh, line 41: don't use numbers, it's not clear what number represents which action;
create one method for this, actions are almost the same

src/App.tsx
```diff
@@ -1,8 +1,85 @@
...
+
+  const todos = [...initialTodos].sort((a, b) => {
```
irenhh, line 43: try useMemo for this case so that your array will recalculate only if initialTodos are changed

PR #245, opened 2020-09-10
solution

furude-rika:
- [DEMO LINK](https://furude-rika.github.io/react_dynamic-list-of-todos/)
Implemented 1 - 9

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -3,26 +3,70 @@ import './App.scss';
 import './styles/general.scss';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
+import { getTodos, getUserInfo } from './api/api';
 
 class App extends React.Component {
   state = {
     todos: [],
     selectedUserId: 0,
+    title: '',
+    isCompleted: '',
   };
 
+  componentDidMount() {
+    getTodos()
+      .then((todos) => {
+        this.setState({ todos });
+      });
+  }
+
+  setUserId = (id) => {
+    this.setState({ selectedUserId: id });
+  }
+
+  clearUserId = () => {
+    this.setState({ selectedUserId: '' });
+  }
+
+  filteredByTitle = (subtitle) => {
+    this.setState({ title: subtitle });
+  }
+
+  filteredByStatus = (status) => {
+    this.setState({ isCompleted: status });
+  }
+
   render() {
-    const { todos, selectedUserId } = this.state;
+    const { todos, selectedUserId, title, isCompleted } = this.state;
+    let visibleTodos = todos.filter(todo => todo.title.includes(title));
+
+    if (isCompleted === 'completed') {
+      visibleTodos = visibleTodos.filter(todo => todo.completed === true);
+    }
+
+    if (isCompleted === 'active') {
+      visibleTodos = visibleTodos.filter(todo => todo.completed === false);
+    }
 
     return (
       <div className="App">
         <div className="App__sidebar">
-          <TodoList todos={todos} />
+          <TodoList
+            todos={visibleTodos}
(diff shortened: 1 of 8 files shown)


martakupka (approved):

src/App.jsx
```diff
@@ -3,26 +3,70 @@ import './App.scss';
...
+
+  setUserId = (id) => {
+    this.setState({ selectedUserId: id });
+  }
+
+  clearUserId = () => {
+    this.setState({ selectedUserId: '' });
+  }
+
+  filteredByTitle = (subtitle) => {
```
irenhh, line 31: this method doesn't filter

src/App.jsx
```diff
@@ -3,26 +3,70 @@ import './App.scss';
...
+    this.setState({ title: subtitle });
+  }
+
+  filteredByStatus = (status) => {
```
irenhh, line 35: the same

src/api/api.js
```diff
@@ -0,0 +1,11 @@
+const API_URL = 'https://mate-api.herokuapp.com/';
+// https://mate-api.herokuapp.com/users/1
```
irenhh, line 2: ?

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,93 @@
...
-
-          <button
-            className="TodoList__user-button button"
-            type="button"
-          >
-            User&nbsp;#2
-          </button>
-        </li>
+            <label>
+              <input type="checkbox" checked={todo.completed} readOnly />
```
irenhh, line 52: checked should be changable for the user

src/App.jsx
```diff
@@ -3,26 +3,86 @@ import './App.scss';
...
+          />
         </div>
 
         <div className="App__content">
           <div className="App__content-container">
             {selectedUserId ? (
-              <CurrentUser userId={selectedUserId} />
+              <CurrentUser
+                userId={selectedUserId}
+                getInfo={getUserInfo}
```
vsm2k17, line 83: no need to pass getInfo as a props, you can import it inside CurrentUser

src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,64 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import './CurrentUser.scss';
 
-export const CurrentUser = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+export class CurrentUser extends React.Component {
+  state = {
+    user: '',
```
vsm2k17, line 7: user its an object, right? then why here is a string

src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,64 @@
...
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.userId === this.props.userId) {
+      return;
+    }
+
+    this.props.getInfo(this.props.userId)
+      .then((user) => {
+        this.setState({ user });
+      });
```
vsm2k17, line 25: handle cases when user is null

vsm2k17 (changes requested):
Screenshot 2020-09-15 at 11 09 37
after I clicked user 11

src/components/CurrentUser/CurrentUser.jsx
furude-rika: I thought we need to set any empty value by default, but I'll fix it

vsm2k17: the value of state should never change his type

vsm2k17 (approved):

PR #261, opened 2020-09-14
Solution

uni-8:
[DEMO LINK](https://uni-8.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -3,26 +3,99 @@ import './App.scss';
 import './styles/general.scss';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
+import { getTodos, getUser } from './helper/api';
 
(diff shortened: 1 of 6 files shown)


zarva6596 (approved):

src/App.jsx
```diff
@@ -3,26 +3,99 @@ import './App.scss';
...
+          case 'completed':
+            return todo.completed;
+          case 'active':
+            return !todo.completed;
+          default:
+            return todo;
+        }
+      }),
+    }));
+  }
```
vkryvytskyy, line 76: use this function in the TodoList component to dynamically get your preparedTodos there, and don't store them in state here

vkryvytskyy (changes requested):
Would be nice if filtering by status could works along with filtering by title.

PR #325, opened 2021-01-26
Develop

HaidaiAndrii:
[DEMO LINK](https://HaidaiAndrii.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -3,6 +3,7 @@ import './App.scss';
 import './styles/general.scss';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
+import { getTodos } from './api';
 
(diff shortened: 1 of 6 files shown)


Lara-l-art (approved):
Great!

src/App.jsx
```diff
@@ -3,26 +3,61 @@ import './App.scss';
...
+  };
+
+  setQuery = (str) => {
+    this.setState({
+      query: str,
+    });
+    this.setState(state => ({
+      todos: state.todos,
+    }));
+  }
```
aionova, line 30: why do you need to keep query in this component? You can have it in the TodoList and not to pass it anywhere

aionova (changes requested):
Nice start!

PR #338, opened 2021-01-27
Solution

Iryna713:
[DEMO LINK](https://Iryna713.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/CurrentUser/CurrentUser.jsx
+++ b/src/components/CurrentUser/CurrentUser.jsx
@@ -1,12 +1,92 @@
 import React from 'react';
 import './CurrentUser.scss';
+import PropTypes from 'prop-types';
+import { getUser } from '../../api';
 
-export const CurrentUser = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+export class CurrentUser extends React.PureComponent {
+  state = {
+    user: {},
+    loadError: false,
+  }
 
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,137 @@
...
+                    checked={todo.completed}
+                  />
+                  <p>{todo.title}</p>
+                </label>
+
+                <button
+                  className={classnames(
+                    'TodoList__user-button',
+                    'button',
+                    { 'TodoList__user-button--selected': true },
```
ira-cherpak, line 120: ` { 'TodoList__user-button--selected': todo.userId === selectedUserId}, `

ira-cherpak (approved):
Good job

src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,92 @@
...
+  render() {
+    const { user, loadError } = this.state;
+    const { clearUser } = this.props;
+
+    return (
+      <div className="CurrentUser">
+        <h2 className="CurrentUser__title">
+          <span>
+            Selected user:&nbsp;
+            {user && user.id}
```
aionova, line 58: we can check if user exists outside return statement and return null if it doesn't. In this case you won't need to do 4 user checks in return.

aionova (approved):
Good job!
Next challenge is to make todos change their color when checked

PR #348, opened 2021-01-28
Develop

Layrovell:
[DEMO LINK](https://Layrovell.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -12,56 +12,22 @@ import { LoadingError } from './components/LoadingError/LoadingError';
 class App extends React.Component {
   state = {
     todos: [],
-    filteredTodos: [],
     selectedUserId: 0,
-    query: '',
     selectedStatus: '',
-    loading: false,
-    hasLoadingError: false,
-    option: '',
   };
 
-  loadTodos = async() => {
-    this.setState({
-      loading: true,
-      hasLoadingError: false,
-    });
-
-    try {
-      const todos = await getTodos()
-      this.setState({
-        todos: todos,
-      });
-    } catch (error) {
-      this.setState({
-        hasLoadingError: true,
-        loading: false,
-      });
-    }
-  }
-
-  componentDidUpdate(prevProps, prevState, snapshot) {
-    const { query, todos } = this.state;
-    if (prevState.query !== query) {
-      const filteredTodos = todos
-        .filter(todo => {
-          if (todo.title === null) {
-            return false;
-          } else {
-          return todo.title.includes(query);
-          }
-        });
+  async componentDidMount() {
+      try {
+        const todos = await getTodos();
 
         this.setState({
-        filteredTodos,
-      })
-    }
-  }
-
-  filterHandler = (e) => {
-    this.setState({
-      query: e.target.value
-    });
+          todos: todos.filter(todo => todo.userId !== null && todo.title !== ''),
+        });
+      } catch (error) {
+        this.setState({
+          hasLoadingError: true,
+        });
+      }
   }
 
   deleteUser = () => {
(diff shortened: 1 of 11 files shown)


src/App.jsx
```diff
@@ -1,32 +1,127 @@
...
 
-        <div className="App__content">
-          <div className="App__content-container">
-            {selectedUserId ? (
-              <CurrentUser userId={selectedUserId} />
-            ) : 'No user selected'}
+                <TodoList
+                  todos={filteredTodos.length > 0
+                    ? filteredTodos
+                    : todos}
```
varseniuk, line 101: it's better to already prepare todos before passing them to 'TodoList' component

src/api/api.js
```diff
@@ -0,0 +1,29 @@
...
+  return request('/todos');
+};
+
+export const getUsers = () => request('/users');
+
+export const getUser = selectedUserId => request(`/users/${selectedUserId}`);
+
+// export const getUserTodo = (selectedUserId) => {
+//   return request(`/users/${selectedUserId}/todos`);
+// };
```
varseniuk, line 29:
```suggestion
```

src/api/api.js
```diff
@@ -0,0 +1,29 @@
...
+  return new Promise(resolve => setTimeout(resolve, delay));
+}
+
+export const getTodos = async() => {
+  await wait(1000);
+
+  return request('/todos');
+};
+
+export const getUsers = () => request('/users');
```
varseniuk, line 23:
```suggestion
```
Why do you need this? You don't use this function never after

src/api/api.js
```diff
@@ -0,0 +1,29 @@
...
+
+function wait(delay) {
+  return new Promise(resolve => setTimeout(resolve, delay));
+}
+
+export const getTodos = async() => {
+  await wait(1000);
+
+  return request('/todos');
+};
```
varseniuk, line 21: This part is redundant. Single 'fetch' (without 'options') would be enough

src/api/api.js
```diff
@@ -0,0 +1,29 @@
+const TODO_API_URL = 'https://mate-api.herokuapp.com';
+
+export const request = (url, options) => fetch(`${TODO_API_URL}${url}`, options)
```
varseniuk, line 3:
```suggestion
export const request = url => fetch(`${TODO_API_URL}${url}`)
```

src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,50 @@
...
+  };
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+  async componentDidMount() {
+    const user = await getUser(this.props.userId);
+    this.setState({ user });
```
varseniuk, line 13: You are duplicating this code here and in componentDidUpdate.
Let's move it to a separate function and call it twice

src/components/TodoItem/TodoItem.jsx
```diff
@@ -0,0 +1,49 @@
...
+                button is-link
+                {/*TodoList__user-button TodoList__user-button--selected button*/}
+                "
+                type="button">User: {todo.userId}
+              </button>)
+          : ( <button
+                onClick={() => { selectUser(todo.id) }}
+                className="
+                button is-info is-outlined
+                {/*TodoList__user-button TodoList__user-button--selected button/}
```
varseniuk, line 28: ? Looks strange

varseniuk (changes requested):
Why when are you are selecting one user, your app displaying another?





The filter is not working. I can enter some weird query and it will show all todos instead of none

src/App.jsx
```diff
@@ -1,32 +1,72 @@
...
     selectedUserId: 0,
+    selectedStatus: '',
   };
 
+  async componentDidMount() {
+      try {
+        const todos = await getTodos();
+
+        this.setState({
+          todos: todos.filter(todo => todo.userId !== null && todo.title !== ''),
```
varseniuk, line 24: Без названия (6)

There are todos with no data in the list. You need to filter if data is not null as well

varseniuk (changes requested):
Some more little things to do left

varseniuk (approved):
Good work!

PR #351, opened 2021-01-28
Solution

DarMatt:
[DEMO LINK](https://DarMatt.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/CurrentUser/CurrentUser.jsx
+++ b/src/components/CurrentUser/CurrentUser.jsx
@@ -1,12 +1,54 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+import { getUser } from '../../api/api';
 import './CurrentUser.scss';
 
-export const CurrentUser = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+export class CurrentUser extends React.Component {
+  state = {
+    user: null,
+  };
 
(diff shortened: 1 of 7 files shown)


zhenya-mezhueva (approved):

src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,54 @@
...
+    if (prevProps.userId !== this.props.userId) {
+      getUser(this.props.userId)
+        .then(user => this.setState({ user }));
+    }
+  }
+
+  render() {
+    const { clearUser } = this.props;
+    const { user } = this.state;
+
```
aionova, line 26: before return statement you can check if user exists, and return null otherwise. Then you won't need to check if the user is found 4 times

src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,54 @@
...
+            {user && user.id}
+          </span>
+        </h2>
+
+        <h3 className="CurrentUser__name">{user && user.name}</h3>
+        <p className="CurrentUser__email">{user && user.email}</p>
+        <p className="CurrentUser__phone">{user && user.phone}</p>
+        <button
+          type="button"
+          onClick={() => clearUser()}
```
aionova, line 42:
```suggestion
          onClick={clearUser}
```

aionova (changes requested):
Nice start!

PR #356, opened 2021-01-28
Develop

dtsehelnyk:
[DEMO LINK](https://dtsehelnyk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.jsx
+++ b/src/components/TodoList/TodoList.jsx
@@ -1,44 +1,120 @@
 import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
 import './TodoList.scss';
 
-export const TodoList = () => (
-  <div className="TodoList">
-    <h2>Todos:</h2>
+export class TodoList extends React.Component {
+  state = {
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,120 @@
...
+        filteredTodos = todos.filter(todo => todo.completed);
+        break;
+
+      default:
+        filteredTodos = todos;
+        break;
+    }
+
+    if (todoTitle.length > 0) {
+      filteredTodos = filteredTodos.filter((todo) => {
```
varseniuk, line 41: For the future - it's better to make all operations (filtering) in the upper level, and pass ALREADy prepared todos to 'TodoList' component. It should only render the list

varseniuk (changes requested):
Без названия (2)

Let's filter incoming todos, because there are 'broken' todos here. E.g. some user doesn't exist. Pay more attetion to the source file

varseniuk (approved):
Good job! 
It's nice that you have remade it with hooks 👍

PR #359, opened 2021-01-29
Solution

MaximZhuravlov:
- [DEMO LINK](https://MaximZhuravlov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/CurrentUser/CurrentUser.jsx
+++ b/src/components/CurrentUser/CurrentUser.jsx
@@ -1,12 +1,66 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+import { getUser } from '../../api/api';
+
 import './CurrentUser.scss';
 
-export const CurrentUser = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+export class CurrentUser extends React.Component {
+  state = {
+    user: null,
+  };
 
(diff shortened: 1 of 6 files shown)


src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,66 @@
...
+  state = {
+    user: null,
+  };
+
+  async componentDidMount() {
+    const { userId } = this.props;
+
+    const user = await getUser(userId);
+
+    this.setState({ user });
```
varseniuk, line 17: You are duplicating this part of code twice. Let's make a function for this

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,124 @@
...
+        switch (criterion) {
+          case 'active':
+            return !item.completed;
+          case 'completed':
+            return item.completed;
+          default:
+            return true;
+        }
+      })
+      .filter((item) => {
```
varseniuk, line 42: For the future - better to make all the filtering on the upper level and pass here only ALREADY prepared todos

varseniuk (changes requested):
Без названия (4)


Let's filter incoming todos, because there are 'broken' todos here. E.g. some user doesn't exist. Pay more attetion to the source file

varseniuk (approved):

PR #360, opened 2021-01-31
dynamic list of todos

WhiteRhin0:
- [DEMO LINK](https://WhiteRhin0.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/CurrentUser/CurrentUser.jsx
+++ b/src/components/CurrentUser/CurrentUser.jsx
@@ -1,12 +1,65 @@
 import React from 'react';
 import './CurrentUser.scss';
+import PropTypes from 'prop-types';
+import { getUser } from '../../api';
 
-export const CurrentUser = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+export class CurrentUser extends React.Component {
+  state = {
+    user: null,
+  }
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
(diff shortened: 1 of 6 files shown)


src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,65 @@
...
+    });
+  }
+
+  render() {
+    const { user } = this.state;
+    const { clearUser } = this.props;
+
+    if (user === null) {
+      return null;
+    }
```
varseniuk, line 35:
```suggestion
    if (!user) {
      return;
    }
```

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,120 @@
...
+  render() {
+    const { todos, selectUser, selectedUserId } = this.props;
+    const { taskToFind, selectedTodos } = this.state;
+
+    const findTodos = todos.filter(todo => (
+      todo.title !== null
+      && todo.title.toLowerCase().includes(taskToFind.toLowerCase())
+    ));
+
+    const filterTodos = findTodos.filter((todo) => {
```
varseniuk, line 27: For the future - it's better to make all filtering at the top level and pass already prepared todos to the 'TodoList' component

varseniuk (changes requested):
Без названия (1)

Let's filter incoming todos, because there are 'broken' todos here. E.g. some user doesn't exist. PAy more attetion to the source file

varseniuk (approved):
👍

PR #363, opened 2021-02-01
Dynamic list of todos	

kerpichonka:
[DEMO LINK](https://kerpichonka.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/CurrentUser/CurrentUser.jsx
+++ b/src/components/CurrentUser/CurrentUser.jsx
@@ -1,12 +1,92 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+import { getUserId } from '../../data/api';
 import './CurrentUser.scss';
 
-export const CurrentUser = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+export class CurrentUser extends React.Component {
+  state = {
+    user: null,
+    loading: false,
+    error: false,
+  }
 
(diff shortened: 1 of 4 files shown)


src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,92 @@
...
+    this.setState({ loading: true });
+    getUserId(userId).then((response) => {
+      this.setState({
+        user: response.data,
+        loading: false,
+      });
+    }).catch(() => this.setState({
+      loading: false,
+      error: true,
+    }));
```
varseniuk, line 25: You are duplicating this code twice in your code. In componentDiDMount and in ComponentDidUpdate. Let's make separate function for fetching user and use it in two places instead of repeating your code

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,126 @@
...
+    }
+
+    switch (status) {
+      case 'Active':
+        filterTodos = filterTodos.filter(
+          todo => todo.completed !== null && !todo.completed,
+        );
+        break;
+      case 'Completed':
+        filterTodos = filterTodos.filter(
```
varseniuk, line 37: For the future - it would be better to make all operations (filtering) in the upper level and pass ALREADY prepared todos to the 'TodoList' component

varseniuk (changes requested):
Без названия (1)

Let's filter incoming todos, because there are 'broken' todos here. E.g. some user doesn't exist. PAy more attetion to the source file

PR #431, opened 2021-06-02
solution

ched88xtc:
[DEMO LINK](https://ched88xtc.github.io/react_dynamic-list-of-todos/)


src/App.jsx
```diff
@@ -3,26 +3,87 @@ import './App.scss';
...
+
+  titleSearchChangeHandler = ({ name, value }) => {
+    this.setState({ [name]: value });
+  }
+
+  setTodosStatus = (status) => {
+    this.setState({ todosStatus: status });
+  }
+
+  prepareTodos = () => {
```
victor-arutiunov, line 41: You can create list prepareTodos in render without method.

src/App.jsx
```diff
@@ -3,26 +3,87 @@ import './App.scss';
...
+
+    let visibleTodos = [...todos];
+
+    if (todosStatus === 'Active') {
+      visibleTodos = [...todos].filter(({ completed }) => !completed);
+    }
+
+    if (todosStatus === 'Completed') {
+      visibleTodos = [...todos].filter(({ completed }) => completed);
+    }
```
victor-arutiunov, line 52: Maybe it will be better to use one filter with switch case inside.

src/App.jsx
```diff
@@ -3,26 +3,87 @@ import './App.scss';
...
+
+    if (titleSearch.trim()) {
+      visibleTodos = [...visibleTodos].filter(todo => (
+        (`${todo.title}`).toLowerCase().includes(titleSearch.toLowerCase())
+      ));
+    }
```
victor-arutiunov, line 58: And after first filter immediately use second filter, like .filter().filter()

src/api/api.js
```diff
@@ -0,0 +1,8 @@
+const BASE_URL = 'https://mate-api.herokuapp.com';
+
+export function request(url, options) {
+  return fetch(
+    !options ? `${BASE_URL}${url}` : `${BASE_URL}${url}${options}`,
```
victor-arutiunov, line 5: Looks interesting, like it)

src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,53 @@
...
+  }
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+  componentDidMount() {
+    request('/users', `/${this.props.userId}`)
+      .then(({ data }) => this.setState({ user: { ...data } }));
```
victor-arutiunov, line 13: When you write in url userId you receive user obj, not data. So you can make it simpler. Like: _.then(user => this.setState({ user }) )_

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,104 @@
...
+        Select todo status
+        <select
+          name="todo-status"
+          id="todo-status"
+          className="TodoList__status-select"
+          onChange={event => setTodosStatus(event.target.value)}
+        >
+          <option
+            value="All"
+          >
```
victor-arutiunov, line 37: When use one attribute you can write it one line.

victor-arutiunov (commented):
Nice work!

src/components/CurrentUser/CurrentUser.jsx
ched88xtc: it doesn't work)

VinnieJ-2k20: Destructuring is used here, so it needs exact property name to work correctly.

src/App.jsx
VinnieJ-2k20: It is strongly recommended not to overload render method with logic, so this way is correct.

src/App.jsx
VinnieJ-2k20: Agree, chained methods look better, but it's not critical.

VinnieJ-2k20 (approved):
Good job 👍

src/App.jsx
victor-arutiunov: Okay, sorry for missadvice

PR #435, opened 2021-06-02
Solution

avgusta92:
- [DEMO LINK](https://avgusta92.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -4,28 +4,53 @@ import './styles/general.scss';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
 
+import { getTodos } from './api';
+
 class App extends React.Component {
   state = {
     todos: [],
     selectedUserId: 0,
   };
 
+  async componentDidMount() {
+    const todos = await getTodos();
+
+    this.setState({
+      todos,
+    });
+  }
+
+  selectUser = (selectedUserId) => {
+    this.setState({ selectedUserId });
+  }
+
   render() {
     const { todos, selectedUserId } = this.state;
 
-    return (
+    return todos.length !== 0 ? (
       <div className="App">
         <div className="App__sidebar">
-          <TodoList todos={todos} />
+          <TodoList
+            todos={todos.data}
+            selectedUserId={selectedUserId}
+            onSelectUser={this.selectUser}
+          />
         </div>
 
         <div className="App__content">
           <div className="App__content-container">
             {selectedUserId ? (
-              <CurrentUser userId={selectedUserId} />
+              <CurrentUser
+                userId={selectedUserId}
+                onClearSelectUser={this.selectUser}
+              />
             ) : 'No user selected'}
           </div>
         </div>
+      </div>
+    ) : (
+      <div>
+        <p>Loading...</p>
       </div>
     );
   }
--- a/src/api.js
+++ b/src/api.js
@@ -1,0 +1,14 @@
+const BASE_URL = 'https://mate-api.herokuapp.com';
+
+export const request = (url, option) => fetch(`${BASE_URL}${url}`, option)
+  .then((res) => {
+    if (!res.ok) {
+      throw new Error(`${res.status} – ${res.statusText}`);
+    }
+
+    return res.json();
+  });
+
+export const getTodos = async() => request('/todos');
+
+export const getUser = async userId => request(`/users/${userId}`);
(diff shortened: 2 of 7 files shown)


src/App.jsx
```diff
@@ -4,29 +4,54 @@ import './styles/general.scss';
...
   render() {
     const { todos, selectedUserId } = this.state;
 
-    return (
+    return todos.length !== 0 ? (
       <div className="App">
         <div className="App__sidebar">
-          <TodoList todos={todos} />
+          <TodoList
+            todos={todos.data}
```
Ihor-Karpyn, line 34: Better take data in your API file, because if server system will change you must solve mistake in a lot of places
```suggestion
            todos={todos.data}
```

src/api.js
```diff
@@ -0,0 +1,14 @@
...
+    if (!res.ok) {
+      throw new Error(`${res.status} – ${res.statusText}`);
+    }
+
+    return res.json();
+  });
+
+export const getTodos = async() => request('/todos');
+
+export const getUser = async userId => request(`/users/${userId}`);
```
Ihor-Karpyn, line 14: Why you use 'async' without 'await'. Async redundant here.

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,126 @@
...
+
+    return todos.filter(todo => todo.title
+      && todo.title.toLowerCase().includes(normalizedQuery));
+  }
+
+  filterCompletedTodos = (todos, completedFilter) => todos.filter(todo => (
+    completedFilter === 'completed'
+      ? todo.completed
+      : !todo.completed
+  ))
```
Ihor-Karpyn, line 32: I think in this case switch will better, because if you will want to add one more sorting option it will be difficult

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,126 @@
...
-            User&nbsp;#2
-          </button>
-        </li>
-      </ul>
-    </div>
-  </div>
-);
+            <option name="all">all</option>
+            <option name="active">active</option>
+            <option name="completed">completed</option>
```
Ihor-Karpyn, line 64:
```suggestion
            <option value="all">all</option>
            <option value="active">active</option>
            <option value="completed">completed</option>
```

Ihor-Karpyn (approved):
Good job!1 Left a few comments

src/App.jsx
aliceblunt8: pass from getTodos todos.data, so you won't need to use in component

src/components/CurrentUser/CurrentUser.jsx
```diff
@@ -1,12 +1,62 @@
...
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+export class CurrentUser extends React.Component {
+  state = {
+    user: null,
+  };
+
+  componentDidMount() {
```
aliceblunt8, line 12: remove componentDidMount, componentDidUpdate is enough

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,126 @@
...
-            <input type="checkbox" checked readOnly />
-            <p>distinctio vitae autem nihil ut molestias quo</p>
-          </label>
-
-          <button
-            className="TodoList__user-button button"
-            type="button"
+export class TodoList extends React.PureComponent {
+  state = {
+    todos: this.props.todos,
```
aliceblunt8, line 8: don't add todos to TodoList state, pass already filtered todos

aliceblunt8: use them as props inside this component

src/components/TodoList/TodoList.jsx
aliceblunt8, line 32: do this in App.js

src/components/TodoList/TodoList.jsx
aliceblunt8, line 64: move to App.js

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,126 @@
...
+                        === selectedUserId,
+                    },
+                  )}
+                  type="button"
+                  onClick={() => onSelectUser(todo.userId)}
+                >
+                  User&nbsp;#
+                  {todo.userId}
+                </button>
+              </li>
```
aliceblunt8, line 101: Todo component?

varseniuk (changes requested):
https://user-images.githubusercontent.com/66570142/120992881-b2424080-c78b-11eb-81d7-0ba68bfe7bbf.mov

Something goes wrong when we try to enter a query that does not match any title. Let's fix this

PR #479, opened 2021-06-06
test

C-y-y-y:
[DEMO LINK](https://C-y-y-y.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,8 +1,12 @@
 import React from 'react';
+
 import './App.scss';
 import './styles/general.scss';
+
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
+
+import { getTodos } from './api/todos';
 
 class App extends React.Component {
   state = {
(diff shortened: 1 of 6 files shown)


senkiv-oleh (approved):
Було б добре, щоб після натискання кнопки Clear кнопка User ставала сірою

src/App.jsx
```diff
@@ -1,28 +1,50 @@
...
   render() {
     const { todos, selectedUserId } = this.state;
 
     return (
       <div className="App">
         <div className="App__sidebar">
-          <TodoList todos={todos} />
+          <TodoList
+            todos={todos}
+            selectedUser={selectedUserId}
```
varseniuk, line 36:
```suggestion
```
You are passing `selectedUser` to the `TodoList` component but never use it there

src/api/todos.js
```diff
@@ -0,0 +1,14 @@
+const BASE_URL = 'https://mate-api.herokuapp.com';
+
```
varseniuk, line 2: this file better is called `api.js`. Here you have logic not only for fetching todos but also for fetching users. So the name `todos.js` does not actually mean what it should do.

src/components/TodoList/TodoList.jsx
```diff
@@ -1,44 +1,130 @@
...
+
+  filterByStatus = (todo) => {
+    const { filterStatus } = this.state;
+
+    switch (filterStatus) {
+      case 'completed':
+        return todo.completed;
+      case 'active':
+        return !todo.completed;
+      default: return true;
```
varseniuk, line 40:
```suggestion
      default: return;
```

varseniuk (approved):
👍 
Left some comments

PR #539, opened 2021-09-13
Solution

ivanchenko-kirill:
[DEMO LINK](https://ivanchenko-kirill.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -14,7 +14,6 @@ interface State {
   searchQuery: string;
   filterByStatus: FilterByStatusTypes;
   hasLoadingError: boolean;
-  isShuffled: boolean
 }
 
 interface Props {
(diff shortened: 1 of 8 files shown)


alex-glotov67 (approved):
Great Job!

Elena-brv (changes requested):
when you load the first user it shouldn't show this - (you can detect it on slow 3G)
image

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,172 @@
...
-  </div>
-);
+
+          <select
+            onChange={this.handleSelect}
+            value={filterByStatus}
+          >
+            <option value="" disabled>
+              Choose status type
+            </option>
```
Elena-brv, line 115: it's no sense to have this option, because when you will choose All - nothing gonna be changed

Elena-brv (changes requested):
image

check it on a network tab

image

Elena-brv (approved):

PR #541, opened 2021-09-14
Dynamic list of goods

nastyyvolk:
[DEMO LINK](https://nastyyvolk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/CurrentUser/CurrentUser.tsx
+++ b/src/components/CurrentUser/CurrentUser.tsx
@@ -1,12 +1,63 @@
 import React from 'react';
 import './CurrentUser.scss';
+import { getUser } from '../../api';
 
-export const CurrentUser: React.FC = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+type Props = {
+  userId: number;
+  selectUser: (id: number) => void;
+};
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+type State = {
+  user: User | null;
+};
+
+export class CurrentUser extends React.Component<Props, State> {
+  state: State = {
+    user: null,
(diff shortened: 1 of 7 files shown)


src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,69 @@
...
+      });
+  };
+
+  render() {
+    const { user, loadingError } = this.state;
+    const { selectUser } = this.props;
+
+    return (
+      <div className="CurrentUser">
+        {loadingError && (<p> Loading error. Please, try again... </p>)}
```
lera-yaskevych, line 47: When I'm trying to load user#197 I see no error, for some reason. Perhaps the reason for this is in the try{} catch{} statement

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,69 @@
...
+
+  getUser = () => {
+    getUser(this.props.userId)
+      .then(user => {
+        this.setState({ user });
+      })
+      .catch(()=> {
+        this.setState({ loadingError: true })
+      });
+  };
```
lera-yaskevych, line 39:
```suggestion
  getUser = () => {
  try {
    getUser(this.props.userId)
      .then(user => {
        this.setState({ user });
      });
  } catch (error) {
        this.setState({ loadingError: true })
      });
  };
```
Try this. Let me know whether it fixes the problem

lera-yaskevych (commented):
Left one remark. Apart from that everything is nice 👍

lera-yaskevych (approved):

src/components/CurrentUser/CurrentUser.tsx
nastyyvolk: thanks! I'll check that 🤔

src/components/CurrentUser/CurrentUser.tsx
nastyyvolk: I'm not sure whether it is the correct syntax to use 'try-catch' with 'then'... But I'll see what I can do

danheim (approved):

PR #544, opened 2021-09-15
Develop

Moroz-Dmytro:
[DEMO LINK](https://Moroz-Dmytro.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/CurrentUser/CurrentUser.tsx
+++ b/src/components/CurrentUser/CurrentUser.tsx
@@ -1,12 +1,76 @@
 import React from 'react';
 import './CurrentUser.scss';
+import { loadUserData } from '../../api';
 
-export const CurrentUser: React.FC = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+interface Props {
+  userId: number;
+  clearSelectedUser: () => void;
+}
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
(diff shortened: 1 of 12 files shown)


DmytroMykoliv (approved):
As for me looks good )

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,116 @@
...
+  };
+
+  getFilteredTodos = (todos: Todo[]) => {
+    const { titlePhrase } = this.state;
+
+    return todos.filter(
+      todo => {
+        const { title, completed } = todo;
+
+        return title && title.includes(titlePhrase) && this.getTodoStatus(completed);
```
Elena-brv, line 56: it will not find if the search is entered in uppercase. In such cases, it is better to add the ability to print in any case and find the results

Elena-brv (changes requested):
when you load the first user it shouldn't show this - (you can detect it on slow 3G)

image

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,99 @@
...
+        {
+          loading ? (
+            <h2>Loading...</h2>
+          ) : (
+            <>
+              {error && (
+                <h2>Ups... Something went wrong</h2>
+              )}
+              {!error && (
+                <div className="CurrentUser">
```
Elena-brv, line 76: ternary operator

Elena-brv (approved):

PR #550, opened 2021-09-16
error

dima-tkachenko:


UlaDutchak (approved):

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,64 @@
...
+
+  async componentDidMount() {
+    this.reloadUser(this.props.userId);
+  }
+
+  async componentDidUpdate(prevProps: Props) {
+    if (prevProps.userId !== this.props.userId) {
+      this.reloadUser(this.props.userId);
+    }
+  }
```
Elena-brv, line 27: no need async

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,64 @@
...
+    });
+  }
+
+  render() {
+    const { user } = this.state;
+
+    if (!user) {
+      return (
+        <p>
+          error
```
Elena-brv, line 43:
```suggestion
           Error
```

Elena-brv (changes requested):
the order should not change by clicking on the user
https://www.loom.com/share/c82113121c1a4c0cb1e6225c71e8e2ad

Elena-brv (changes requested):
after clicking Randomize button and then click the user the order changes every time. I don't see any fixes to this problem

PR #561, opened 2021-09-17
Solution

poliandiya:
[DEMO LINK](https://poliandiya.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/CurrentUser/CurrentUser.tsx
+++ b/src/components/CurrentUser/CurrentUser.tsx
@@ -1,12 +1,67 @@
 import React from 'react';
+import { loadUser } from '../../api';
 import './CurrentUser.scss';
 
-export const CurrentUser: React.FC = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+interface Props {
+  userId: number;
+  clear: () => void;
+}
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+type State = {
+  user: User | null;
+};
+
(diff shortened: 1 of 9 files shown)


src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,67 @@
...
+    if (prevProps.userId !== this.props.userId) {
+      this.reloadUser(this.props.userId);
+    }
+  }
+
+  async reloadUser(userId: number) {
+    const user = await loadUser(userId);
+
+    this.setState({ user });
+  }
```
Elena-brv, line 33:
```suggestion
  componentDidMount() {
    this.reloadUser();
  }

  componentDidUpdate(prevProps: Props) {
    if (prevProps.userId !== this.props.userId) {
      this.reloadUser();
    }
  }

  async reloadUser(userId: number) {
    const user = await loadUser(this.props.userId);

    this.setState({ user });
  }
```
no need to use async on lifecycle methods here

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,67 @@
...
+
+  render() {
+    const { user } = this.state;
+
+    if (!user) {
+      return (
+        <div>User not found</div>
```
Elena-brv, line 40:
```suggestion
        <p>User not found<p>
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,110 @@
...
+    let filteredTodos = todos
+      .filter(todo => todo.title.includes(query));
+
+    if (selectStateOfTodo === 'active') {
+      filteredTodos = filteredTodos
+        .filter(todo => !todo.completed);
+    } else if (selectStateOfTodo === 'completed') {
+      filteredTodos = filteredTodos
+        .filter(todo => todo.completed);
+    }
```
Elena-brv, line 33: switch/case

Elena-brv (approved):

PR #564, opened 2021-09-17
Develop

Oleksandr-Zhitenko:
[DEMO LINK](https://Oleksandr-Zhitenko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,29 +3,57 @@ import './App.scss';
 import './styles/general.scss';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
+import { loadTodos } from './api';
 
 interface State {
+  todos: Todo[],
   selectedUserId: number;
 }
 
 class App extends React.Component<{}, State> {
   state: State = {
+    todos: [],
     selectedUserId: 0,
   };
 
+  async componentDidMount() {
+    const data = await loadTodos();
+
+    this.setState({
+      todos: data,
+    });
+  }
+
+  handeUserSelection = (userId: number) => {
+    this.setState({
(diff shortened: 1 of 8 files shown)


zadorojniiarsen (approved):

src/App.tsx
```diff
@@ -3,29 +3,57 @@ import './App.scss';
...
+    this.setState({
+      selectedUserId: 0,
+    });
+  };
+
   render() {
-    const { selectedUserId } = this.state;
+    const { selectedUserId, todos } = this.state;
+
+    // eslint-disable-next-line no-console
```
Elena-brv, line 42: remove

src/api.js
```diff
@@ -0,0 +1,13 @@
+const API_URL = 'https://mate.academy/students-api';
+
+export const loadTodos = async () => {
+  const response = await fetch(`${API_URL}/todos`);
+
+  return response.json();
+};
+
+export const loadUsers = async (userId) => {
```
Elena-brv, line 9:
```suggestion
export const loadUser = async (userId) => {
```

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,71 @@
...
+  state: State = {
+    user: null,
+  };
+
+  async componentDidMount() {
+    const user = await loadUsers(this.props.userId);
+
+    this.setState({
+      user,
+    });
```
Elena-brv, line 24: make a separate function and call it here. You use the same code here and in componentDidUpdate

Elena-brv (changes requested):
filter by title doesn't work if all todos chosed 
image

PR #573, opened 2021-09-18
Developo

Vova-Stoliar:
[DEMO LINK](https://Vova-Stoliar.github.io/react_dynamic-list-of-todos/)


Elena-brv (changes requested):
filter by status and filter by title should be synchronized. And do not reset if something has changed
https://www.loom.com/share/9a329645281041c183d3e8a24ccf9efa

PR #575, opened 2021-09-19
[DEMO LINK](https://george-w-boo.github.io/react_dynamic-list-of-todos/)

george-w-boo:


src/App.tsx
```diff
@@ -1,31 +1,99 @@
...
+    switch (todoStatus) {
+      case 'completed':
+        this.setState(currentState => {
+          const filtered = currentState.todos.filter(todo => todo.completed);
+
+          return ({
+            filteredTodos: [...filtered],
+          });
+        });
+        break;
```
Elena-brv, line 60: filter makes a new array, it's no need to make a copy

src/api.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (userId) {
+    response = await fetch(`${API_URL}/${path}/${userId}`);
+
+    return response.json();
+  }
+
+  response = await fetch(`${API_URL}/${path}`);
+
+  return response.json();
+};
```
Elena-brv, line 15: better make different functions for loading different information (for more readable code)

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,71 @@
...
+    }
+  }
+
+  async reloadUser(userId: number) {
+    const user = await getData('users', userId);
+
+    this.setState({
+      user,
+    });
+  }
```
Elena-brv, line 35:
```suggestion
  async componentDidMount() {
    this.reloadUser();
  }

  async componentDidUpdate(prevProps: Props) {
    if (prevProps.userId !== this.props.userId) {
      this.reloadUser();
    }
  }

  async reloadUser(userId: number) {
    const user = await getData('users', this.props.userId);

    this.setState({
      user,
    });
  }
```

Elena-brv (changes requested):
the status by which I am filtering should be visible here
image

george-w-boo:
> the status by which I am filtering should be visible here
> image

I don't think it's a good idea in terms of UX. Whenever the user changes the filter, the 'Filter by' won't be visible anymore ((

Elena-brv:
> > the status by which I am filtering should be visible here
> > image
> 
> I don't think it's a good idea in terms of UX. Whenever the user changes the filter, the 'Filter by' won't be visible anymore ((

But now it's difficult to recognize what todos I see now

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,71 @@
...
+
+  async componentDidMount() {
+    this.reloadUser();
+  }
+
+  async componentDidUpdate(prevProps: Props) {
+    if (prevProps.userId !== this.props.userId) {
+      this.reloadUser();
+    }
+  }
```
Elena-brv, line 27: no need async here

Elena-brv (changes requested):
filter by title and filter by status should be synchronized (now if I filter by title, filter by status resets, and vice versa)

src/App.tsx
```diff
@@ -1,31 +1,121 @@
...
+  };
+
+  handleUserCleaning = () => {
+    this.setState({
+      selectedUserId: 0,
+    });
+  };
+
+  filterTodosByTitle = (title: string) => {
+    if (this.state.filterType === 'all') {
```
Elena-brv, line 47: switch/case

Elena-brv: And you make similar filtering in filterByTitle and filterByStatus

Elena-brv (changes requested):
https://www.loom.com/share/c75ddcbdef8b48e998f6349b1a073115

PR #632, opened 2021-11-18
Develop

meforoma:
https://meforoma.github.io/react_dynamic-list-of-todos/


src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,126 @@
-import React from 'react';
+/* eslint-disable no-console */
```
varseniuk, line 1:
```suggestion
```

varseniuk (approved):
good job! 🔥

PR #683, opened 2021-12-26
Solution

Oleh-Khashchevskyi:
[DEMO LINK](https://Oleh-Khashchevskyi.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -3,29 +3,113 @@ import './App.scss';
...
+                  className="search__todo"
+                  placeholder="search"
+                  value={searchValue}
+                  onChange={(event) => {
+                    this.setState(state => ({
+                      searchValue: event.target.value,
+                      visibleTodo: state.todos
+                        .filter(todo => todo.title.toLowerCase()
+                          .includes(event.target.value.toLowerCase())),
+                    }));
```
aliceblunt8, line 69: move to separate function?

src/App.tsx
```diff
@@ -3,29 +3,113 @@ import './App.scss';
...
+                      this.filterTodo(true);
+                    }}
+                  >
+                    Active
+                  </button>
+                  <button
+                    type="button"
+                    onClick={() => {
+                      this.filterTodo(false);
+                    }}
```
aliceblunt8, line 94: it's not obvious what true and false mean when you pass them to filterTodo function
I got it, but I think it would be better to store a value for "filter mode" (like, 'all' (default), 'completed', 'active') and then with switch case return filtered todos for each filter case

src/styles/general.scss
```diff
@@ -13,3 +13,19 @@
...
+
+.search__todo {
+  padding: 5px 10px;
+}
+
+.todos__filter {
+  display: flex;
+  align-items: center;
+  gap: 5px;
+}
```
aliceblunt8, line 31: add new line at the of the file

aliceblunt8 (changes requested):

src/App.tsx
```diff
@@ -3,29 +3,124 @@ import './App.scss';
...
+          <h2>Todos:</h2>
+          {visibleTodo && (
+            <>
+              <div className="todos__nav">
+                <input
+                  type="text"
+                  className="search__todo"
+                  placeholder="search"
+                  value={searchValue}
+                  onChange={(event) => this.handleChange(event)}
```
aliceblunt8, line 81:
```suggestion
                  onChange={this.handleChange}
```

aliceblunt8: it's the same

aliceblunt8 (approved):

PR #710, opened 2022-02-02
My solution

SavelovSerhii:
https://savelovserhii.github.io/react_dynamic-list-of-todos/


Mikedacheka (approved):

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,127 @@
...
+      createdAt: '',
+      updatedAt: '',
+      userId: 0,
+      title: '',
+      completed: false,
+    }],
+    filterByStatus: 'all',
+    titleFilter: '',
+    prevStatus: '',
+    prevTitle: '',
```
Olexandr6, line 30: These values are unnecessary here, try to implement the solution without these variables

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,127 @@
...
+  async componentDidMount() {
+    this.setState({ todos: await getTodos() });
+  }
+
+  componentDidUpdate() {
+    if (this.state.filterByStatus !== this.state.prevStatus
+      || this.state.prevTitle !== this.state.titleFilter) {
+      this.handleStatusFilter();
+    }
+  }
```
Olexandr6, line 42: `componentDidUpdate()` takes as its first two arguments the previous props and the previous state. Read more https://dev.to/cesareferrari/how-to-use-componentdidupdate-in-react-30en

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,127 @@
...
+      this.setState({
+        todos: await getTodos().then(todos => todos
+          .filter(todo => todo.completed === false && todo.title.includes(titleFilter))),
+      });
+    } else if (this.state.filterByStatus === 'all') {
+      this.setState({
+        todos: await getTodos().then(todos => todos
+          .filter(todo => todo.title.includes(titleFilter))),
+      });
+    }
```
Olexandr6, line 72: In all cases you call getTodos, so let's call this function only once

Olexandr6: Use `swith case` statement instead

Olexandr6 (changes requested):
Good job 👍  But some things should be optimized, check comments

Olexandr6 (approved):
Looks good 👍  Well done 🔥

PR #772, opened 2022-04-06
Solution

Anna-Morozevych:
[DEMO LINK](https://Anna-Morozevych.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,7 +14,7 @@ const App: React.FC = () => {
       .then(todos => setTodosList(todos));
   }, []);
 
-  const handlerSelectUser = (userId: number): void => {
+  const selectUser = (userId: number): void => {
     setSelectedUserId(userId);
   };
 
@@ -27,7 +27,7 @@ const App: React.FC = () => {
       <div className="App__sidebar">
         <TodoList
           todos={todosList}
-          onClickSelectUser={handlerSelectUser}
+          selectUser={selectUser}
         />
       </div>
 
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -7,9 +7,6 @@ export function getTodos(): Promise<Todo[]> {
 }
 
 export function getUsers(userId: number) {
-  return fetch(`${API_URL_USERS}`)
-    .then(response => response.json())
-    .then(users => {
-      return users.find((user: { id: number; }) => user.id === userId);
-    });
+  return fetch(`${API_URL_USERS}/${userId}`)
+    .then(response => response.json());
 }
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -4,45 +4,29 @@ import './TodoList.scss';
 
 type Props = {
   todos: Todo[],
-  onClickSelectUser: (userId: number) => void,
+  selectUser : (userId: number) => void,
 };
 
 export const TodoList: React.FC<Props> = ({
   todos,
-  onClickSelectUser,
+  selectUser,
 }) => {
   const [inputValue, setInputValue] = useState('');
   const [selectValue, setSelectValue] = useState('all');
 
-  const todosSelectArrFilter = (arr: Todo[], completedValue: string) => {
-    switch (completedValue) {
+  const getFilteredTodos = () => {
+    const result = todos.filter(todo => todo.title.includes(inputValue));
+
+    switch (selectValue) {
       case 'active':
-        return arr.filter(todo => todo.completed === false);
+        return result.filter(todo => todo.completed === false);
 
       case 'completed':
-        return arr.filter(todo => todo.completed === true);
+        return result.filter(todo => todo.completed === true);
 
       default:
-        return arr;
+        return result;
     }
-  };
-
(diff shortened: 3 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,76 @@
...
+const defaultUser = {
+  id: 0,
+  createdAt: '',
+  updatedAt: '',
+  name: '',
+  username: '',
+  email: '',
+  phone: '',
+  website: '',
+};
```
ArtemTopolskyi, line 17: you don't need this fake data. Just set null in state and draw loader if no user yet

src/App.tsx
```diff
@@ -1,27 +1,76 @@
...
 
 const App: React.FC = () => {
   const [selectedUserId, setSelectedUserId] = useState(0);
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [currentUser, setCurrentUser] = useState<User>(defaultUser);
```
ArtemTopolskyi, line 22: maybe move this state to User component and fetch it there. You only need to pass userId there

ArtemTopolskyi (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,56 @@
...
 const App: React.FC = () => {
   const [selectedUserId, setSelectedUserId] = useState(0);
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+
+  useEffect(() => {
+    getTodos()
+      .then(todos => setTodosList(todos));
+  }, []);
+
+  const handlerSelectUser = (userId: number): void => {
```
aliceblunt8, line 17:
```suggestion
  const selectUser = (userId: number): void => {
```

src/App.tsx
```diff
@@ -1,27 +1,56 @@
...
+    setSelectedUserId(0);
+  };
 
   return (
     <div className="App">
       <div className="App__sidebar">
-        <TodoList />
+        <TodoList
+          todos={todosList}
+          onClickSelectUser={handlerSelectUser}
```
aliceblunt8, line 30:
```suggestion
          selectUser={selectUser}
```

src/api/api.ts
```diff
@@ -0,0 +1,15 @@
...
+  return fetch(API_URL_TODOS)
+    .then(response => response.json());
+}
+
+export function getUsers(userId: number) {
+  return fetch(`${API_URL_USERS}`)
+    .then(response => response.json())
+    .then(users => {
+      return users.find((user: { id: number; }) => user.id === userId);
+    });
```
aliceblunt8, line 14: now you fetch ALL users and then use find to find a single one
is it effective to fetch 1000000000 users each time, for instance, in order to show information of just ONE of them?

let's use a dedicated endpoint to fetch a single user information by it's id
> CurrentUser component receives userId as a prop and loads user details from [GET user endpoint](https://mate.academy/students-api/users/1) (replace 1 with a given userId).

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,127 @@
...
+
+  const todosInputFilter = (arr: Todo[], titleValue: string) => {
+    if (!titleValue) {
+      return arr;
+    }
+
+    return arr.filter(todo => todo.title.includes(titleValue));
+  };
+
+  const preparedTodosArr = () => {
```
aliceblunt8, line 38:
```suggestion
  const getFilteredTodos = () => {
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,127 @@
...
+    const copyTodos = [...todos];
```
aliceblunt8, line 39: no need to make a copy, filter method doesn't mutate an original array

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,127 @@
...
+};
+
+export const TodoList: React.FC<Props> = ({
+  todos,
+  onClickSelectUser,
+}) => {
+  const [inputValue, setInputValue] = useState('');
+  const [selectValue, setSelectValue] = useState('all');
+
+  const todosSelectArrFilter = (arr: Todo[], completedValue: string) => {
```
aliceblunt8, line 17: no need to pass these arguments, just use them inside, cause this function has access to them anyway

```suggestion
  const todosSelectArrFilter = () => {
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,127 @@
...
+
+    return arr.filter(todo => todo.title.includes(titleValue));
+  };
+
+  const preparedTodosArr = () => {
+    const copyTodos = [...todos];
+
+    const filterByInput = todosInputFilter(copyTodos, inputValue);
+
+    const filterBySelect = todosSelectArrFilter(filterByInput, selectValue);
```
aliceblunt8, line 43: no need for these separate functions, just add switch case here, then filter by title and then just return a result

aliceblunt8 (changes requested):

ArtemTopolskyi (approved):

PR #798, opened 2022-04-21
 react_dynamic-list-of-todos

IvanShulhan:
[DEMO LINK](https://ivanshulhan.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,25 +1,98 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [value, setValue] = useState('');
+  const [sortBy, setSortBy] = useState('');
+  const [randomSort, setRandomSort] = useState(false);
+
+  useEffect(() => {
+    getTodos()
+      .then(res => {
+        setTodos(res);
+      });
```
aliceblunt8, line 22:
```suggestion
      .then(setTodos);
```

aliceblunt8: magic ;)

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,117 @@
...
-          </button>
-        </li>
-      </ul>
-    </div>
-  </div>
+        </div>
+
+        <div className="TodoList__list-container">
+          {todos.length > 0 ? (
+            <ul className="TodoList__list">
```
aliceblunt8, line 70: TodoList component should contain only the list itself and nothing more, move the form to App component

aliceblunt8 (changes requested):

src/App.tsx
IvanShulhan: )

src/components/TodoList/TodoList.tsx
IvanShulhan: Ok, don't prfoblem, but in tasks was something like this:
image

Bohdan-Maksymiuk (approved):

PR #845, opened 2022-05-25
solution

plankton9991:
[DEMO LINK](https://plankton9991.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,44 +1,103 @@
-import React from 'react';
+import React, { useState, Dispatch, useEffect } from 'react';
 import './TodoList.scss';
 
-export const TodoList: React.FC = () => (
-  <div className="TodoList">
-    <h2>Todos:</h2>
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,103 @@
...
+        case 'not completed':
+          return titleToLower.includes(filterByTitleToLower)
+          && todo.completed === false;
+
+        case 'completed':
+          return titleToLower.includes(filterByTitleToLower)
+          && todo.completed === true;
+
+        default:
+          return todo;
```
misha200119, line 35: better just throw exception there smth like "not existing type of filter exception"

misha200119 (approved):

AlinaMatuschak (approved):
- the bad name for commit (should describe what you have done in this commit)
Screenshot 2022-05-26 at 11 24 30
- should create more than one commit per task
- don’t need to render user button if todo doesn’t have user id

PR #858, opened 2022-06-13
Solution

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,44 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import './styles/general.scss';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
+import { getTodos } from './components/API/api';
 
 const App: React.FC = () => {
   const [
     selectedUserId,
-    // setSelectedUserId,
+    setSelectedUserId,
   ] = useState(0);
+
+  const [currentTodos, setCurrentTodos] = useState<Todo[]>([]);
+
+  async function todosGetter() {
+    const result = await getTodos();
+
+    setCurrentTodos(result);
+  }
+
+  useEffect(() => {
+    todosGetter();
+  }, [selectedUserId]);
 
   return (
     <div className="App">
       <div className="App__sidebar">
-        <TodoList />
+        <TodoList
+          listOfTodos={currentTodos}
+          callbackForUserSelect={setSelectedUserId}
+        />
       </div>
 
       <div className="App__content">
         <div className="App__content-container">
           {selectedUserId ? (
-            <CurrentUser />
+            <CurrentUser
+              userId={selectedUserId}
+              clearUser={setSelectedUserId}
+            />
           ) : 'No user selected'}
         </div>
       </div>
--- a/src/components/API/api.ts
+++ b/src/components/API/api.ts
@@ -1,0 +1,18 @@
+const BASE_URL = 'https://mate.academy/students-api';
+
+export function request(url : string) {
+  return fetch(`${BASE_URL}${url}`)
+    .then(res => res.json());
+}
+
+export async function getTodos() {
+  const result = await (request('/todos'));
+
+  return result;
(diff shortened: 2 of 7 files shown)


src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,42 @@
...
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+export const CurrentUser: React.FC<Props> = ({ userId, clearUser }) => {
+  const [currentUser, setCurrentUser] = useState<User>();
+
+  async function singleUser(id : number) {
+    const result = await getUsers();
+
+    setCurrentUser(result.find((el : User) => el.id === id));
```
KostiaO, line 16: you should do request with current userId url/users/${userId},

KostiaO (changes requested):

src/App.tsx
```diff
@@ -1,25 +1,44 @@
...
+
+  async function todosGetter() {
+    const result = await getTodos();
+
+    setCurrentTodos(result);
+  }
+
+  useEffect(() => {
+    todosGetter();
+  }, [selectedUserId]);
```
KostiaO, line 24: there is no need in this dependence, you will do requests to todos on every peek of userId

EnjSadman: There is need. Todos getter updates state. So it's infinite loop without it

Anna-Tkachenko: You don't need to load todos every time when selectedUserId change
You should load them only once, when your component mounted on the page (use componentDidMount method)
If you use hook, so you should pass an empty array as a second argument to make this functionality the same as for componentDidMount method

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,109 @@
...
+            active
+          </option>
+        </select>
+      </label>
+
+      <h2>Todos:</h2>
+
+      <div className="TodoList__list-container">
+        <ul className="TodoList__list">
+          {sorter(listOfTodos).map(singleTodo => (
```
Anna-Tkachenko, line 69: Will be better to create separate variable before the return method where you can store the result of sorter function
And then use this variable in jsx

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,109 @@
...
+            singleTodo.title.includes(currentFilter) && (
```
Anna-Tkachenko, line 70: You shouldn't check it each time for each todo
Call filter with needed callback and store the result in some variable before the return statement

Anna-Tkachenko (changes requested):

Denchepornyuk:
Screenshot 2022-06-20 at 11 57 27

user should be selected only when someone clicked on the user

src/components/API/api.ts
```diff
@@ -0,0 +1,24 @@
+const BASE_URL = 'https://mate.academy/students-api';
+
+export function request(url : string) {
```
Denchepornyuk, line 3:
```suggestion
export function request(url: string) {
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,121 @@
...
+      }
+
+      default: {
+        return list;
+      }
+    }
+  }
+
+  const filtrato = (option : string) => {
+    const sorted = listOfTodos.filter(el => el.title.includes(option));
```
Denchepornyuk, line 34: don't forget to handle lower and upper cases

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,121 @@
...
+        return list.filter(el => el.completed === false);
+      }
+
+      default: {
+        return list;
+      }
+    }
+  }
+
+  const filtrato = (option : string) => {
```
Denchepornyuk, line 33: interesting name for the function

Denchepornyuk (approved):

PR #861, opened 2022-06-15
list todo

Andriyk0:
[DEMO LINK](https://Andriyk0.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import React, { useState } from 'react';
 import './App.scss';
 import './styles/general.scss';
@@ -5,21 +6,21 @@ import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
 
 const App: React.FC = () => {
-  const [
-    selectedUserId,
-    // setSelectedUserId,
-  ] = useState(0);
+  const [selectedUserId, setSelectedUserId] = useState(0);
 
   return (
     <div className="App">
       <div className="App__sidebar">
-        <TodoList />
+        <TodoList selectUserId={setSelectedUserId} />
       </div>
 
       <div className="App__content">
         <div className="App__content-container">
           {selectedUserId ? (
-            <CurrentUser />
+            <CurrentUser
+              userId={selectedUserId}
+              setSelectedUserId={setSelectedUserId}
+            />
           ) : 'No user selected'}
         </div>
       </div>
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -1,0 +1,13 @@
+/* eslint-disable no-console */
+const API_URL_USERS = 'https://mate.academy/students-api/users';
+const API_URL_TODOS = 'https://mate.academy/students-api/todos';
+
+export function getUsers(userId:number) {
+  return fetch(`${API_URL_USERS}/${userId}`)
+    .then(response => response.json());
+}
+
+export function getTodos() {
+  return fetch(API_URL_TODOS)
+    .then(response => response.json());
+}
--- a/src/components/CurrentUser/CurrentUser.tsx
+++ b/src/components/CurrentUser/CurrentUser.tsx
@@ -1,12 +1,39 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { getUsers } from '../../api/api';
 import './CurrentUser.scss';
 
-export const CurrentUser: React.FC = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+interface Prop {
+  userId: number,
+  setSelectedUserId: React.Dispatch<React.SetStateAction<number>>,
+}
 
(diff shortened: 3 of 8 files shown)


src/App.tsx
```diff
@@ -1,25 +1,26 @@
...
+        <TodoList selectUserId={setSelectedUserId} />
       </div>
 
       <div className="App__content">
         <div className="App__content-container">
           {selectedUserId ? (
-            <CurrentUser />
+            <CurrentUser
+              userId={selectedUserId}
+              setSelectedUserId={setSelectedUserId}
```
SavMary, line 22: It will be better, if you create callback and pass it to component CurrentUser instead state 'setSelectedUserId', because passing state is not a good idea

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,113 @@
...
+                  </label>
+
+                  <button
+                    className="TodoList__user-button button"
+                    type="button"
+                    onClick={() => selectUserId(todo.userId)}
+                  >
+                    User&nbsp;
+                    {todo.userId}
+                  </button>
```
SavMary, line 105: You repeat the layout twice. You could just write a condition when your todo has be checked

SavMary (approved):

src/api/api.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable no-console */
+const API_URL_USERS = 'https://mate.academy/students-api/users';
+const API_URL_TODOS = 'https://mate.academy/students-api/todos';
+
+export function getUsers(userId:number) {
```
Anna-Tkachenko, line 5:
```suggestion
export function getUser(userId:number) {
```

src/components/TodoList/TodoList.tsx
Anna-Tkachenko: Pay attention to @SavMary comments

src/App.tsx
Anna-Tkachenko: Pay attention to @SavMary comments

Anna-Tkachenko (changes requested):

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,39 @@
...
+  setSelectedUserId: (value: number) => void,
+}
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+export const CurrentUser: React.FC<Prop> = ({ userId, setSelectedUserId }) => {
+  const [user, setUser] = useState<User>();
```
```suggestion
  const [user, setUser] = useState<User | null>(null);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,92 @@
...
-        </li>
-      </ul>
+interface Prop {
+  selectUserId: (value: number) => void,
+}
+
+export const TodoList: React.FC<Prop> = ({ selectUserId }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [optionValue, setOptionValue] = useState(' ');
```
Denchepornyuk, line 13: would be great to add types for optionValue, you can create enum with active, completed, all values

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,92 @@
...
+
+  const selectTodos = () => {
+    let rezult;
+
+    if (optionValue === 'active') {
+      rezult = filterTodos.filter(todo => todo.completed === false);
+    } else if (optionValue === 'completed') {
+      rezult = filterTodos.filter(todo => todo.completed === true);
+    } else {
+      rezult = filterTodos;
```
Denchepornyuk, line 30: use switch case here

Denchepornyuk (changes requested):

src/components/TodoList/TodoList.tsx
```diff
@@ -7,10 +8,17 @@ interface Prop {
...
+  All = 'All',
+  active = 'active',
+  completed = 'completed',
+}
+
 export const TodoList: React.FC<Prop> = ({ selectUserId }) => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [query, setQuery] = useState('');
-  const [optionValue, setOptionValue] = useState(' ');
+  const [optionValue, setOptionValue] = useState<TodosFilterValue | string>(TodosFilterValue.checked);
```
Denchepornyuk, line 21: can we remove `| string` here?

src/components/TodoList/TodoList.tsx
```diff
@@ -52,10 +62,10 @@ export const TodoList: React.FC<Prop> = ({ selectUserId }) => {
           setOptionValue(event.target.value);
```
Denchepornyuk, line 62: add `as` for event.target.value and you can remove string from optionValue
```
setOptionValue(event.target.value as TodosFilterValue);
```

Denchepornyuk (approved):

PR #865, opened 2022-06-16
Develop

dellachaise:
[DEMO LINK](https://dellachaise.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,21 +8,89 @@ import { getTodos } from './api';
 const App: React.FC = () => {
   const [selectedUserId, setSelectedUserId] = useState(0);
   const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
 
   const changeUser = (userId: number) => setSelectedUserId(userId);
 
-  useEffect(() => {
+  const getCurrentTodos = () => {
     getTodos()
-      .then((todosFromServer) => {
-        setTodos(todosFromServer);
+      .then((currentTodosFromServer) => {
+        setTodos(currentTodosFromServer);
       });
+  };
+
+  useEffect(() => {
+    getCurrentTodos();
   }, []);
+
+  const selectHandler = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    if (event.target.value === 'all') {
+      getCurrentTodos();
+    }
+
+    if (event.target.value === 'complited') {
+      getTodos()
+        .then((currentTodosFromServer) => {
+          setTodos(
+            currentTodosFromServer
+              .filter((todo: Todo) => todo.completed),
+          );
+        });
+    }
+
+    if (event.target.value === 'uncomplited') {
+      getTodos()
+        .then((currentTodosFromServer) => {
+          setTodos(
+            currentTodosFromServer
+              .filter((todo: Todo) => !todo.completed),
+          );
+        });
+    }
+  };
 
   return (
     <div className="App">
       <div className="App__sidebar">
+        <input
+          type="text"
+          id="search-query"
+          className="App__search"
+          value={query}
+          placeholder="Type search word"
+          onChange={event => setQuery(event.target.value.toLowerCase())}
+          data-cy="filterByTitle"
+        />
+        <select
+          value="all"
+          className="App__select"
+          onChange={selectHandler}
+        >
+          <option
+            value="all"
+            disabled
+          >
+            Choose task status
+          </option>
+          <option
+            value="all"
+          >
+            Show all
+          </option>
+          <option
+            value="complited"
+          >
+            Show complited
+          </option>
+          <option
+            value="uncomplited"
+          >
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,25 +1,101 @@
...
+  const selectHandler = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    if (event.target.value === 'all') {
+      getCurrentTodos();
+    }
+
+    if (event.target.value === 'complited') {
+      getTodos()
+        .then((currentTodosFromServer) => {
+          setTodos(
+            currentTodosFromServer
```
anastasiia-nikita, line 35: I think, you can miss promise 32-33 lines and in setTodos put todos like `setTodos(todos.filter((todo: Todo) => todo.completed)`
Because in 18 line, you have already put todos from server to todos.
it also should be work and write a less code)

src/App.tsx
```diff
@@ -1,25 +1,101 @@
...
+              .filter((todo: Todo) => todo.completed),
+          );
+        });
+    }
+
+    if (event.target.value === 'uncomplited') {
+      getTodos()
+        .then((currentTodosFromServer) => {
+          setTodos(
+            currentTodosFromServer
```
anastasiia-nikita, line 45: and here

anastasiia-nikita (commented):
Work good!

src/App.tsx
dellachaise: Thank you for good marks!

But unfortunately, this way select work unproperly(I tried it yesterday, and today after your review).

The only way to write cleaner, that I see now - it's how you write previously - use switch-case and set state for switch property.

Anya, could you please tell me, is it necessary now - rewrite this way?

src/App.tsx
```diff
@@ -1,25 +1,101 @@
...
+        setTodos(currentTodosFromServer);
+      });
+  };
+
+  useEffect(() => {
+    getCurrentTodos();
+  }, []);
+
+  const selectHandler = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    if (event.target.value === 'all') {
```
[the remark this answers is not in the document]
Denchepornyuk: yep, let's add switch case here

src/App.tsx
```diff
@@ -1,25 +1,101 @@
...
+      getCurrentTodos();
```
Denchepornyuk, line 28: move this function TodoList component and use todos from props instead of getCurrentTodos() or getTodos() functions

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,56 @@
...
+};
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+export const CurrentUser: React.FC<Props>
+= ({ selectedUserId, changeUser }) => {
+  const [selectedUser, setSelectedUser] = useState<User>();
```
Denchepornyuk, line 12: selectedUser can be undefined, 22 line of code
It is better to make type `User | null`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,42 @@
...
-              TodoList__user-button
-              TodoList__user-button--selected
-              button
-            "
-            type="button"
-          >
-            User&nbsp;#1
-          </button>
-        </li>
+    return title.includes(currentQuery);
```
Denchepornyuk, line 21: don't forget to handle lower and upper case

Denchepornyuk (changes requested):

src/App.tsx
```diff
@@ -1,25 +1,40 @@
...
+  const [selectedUserId, setSelectedUserId] = useState(0);
+  const [todos, setTodos] = useState<Todo[]>([]);
+
+  const changeUser = (userId: number) => setSelectedUserId(userId);
+
+  useEffect(() => {
+    getTodos()
+      .then((currentTodosFromServer) => {
+        setTodos(currentTodosFromServer);
+      });
```
Denchepornyuk, line 18: you have to handle errors, if getTodos returns an error you should catch error here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,93 @@
...
+  changeUser,
+  selectedUserId,
+}) => {
+  const [query, setQuery] = useState('');
+  const [filter, setFilter] = useState<FilterOptions>(FilterOptions.all);
+
+  const visibleTodos = todos.filter((todo) => {
+    const { title } = todo;
+
+    return title.toLowerCase().includes(query);
```
Denchepornyuk, line 28: query also should be in lower case

Denchepornyuk (approved):

PR #870, opened 2022-06-16
solution

ZUBOV-ILLIA:
[DEMO LINK](https://ZUBOV-ILLIA.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,35 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import './App.scss';
 import './styles/general.scss';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
+import { getAllTodos } from './api/api';
 
 const App: React.FC = () => {
-  const [
-    selectedUserId,
-    // setSelectedUserId,
-  ] = useState(0);
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [selectedUserId, setSelectedUserId] = useState<number>(0);
+
+  useEffect(() => {
+    getAllTodos()
+      .then(result => setTodosFromServer(result));
+  }, []);
 
   return (
     <div className="App">
       <div className="App__sidebar">
-        <TodoList />
+        <TodoList
+          todosFromServer={todosFromServer}
+          selectUser={setSelectedUserId}
+        />
       </div>
 
       <div className="App__content">
         <div className="App__content-container">
           {selectedUserId ? (
-            <CurrentUser />
+            <CurrentUser
+              selectedUserId={selectedUserId}
+              setSelectedUserId={setSelectedUserId}
+            />
           ) : 'No user selected'}
         </div>
       </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,11 +7,15 @@ import { getAllTodos } from './api/api';
 
 const App: React.FC = () => {
   const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
-  const [selectedUserId, setSelectedUserId] = useState<number>(0);
+  const [selectedUserId, setSelectedUserId] = useState(0);
 
   useEffect(() => {
     getAllTodos()
-      .then(result => setTodosFromServer(result));
+      .then(result => setTodosFromServer(result))
+      .catch(error => {
+        // eslint-disable-next-line no-console
+        console.log(error, 'something wrong here');
+      });
   }, []);
 
   return (
@@ -20,6 +24,7 @@ const App: React.FC = () => {
         <TodoList
           todosFromServer={todosFromServer}
           selectUser={setSelectedUserId}
+          selectedUserId={selectedUserId}
         />
       </div>
 
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -1,0 +1,11 @@
+const API_URL = 'https://mate.academy/students-api';
+
+export function getAllTodos(): Promise<Todo[]> {
+  return fetch(`${API_URL}/todos`)
+    .then(response => response.json());
+}
+
+export function getUser(n: number): Promise<User> {
+  return fetch(`${API_URL}/users/${n}`)
+    .then(response => response.json());
+}
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -5,7 +5,7 @@ export function getAllTodos(): Promise<Todo[]> {
     .then(response => response.json());
 }
 
-export function getUser(n: number): Promise<User> {
-  return fetch(`${API_URL}/users/${n}`)
+export function getUser(userId: number): Promise<User> {
+  return fetch(`${API_URL}/users/${userId}`)
     .then(response => response.json());
 }
--- a/src/components/CurrentUser/CurrentUser.tsx
+++ b/src/components/CurrentUser/CurrentUser.tsx
@@ -13,10 +13,11 @@ export const CurrentUser: React.FC<Props>
 
     useEffect(() => {
       getUser(selectedUserId)
-        .then(result => setUser(result));
-
-      // eslint-disable-next-line no-console
-      console.log('render');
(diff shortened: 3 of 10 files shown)


yelnikov-andrii (approved):

denys-cheporniuk (commented):
Screenshot 2022-06-20 at 13 18 27

Randomize doesn't work

Denchepornyuk:
Screenshot 2022-06-20 at 13 19 21
all buttons with the same user should be selected if one of them selected

src/App.tsx
```diff
@@ -1,25 +1,35 @@
...
 import { CurrentUser } from './components/CurrentUser';
+import { getAllTodos } from './api/api';
 
 const App: React.FC = () => {
-  const [
-    selectedUserId,
-    // setSelectedUserId,
-  ] = useState(0);
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [selectedUserId, setSelectedUserId] = useState<number>(0);
```
Denchepornyuk, line 10:
```suggestion
  const [selectedUserId, setSelectedUserId] = useState(0);
```
the number is a primitive type so you can remove it from useState

src/App.tsx
```diff
@@ -1,25 +1,35 @@
...
+
+  useEffect(() => {
+    getAllTodos()
+      .then(result => setTodosFromServer(result));
```
Denchepornyuk, line 14: you should handle errors here, if getAllTodos function return an error you have to catch this error here

src/api/api.ts
```diff
@@ -0,0 +1,11 @@
+const API_URL = 'https://mate.academy/students-api';
+
+export function getAllTodos(): Promise<Todo[]> {
+  return fetch(`${API_URL}/todos`)
+    .then(response => response.json());
+}
+
+export function getUser(n: number): Promise<User> {
```
Denchepornyuk, line 8: n is not a good name for function argument

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,53 @@
...
+  setSelectedUserId: React.Dispatch<React.SetStateAction<number>>;
+}
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+export const CurrentUser: React.FC<Props>
+  = ({ selectedUserId, setSelectedUserId }) => {
```
Denchepornyuk, line 11:
```suggestion
export const CurrentUser: React.FC<Props> = ({
  selectedUserId, setSelectedUserId,
}) => {
```

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,53 @@
...
+    const [user, setUser] = useState<User>();
```
Denchepornyuk, line 12: user can be null 
```suggestion
    const [user, setUser] = useState<User | null>();
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,114 @@
...
-            "
-            type="button"
+interface Props {
+  todosFromServer: Todo[];
+  selectUser: React.Dispatch<React.SetStateAction<number>>;
+}
+
+export const TodoList: React.FC<Props>
+  = ({ todosFromServer, selectUser }) => {
+    const [filter, setFilter] = useState<string>('');
```
Denchepornyuk, line 12:
```suggestion
    const [filter, setFilter] = useState('');
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,114 @@
...
+    const [selectFilter, setSelectFilter] = useState('all');
```
Denchepornyuk, line 13: would be great to add type for selectFilter, you can create enum

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,114 @@
...
+
+    const filteredByState
+      = todosFromServer.filter(todo => {
+        if (todo.completed === true && selectFilter === 'completed') {
+          return true;
+        }
+
+        if (todo.completed === false && selectFilter === 'active') {
+          return true;
+        }
```
Denchepornyuk, line 23: i belive you can remove `todo.completed === true` and `todo.completed === false`
also, use switch case here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,114 @@
...
-          </button>
-        </li>
-      </ul>
-    </div>
-  </div>
-);
+            {filteredByTitle.map(todo => (
+              <li
+              // className={`TodoList__item TodoList__item--${todo.completed ? 'checked' : 'unchecked'}`}
+              // how can I make it using classNames ? I mean now here have duplicate.
```
Denchepornyuk, line 78: try something like this with classnames
```
  'TodoList__item--unchecked': !todo.completed,
  'TodoList__item--checked': todo.completed,
```

denys-cheporniuk (changes requested):

src/components/TodoList/TodoList.tsx
ZUBOV-ILLIA: image

No unfortunately I can't remove this because without this logic it will show all tasks no matter which option is currently selected.

src/components/TodoList/TodoList.tsx
ZUBOV-ILLIA: image

Yes, its already same in my code but I mean that in classNames I need two times writing 'TodoList__item--' but only ending need change.
In ternar operator its seems like less code ? Or maybe its need to doo just for more understanding code ?

src/components/TodoList/TodoList.tsx
ZUBOV-ILLIA: And in this case I'm not quite sure how to switch case here

src/components/TodoList/TodoList.tsx
Denchepornyuk: In this case, we have almost the same class name `todolist__item-but these are completely different classes, so it is better to separate the logic for them

src/App.tsx
```diff
@@ -1,25 +1,40 @@
...
-  ] = useState(0);
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [selectedUserId, setSelectedUserId] = useState(0);
+
+  useEffect(() => {
+    getAllTodos()
+      .then(result => setTodosFromServer(result))
+      .catch(error => {
+        // eslint-disable-next-line no-console
+        console.log(error, 'something wrong here');
```
Denchepornyuk, line 17: you can move todosFromServer and useEffect to TodoList component

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,116 @@
...
+  all = 'all',
+  completed = 'completed',
+  active = 'active',
+}
+
+export const TodoList: React.FC<Props> = ({
+  todosFromServer, selectUser, selectedUserId,
+}) => {
+  const [filter, setFilter] = useState('');
+  const [selectFilter, setSelectFilter] = useState('all');
```
Denchepornyuk, line 21:
```suggestion
  const [selectFilter, setSelectFilter] = useState<SeeFiltered>('all');
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,116 @@
...
+        onChange={event => {
+          setFilter(event.target.value);
+        }}
+      />
+
+      <div className="select">
+        <select
+          value={selectFilter}
+          onChange={event => {
+            setSelectFilter(event.target.value);
```
Denchepornyuk, line 61:
```suggestion
            setSelectFilter(event.target.value as SeeFiltered);
```

denys-cheporniuk (approved):

PR #876, opened 2022-06-18
React dynamic list of todos

dasha-chekmariova:
[DEMO LINK](https://dasha-chekmariova.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,19 +7,27 @@ import { CurrentUser } from './components/CurrentUser';
 const App: React.FC = () => {
   const [
     selectedUserId,
-    // setSelectedUserId,
+    setSelectedUserId,
   ] = useState(0);
+
+  const userId = (todo: Todo) => (
+    setSelectedUserId(todo.userId)
+  );
+
+  const clear = () => (
+    setSelectedUserId(0)
+  );
 
   return (
     <div className="App">
       <div className="App__sidebar">
-        <TodoList />
+        <TodoList userId={userId} />
       </div>
 
       <div className="App__content">
         <div className="App__content-container">
           {selectedUserId ? (
-            <CurrentUser />
+            <CurrentUser userId={selectedUserId} clear={clear} />
           ) : 'No user selected'}
         </div>
       </div>
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -1,0 +1,13 @@
+const url = 'https://mate.academy/students-api';
+
+export const getTodo = async () => {
+  const response = await fetch(`${url}/todos`);
+
+  return response.json();
+};
+
+export const getUser = async (userId: number) => {
+  const response = await fetch(`${url}/users/${userId}`);
+
+  return response.json();
+};
--- a/src/components/CurrentUser/CurrentUser.tsx
+++ b/src/components/CurrentUser/CurrentUser.tsx
@@ -1,12 +1,49 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './CurrentUser.scss';
+import { getUser } from '../../api/api';
 
-export const CurrentUser: React.FC = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+type Props = {
+  userId: number,
+  clear: () => void,
+};
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+export const CurrentUser: React.FC<Props> = ({ userId, clear }) => {
+  const [selectedUser, setSelectedUser] = useState<User>();
+
+  useEffect(() => {
+    getUser(userId)
+      .then((user: User) => setSelectedUser(user));
+  }, [userId]);
+
+  return (
+    <div className="CurrentUser">
+      {selectedUser
+        ? (
+          <>
+            <h2 className="CurrentUser__title"><span>{`Selected user: ${selectedUser.id}`}</span></h2>
+
+            <button
+              type="button"
+              className="
+              TodoList__user-button
+              TodoList__user-button--selected
+              button
+            "
+              onClick={() => (
+                clear()
+              )}
+            >
+              Clear
+            </button>
+
+            <h3 className="CurrentUser__name">{selectedUser.name}</h3>
+            <p className="CurrentUser__email">{selectedUser.email}</p>
+            <p className="CurrentUser__phone">{selectedUser.phone}</p>
+          </>
+        )
+        : (
+          'Loading...'
+        )}
+    </div>
+  );
+};
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,44 +1,124 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './TodoList.scss';
+import classname from 'classnames';
+import { getTodo } from '../../api/api';
 
-export const TodoList: React.FC = () => (
-  <div className="TodoList">
-    <h2>Todos:</h2>
+type Props = {
+  userId: (todo: Todo) => void,
+};
 
-    <div className="TodoList__list-container">
-      <ul className="TodoList__list">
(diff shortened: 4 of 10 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,124 @@
...
+type Props = {
+  userId: (todo: Todo) => void,
+};
+
+export const TodoList: React.FC<Props> = ({ userId }) => {
+  const [inputValue, setInputValue] = useState('');
+  const [todos, setTodos] = useState<Todo[] | []>([]);
+  const [optionValue, setOptionValue] = useState('123');
+
+  const filterTodo = [...todos]
```
iwanttosmile4u, line 15: You don't need to create array copy in this case, because filter is non-mutating method.

iwanttosmile4u (approved):

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,49 @@
...
+  clear: () => void,
+};
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+export const CurrentUser: React.FC<Props> = ({ userId, clear }) => {
+  const [selectedUser, setSelectedUser] = useState<User>();
```
```suggestion
  const [selectedUser, setSelectedUser] = useState<User | null>(null);
```

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,49 @@
...
+
+  useEffect(() => {
+    getUser(userId)
+      .then((user: User) => setSelectedUser(user));
```
Denchepornyuk, line 15: you have to handle errors, if getUser returns an error you should catch the error here

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,49 @@
...
+  }, [userId]);
+
+  return (
+    <div className="CurrentUser">
+      {selectedUser
+        ? (
+          <>
+            <h2 className="CurrentUser__title"><span>{`Selected user: ${selectedUser.id}`}</span></h2>
```
Denchepornyuk, line 23: don't write code in one line

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,49 @@
...
+            <button
+              type="button"
+              className="
+              TodoList__user-button
+              TodoList__user-button--selected
+              button
+            "
+              onClick={() => (
+                clear()
+              )}
```
Denchepornyuk, line 34:
```suggestion
              onClick={clear}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,124 @@
...
-          <button
-            className="TodoList__user-button button"
-            type="button"
+type Props = {
+  userId: (todo: Todo) => void,
+};
+
+export const TodoList: React.FC<Props> = ({ userId }) => {
+  const [inputValue, setInputValue] = useState('');
+  const [todos, setTodos] = useState<Todo[] | []>([]);
```
Denchepornyuk, line 12:
```suggestion
  const [todos, setTodos] = useState<Todo[]>([]);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,124 @@
...
+  const [optionValue, setOptionValue] = useState('123');
```
Denchepornyuk, line 13: why 123??

src/components/TodoList/TodoList.tsx
Denchepornyuk: agree

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,124 @@
...
+
+  const filterTodo = [...todos]
+    .filter(todo => todo.title.toLocaleLowerCase()
+      .includes(inputValue.toLocaleLowerCase()));
+
+  useEffect(() => {
+    getTodo()
+      .then((todosFromServer: Todo[]) => setTodos(todosFromServer));
```
Denchepornyuk, line 21: you have to handle errors, if getTodo returns an error you should catch the error here

src/api/api.ts
```diff
@@ -0,0 +1,13 @@
+const url = 'https://mate.academy/students-api';
+
+export const getTodo = async () => {
```
Denchepornyuk, line 3:
```suggestion
export const getTodos = async () => {
```

Denchepornyuk, line 3: add return type for functions

Denchepornyuk (changes requested):

src/components/CurrentUser/CurrentUser.tsx
dasha-chekmariova: I'm not sure I know how to do it, but I'll try

src/App.tsx
```diff
@@ -7,19 +7,27 @@ import { CurrentUser } from './components/CurrentUser';
 const App: React.FC = () => {
   const [
     selectedUserId,
-    // setSelectedUserId,
+    setSelectedUserId,
   ] = useState(0);
 
+  const userId = (todo: Todo) => (
+    setSelectedUserId(todo.userId)
+  );
```
Anna-Tkachenko, line 15:
```suggestion
  const changeUserId = (todo: Todo) => (
    setSelectedUserId(todo.userId)
  );
```

Anna-Tkachenko: pay attention to function's name

src/api/api.ts
```diff
@@ -0,0 +1,13 @@
+const url = 'https://mate.academy/students-api';
+
+export const getTodos = async (): Promise<Todo[]> => {
+  const response = await fetch(`${url}/todos`);
+
+  return response.json();
+};
+
+export const getUser = async (userId: number): Promise<User> => {
+  const response = await fetch(`${url}/usersq/${userId}`);
```
Anna-Tkachenko, line 10:
```suggestion
  const response = await fetch(`${url}/users/${userId}`);
```

Anna-Tkachenko: fix url, now I don't see any user info

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,65 @@
...
+  const [selectedUser, setSelectedUser] = useState<User | null>(null);
+  const [hasLoadingError, setHasLoadingError] = useState(true);
+
+  useEffect(() => {
+    try {
+      getUser(userId)
+        .then((user: User) => setSelectedUser(user));
+    } catch (error) {
+      setHasLoadingError(false);
+    }
```
Anna-Tkachenko, line 20: we use try {} catch {} with async await
and catch() after then() with then() functionality, rewrite it

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,135 @@
...
+
+export const TodoList: React.FC<Props> = ({ userId }) => {
+  const [inputValue, setInputValue] = useState('');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [optionValue, setOptionValue] = useState('');
+  const [hasLoadingError, setHasLoadingError] = useState(false);
+
+  const filterTodo = todos
+    .filter(todo => todo.title.toLocaleLowerCase()
+      .includes(inputValue.toLocaleLowerCase()));
```
Anna-Tkachenko, line 18: Now you transform input value to lower case on each iteration
It's redundant here
Will be better to create separate variable before the filter method where you can store the input value in lower case
And use this variable in filter

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,135 @@
...
+
+  useEffect(() => {
+    try {
+      getTodos()
+        .then((todosFromServer: Todo[]) => setTodos(todosFromServer));
+    } catch (error) {
+      setHasLoadingError(true);
+    }
```
Anna-Tkachenko, line 26: Same here with try{} catch{}

try, catch - with async / await
catch() - with then()
.then()
.catch()

Anna-Tkachenko (changes requested):

Denchepornyuk (approved):

PR #880, opened 2022-06-20
solution

alexandra-protyanova:


volodya-leskiv (approved):
Good job!

Denchepornyuk:
Screenshot 2022-06-20 at 16 03 00


 user should be selected only when someone clicked on the user

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,43 @@
...
+  changeUser: (id: number) => void,
+};
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+export const CurrentUser: React.FC<Props> = ({ userId, changeUser }) => {
+  const [selectedUser, setSelectedUser] = useState<User>();
```
```suggestion
  const [selectedUser, setSelectedUser] = useState<User | null>();
```

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,43 @@
...
+
+  useEffect(() => {
+    getUser()
+      .then(allUsers => setSelectedUser((allUsers
+        .find(user => user.id === userId))));
```
Denchepornyuk, line 16: you should handle errors here, if getUsers returns an error you should catch the error here

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,43 @@
...
+  }, [userId]);
+
+  return (
+    <div className="CurrentUser">
+      <h2 className="CurrentUser__title"><span>{`Selected user: ${selectedUser?.id}`}</span></h2>
```
Denchepornyuk, line 21: don't write code in one line

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,116 @@
...
+type Props = {
+  addUserId: (userId: number) => void,
+};
+
+export const TodoList: React.FC<Props> = ({
+  addUserId,
+}) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [input, setInput] = useState('');
+  const [option, setOption] = useState('all');
```
Denchepornyuk, line 15: add type for option, you can create enum

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,116 @@
...
+
+  useEffect(() => {
+    getTodos().then(allTodos => setTodos(allTodos));
```
Denchepornyuk, line 18: you should handle errors here, if getTodos returns an error you should catch the error here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,116 @@
...
+    .filter(todo => todo.title.toLowerCase().includes(input.toLowerCase())
+    && handleOptions(option, todo));
+
+  const randomizer = () => {
+    const copyOfTodos = [...todos];
+
+    for (let i = copyOfTodos.length - 1; i >= 0; i -= 1) {
+      const rev = Math.floor(Math.random() * (i + 1));
+
+      [copyOfTodos[i], copyOfTodos[rev]] = [copyOfTodos[rev], copyOfTodos[i]];
```
Denchepornyuk, line 42: 🔥

Denchepornyuk (changes requested):

src/components/TodoList/TodoList.tsx
```diff
@@ -60,12 +71,12 @@ export const TodoList: React.FC<Props> = ({
         <select
           value={option}
           onChange={(event) => {
-            setOption(event.target.value);
+            setOption(+event.target.value);
```
Denchepornyuk, line 74: do you need `+` here?

Denchepornyuk (approved):

src/components/TodoList/TodoList.tsx
alexandra-protyanova: yes, I do 😅 
I asked you about the problem with typscript, it's not allowing to accept strings, because he is waiting for member of enum. This trick actually helped

PR #898, opened 2022-06-29
solution

LukaszGryziewicz:
[DEMO LINK](https://LukaszGryziewicz.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,25 +1,42 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import './styles/general.scss';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
+import { getTodos } from './api';
 
 const App: React.FC = () => {
-  const [
-    selectedUserId,
-    // setSelectedUserId,
-  ] = useState(0);
+  const [todos, setTodos] = useState<Todo[]>([]);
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,25 +1,42 @@
...
-  ] = useState(0);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedUserId, setSelectedUserId] = useState(0);
+
+  useEffect(() => {
+    (async () => {
+      const loadedTodos = await getTodos();
+
+      setTodos(loadedTodos);
+    })();
```
efdalyalcin, line 17: I would prefer writing a handler for this function, it is hard read what's going on here.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,83 @@
...
+        <option value="completed">Completed</option>
+        <option value="active">Active</option>
+      </select>
+
+      <div className="TodoList__list-container">
+        <ul className="TodoList__list">
+          {filteredTodos.map(todo => (
+            <li
+              key={todo.id}
+              className="TodoList__item TodoList__item--unchecked"
```
efdalyalcin, line 54: I think you need a conditional `TodoList__item--unchecked || --checked` here so that completed task are now red.

efdalyalcin (changes requested):
It is almost done, just 2 little comments.

PR #901, opened 2022-06-29
solution

iLebo-stack:
[DEMO LINK](https://iLebo-stack.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,44 +1,125 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { getTodos } from '../../api\'s/api';
 import './TodoList.scss';
 
-export const TodoList: React.FC = () => (
-  <div className="TodoList">
-    <h2>Todos:</h2>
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,125 @@
...
+        </select>
+      </div>
+
+      <br />
+      <br />
+
+      <div className="TodoList__list-container">
+        <ul className="TodoList__list">
+          {visibleTodos.map(({ id, title, userId }) => (
+            <li className="TodoList__item TodoList__item--unchecked" key={id}>
```
59MrRobot, line 85: The unchecked modifier should only be on the todos that are not completed so the the complete ones are green

59MrRobot (changes requested):
Good job! Just one suggestion regarding the todos background colors

59MrRobot (approved):
Nice!

meforoma (approved):
approve, but few suggestions are worth consideration.
Apply in spare time.

PR #904, opened 2022-07-01
Develop

MabuzaM:
[DEMO LINK](https://MabuzaM.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useState } from 'react';
 import './App.scss';
 import './styles/general.scss';
 import { Todo } from './react-app-env';
-import { getAllTodos } from './apis/api';
+import { getAllTodos, getUser } from './apis/api';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
 
@@ -26,14 +26,18 @@ const App: React.FC = () => {
   );
 
   const loadSelectedUser = useCallback(
-    (newSelectedUserId: number) => {
+    async (newSelectedUserId: number) => {
       try {
-        setSelectedUserId(newSelectedUserId);
+        const checkUser = await getUser(newSelectedUserId);
+
+        if (checkUser?.id) {
+          setSelectedUserId(newSelectedUserId);
+        }
       } catch {
         setSelectedUserId(0);
       }
     },
-    [selectedUserId],
+    [],
   );
 
   const searchTodos = async (criteria: string) => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,25 +1,93 @@
...
+    },
+    [],
+  );
+
+  const loadSelectedUser = useCallback(
+    (newSelectedUserId: number) => {
+      setSelectedUserId(newSelectedUserId);
+    },
+    [selectedUserId],
+  );
```
efdalyalcin, line 33: if there is no user info your apps selection part collapses. You should put a try catch here. You can try the user #17 to see.

src/App.tsx
```diff
@@ -1,25 +1,93 @@
...
+        setTodos(activeTodos);
+        break;
+
+      case 'completed':
+        completedtodos = (await loadTodos()).filter(todo => todo.completed);
+        setTodos(completedtodos);
+        break;
+
+      default:
+        throw Error('Something went wrong');
```
efdalyalcin, line 68: There is no way this default activates but still I think default case shouldn't break the app with an error

src/apis/api.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const BASE_URL = 'https://mate.academy/students-api';
+
+export const getAllTodos = (): Promise<Todo[]> => (
+  fetch(`${BASE_URL}/todos`)
+    .then(response => response.json())
+);
+
+export const getUser = async (id: number): Promise<User> => {
+  const userFromServer = await fetch(`${BASE_URL}/users/${id.toString()}`)
```
efdalyalcin, line 11: no need for toString() method. It will convert into string anyway

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,65 @@
...
+
+      setSelectedUser(await getUser(userId));
+    }, [selectedUser],
+  );
+
+  useEffect(
+    () => {
+      LoadUser();
+
+      // return () => setSelectedUser(null);
```
efdalyalcin, line 29: you should clear the comments on the final submit

efdalyalcin (changes requested):
Almost done.. There are some minor changes you need to make.

src/apis/api.ts
MabuzaM: Fixed that

src/App.tsx
MabuzaM: Fixed that thanx

meforoma (approved):
👍

PR #905, opened 2022-07-01
Develop

andreisucman:


efdalyalcin (approved):
It looks well done!!

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,117 @@
...
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+
+  const [title, setTitle] = useState('');
+  const [option, setOption] = useState('all');
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+
+  useEffect(
+    () => {
+      setFilteredTodos(todoList.filter(todo => {
+        if (todo.title.includes(title)) {
```
meforoma, line 25: would be cool if search is not case-sensitive 🙂 
... `toLowerCase`

meforoma (approved):

PR #908, opened 2022-07-05
Dynamic list of todos

proznerka:
[DEMO LINK](https://proznerka.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,56 @@
-import React, { useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import './App.scss';
 import './styles/general.scss';
 import { TodoList } from './components/TodoList';
 import { CurrentUser } from './components/CurrentUser';
+import { getTodos } from './helpers/api';
 
 const App: React.FC = () => {
   const [
     selectedUserId,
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,26 +1,56 @@
...
-    // setSelectedUserId,
+    setSelectedUserId,
   ] = useState(0);
 
+  const [todos, setTodos] = useState<Todo[]>([]);
+
+  const selectedUserChangeHandler = useCallback(
+    (newSelectedUserId: number) => setSelectedUserId(newSelectedUserId),
+    [],
+  );
```
ihor-jpeg, line 19: but why? you only have to use useCallback with custom callbacks, no need to do it got functions that are returned by useState

src/App.tsx
```diff
@@ -1,26 +1,56 @@
...
+
+  const loadTodos = useCallback(
+    async () => {
+      const todoList = await getTodos();
+
+      setTodos(todoList);
+    },
+    [],
```
ihor-jpeg, line 27: maybe you want a different callback when setTodos changes?

ihor-jpeg (approved):

PR #913, opened 2022-07-16
React dynamic list of todos

vsupruniuk:
[DEMO LINK](https://vsupruniuk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/CurrentUser/CurrentUser.tsx
+++ b/src/components/CurrentUser/CurrentUser.tsx
@@ -9,8 +9,13 @@ type Props = {
   onUserIdChange: (userId: number) => void,
 };
 
-export const CurrentUser: React.FC<Props> = ({ userId, onUserIdChange }) => {
-  const [selectedUser, setSelectedUser] = useState<User | null>(null);
+export const CurrentUser: React.FC<Props> = React.memo(({
+  userId, onUserIdChange,
+}) => {
+  const [
+    selectedUser,
+    setSelectedUser,
+  ] = useState<User | null>(null);
 
   useEffect(() => {
     getUser(userId).then(user => setSelectedUser(user));
@@ -26,7 +31,9 @@ export const CurrentUser: React.FC<Props> = ({ userId, onUserIdChange }) => {
         data-cy="userName"
         className="CurrentUser__name"
       >
-        {selectedUser?.name}
+        {selectedUser && (
+          selectedUser.name
+        )}
       </h3>
 
       <p className="CurrentUser__email">{selectedUser?.email}</p>
@@ -43,4 +50,4 @@ export const CurrentUser: React.FC<Props> = ({ userId, onUserIdChange }) => {
       </button>
     </div>
   );
-};
+});
--- a/src/components/CurrentUser/CurrentUser.tsx
+++ b/src/components/CurrentUser/CurrentUser.tsx
@@ -10,7 +10,8 @@ type Props = {
 };
 
 export const CurrentUser: React.FC<Props> = React.memo(({
-  userId, onUserIdChange,
+  userId,
+  onUserIdChange,
 }) => {
   const [
     selectedUser,
--- a/src/components/CurrentUser/CurrentUser.tsx
+++ b/src/components/CurrentUser/CurrentUser.tsx
@@ -1,12 +1,46 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { getUser } from '../../api/api';
+import { User } from '../../react-app-env';
+
 import './CurrentUser.scss';
 
-export const CurrentUser: React.FC = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+type Props = {
+  userId: number,
+  onUserIdChange: (userId: number) => void,
+};
 
-    <h3 className="CurrentUser__name">Ervin Howell</h3>
-    <p className="CurrentUser__email">Shanna@melissa.tv</p>
-    <p className="CurrentUser__phone">010-692-6593 x09125</p>
-  </div>
-);
+export const CurrentUser: React.FC<Props> = ({ userId, onUserIdChange }) => {
+  const [selectedUser, setSelectedUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    getUser(userId).then(user => setSelectedUser(user));
+  }, [userId]);
+
+  return (
+    <div className="CurrentUser">
+      <h2 className="CurrentUser__title">
+        <span>{`Selected user: ${selectedUser?.id}`}</span>
+      </h2>
+
+      <h3
+        data-cy="userName"
+        className="CurrentUser__name"
+      >
+        {selectedUser?.name}
+      </h3>
+
+      <p className="CurrentUser__email">{selectedUser?.email}</p>
+      <p className="CurrentUser__phone">{selectedUser?.phone}</p>
+
+      <button
+        className="btn btn-danger"
+        type="button"
+        onClick={() => {
+          onUserIdChange(0);
+        }}
+      >
+        Clear
+      </button>
+    </div>
+  );
+};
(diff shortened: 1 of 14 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,149 @@
...
-            User&nbsp;#2
-          </button>
-        </li>
-      </ul>
+type Props = {
+  onUserIdChange: (userId: number) => void,
+};
+
+export const TodoList: React.FC<Props> = ({ onUserIdChange }) => {
+  const [todos, setTodos] = useState<Todo[] | []>([]);
```
jstmpelowycz, line 13: suggestion: if you add `...| []` just to make it a default value, you'd rather implement it the following way
```suggestion
  const [todos, setTodos] = useState<Todo[] | null>(null);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,149 @@
...
+
+  const handleQueryChange = (
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    setQuery(event.target.value);
+  };
+
+  const handleFilterChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setFilterByStatus(event.target.value);
+  };
```
jstmpelowycz, line 29: apply `useCallback` to avoid extra re-rendering

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,149 @@
...
+  //     currentIndex -= 1;
+
+  //     [randomizerArray[currentIndex], randomizerArray[randomIndex]]
+  //     = [randomizerArray[randomIndex], randomizerArray[currentIndex]];
+  //   }
+
+  //   return randomizerArray;
+  // };
+
+  let visibleTodos = todos;
```
jstmpelowycz, line 47: don't leave such variables this way, either wrap them or move them to state

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,149 @@
...
+                  TodoList__user-button
+                  TodoList__user-button--selected
+                  button
+                "
+                type="button"
+                onClick={() => {
+                  if (todo.userId) {
+                    onUserIdChange(todo.userId);
+                  }
+                }}
```
jstmpelowycz, line 139: create a separate function

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,149 @@
...
-          >
-            User&nbsp;#2
-          </button>
-        </li>
-      </ul>
+type Props = {
+  onUserIdChange: (userId: number) => void,
+};
+
+export const TodoList: React.FC<Props> = ({ onUserIdChange }) => {
```
jstmpelowycz, line 12: `memo`?

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,46 @@
...
-export const CurrentUser: React.FC = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+type Props = {
+  userId: number,
+  onUserIdChange: (userId: number) => void,
+};
+
+export const CurrentUser: React.FC<Props> = ({ userId, onUserIdChange }) => {
+  const [selectedUser, setSelectedUser] = useState<User | null>(null);
```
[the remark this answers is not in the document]
jstmpelowycz: also,`memo`?

jstmpelowycz (commented):
good job
but – you've already learned more about advanced hooks, thus consider applying them there needed

jstmpelowycz (changes requested):
needs a few fixes

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,168 @@
...
+                  button
+                "
+                type="button"
+                onClick={() => {
+                  handleUserIdChange(todo);
+                }}
+              >
+                {`User #${todo.userId}`}
+              </button>
+            </li>
```
illiarerun, line 162: `TodoInfo` component?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,168 @@
...
+    } catch (error) {
+      throw new Error('Failed to load todos from server');
+    }
+  }, []);
+
+  const onFilteringByQuery = () => {
+    setVisibleTodos([...visibleTodos].filter(
+      todo => todo.title.toLowerCase().includes(query.toLowerCase()),
+    ));
+  };
```
illiarerun, line 37: move this to `handleQueryChange`, you don't need 2 calls in JSX handler

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,168 @@
...
+        break;
+
+      case 'completed':
+        setVisibleTodos([...todos].filter(todo => todo.completed));
+        break;
+
+      default:
+        setVisibleTodos([...todos]);
+    }
+  };
```
illiarerun, line 52: Move this to handleFilterChange

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,168 @@
...
+
+  const onFilteringByQuery = () => {
+    setVisibleTodos([...visibleTodos].filter(
+      todo => todo.title.toLowerCase().includes(query.toLowerCase()),
+    ));
+  };
+
+  const onFilteringByStatus = (sortBy: string) => {
+    switch (sortBy) {
+      case 'active':
```
illiarerun, line 41: Use typescript, create `enum` for these values

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,168 @@
...
+          aria-label="Default select example"
+          value={filterByStatus}
+          onChange={(event) => {
+            handleFilterChange(event);
+            onFilteringByStatus(event.target.value);
+          }}
+        >
+          <option value="all">All</option>
+          <option value="active">Active</option>
+          <option value="completed">Completed</option>
```
illiarerun, line 116: So you can map your enum here

illiarerun (changes requested):
Your search doesn't work
image
Randomize button also doesn't work, try to deploy again

src/components/TodoList/TodoList.tsx
vsupruniuk: This approach causes problems. We try to filter the array by the new state before it is updated. So I combined both filtering methods inside a `useEffect `hook

src/components/TodoList/TodoList.tsx
vsupruniuk: This approach causes problems. We try to filter the array by the new state before it is updated. So I combined both filtering methods inside a `useEffect `hook

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,55 @@
...
+  handleUserIdChange: (todo: Todo) => void,
+};
+
+export const TodoInfo: React.FC<Props> = React.memo(({
+  visibleTodos,
+  handleUserIdChange,
+}) => {
+  return (
+    <ul data-cy="listOfTodos" className="TodoList__list">
+      {visibleTodos?.map(todo => (
```
illiarerun, line 17: No this should be in `TodoList`  component

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,147 @@
...
+          className="form-select"
+          aria-label="Default select example"
+          value={filterByStatus}
+          onChange={(event) => {
+            handleFilterChange(event);
+          }}
+        >
+          <option value={SortType.ALL}>All</option>
+          <option value={SortType.ACTIVE}>Active</option>
+          <option value={SortType.COMPLETED}>Completed</option>
```
illiarerun, line 128: Map it using `Object.keys`

illiarerun (approved):

PR #916, opened 2022-07-24
Solution

naziamoff:
[DEMO LINK](https://naziamoff.github.io/react_dynamic-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,130 @@
...
-            className="TodoList__user-button button"
-            type="button"
-          >
-            User&nbsp;#2
-          </button>
-        </li>
-      </ul>
+export const TodoList: React.FC<TodoListProps> = ({ todos, onUserSelect }) => {
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+  const [filteredTodos, setFilteredTodos] = useState(todos);
```
Bohdan-Maksymiuk, line 12: why do you need save 2 todo arrays in state?
just prepare filtered before rendering them

src/components/TodoList/TodoList.tsx
```diff
@@ -1,44 +1,130 @@
...
+  const [filterBy, setFilterBy] = useState('all');
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
```
Bohdan-Maksymiuk, line 15: same, why do you need 2 queries in state?

Bohdan-Maksymiuk (changes requested):
> Use the input in the TodoFilter to filter the todos by title;
show the x button when the query is entered;
the x button should clear the query and reset the todos;

I don't see `x button` in your solution

* It would be nice to highlight button with selected user

https://user-images.githubusercontent.com/94289038/181254955-417e2400-3ac7-4139-8223-73917597f66a.mov

* And change styles on checkbox checking/unchecking (optional)

https://user-images.githubusercontent.com/94289038/181255121-b17e4292-68b4-4a40-9e67-e454cebbf58c.mov

src/components/TodoList/TodoList.tsx
naziamoff: That is for search optimization. Query hook is for controlling input, and appliedQuery is debounced. So it sends a request to the server not after each letter but only if user hasn't entered anything for 0.5sec

Will this do or should it be somehow rewritten?

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -1,12 +1,46 @@
...
+import { User, CurrentUserProps } from '../../types';
+import { getUser } from '../../api';
 import './CurrentUser.scss';
 
-export const CurrentUser: React.FC = () => (
-  <div className="CurrentUser">
-    <h2 className="CurrentUser__title"><span>Selected user: 2</span></h2>
+// eslint-disable-next-line max-len
+export const CurrentUser: React.FC<CurrentUserProps> = ({ userId, onUnselect }) => {
+  const [selectedUser, setSelectedUser] = useState<User | null>(null);
```
ArturErmolenko, line 8:
```suggestion
// eslint-disable-next-line max-len
export const CurrentUser: React.FC<CurrentUserProps> = ({ 
userId,
onUnselect,
}) => {
  const [selectedUser, setSelectedUser] = useState<User | null>(null);
```

ArturErmolenko (approved):

PR #926, opened 2022-07-26
my solution

Pavlo000:
[DEMO LINK](https://Pavlo000.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,6 +9,7 @@ import { getTodos } from './api';
 import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Condition } from './types/Condition';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+      if (condition === 'active') {
+        return todo.title.includes(value) && !todo.completed;
+      }
+
+      if (condition === 'completed') {
+        return todo.title.includes(value) && todo.completed;
+      }
+
+      return todo.title.includes(value);
+    });
```
jstmpelowycz, line 49: here is a better solution – switch-case + enum

jstmpelowycz (changes requested):
good joob 🔥 a few fixes – and you're done

jstmpelowycz (approved):

PR #933, opened 2022-07-26
Solution

GavrilyukYura:
[DEMO LINK](https://GavrilyukYura.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,48 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { TodoModal } from './components/TodoModal';
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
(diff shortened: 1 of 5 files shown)


IvanGrabovsky (approved):
good job!

src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
+import { TodoModal } from './components/TodoModal';
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [visibleTodos, setVisibleTotos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    const loading = async () => {
```
Olexandr6, line 20:
```suggestion
    const loadTodos = async () => {
```

src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
+      setIsLoading(true);
+    };
+
+    loading();
+  }, []);
+
+  const filtredTodos = (value: string, status: string) => {
+    const todos = todosFromServer.filter(todo => {
+      switch (status) {
+        case 'completed':
```
Olexandr6, line 32: Would be good to use `enum`

src/App.tsx
```diff
@@ -17,18 +49,30 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter filtredTodos={filtredTodos} />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!isLoading
```
Olexandr6, line 56: Why `!IsLoading` here? This is confusing because when `isLoading` === `true` you render the `TodoList` when it should be the other way around.

Olexandr6 (approved):
Good job👍

PR #966, opened 2022-07-29
v1 was added

Bielichenko:
[DEMO LINK](https://Bielichenko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,90 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { useEffect, useState } from 'react';
+import { getTodos } from '../../api';
+import { Todo } from '../../types/Todo';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  setTodos: (todos: Todo[]) => void;
+};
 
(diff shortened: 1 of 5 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,90 @@
...
+    getTodos().then(setInitialTodos);
+  }, []);
+
+  const selectHandler = (selectionKind: string) => {
+    switch (selectionKind) {
+      case 'all':
+        setTodos([...initialTodos]);
+        break;
+      case 'active':
+        setTodos([...initialTodos].filter(todo => todo.completed === false));
```
jstmpelowycz, line 24: does `filter` mutate original array?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,96 @@
...
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+          </th>
+          <th>Title</th>
+          <th> </th>
```
jstmpelowycz, line 26: what is it for?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,90 @@
...
+        break;
+      case 'completed':
+        setTodos([...initialTodos]
+          .filter(todo => todo.completed === true));
+        break;
+      default:
+        break;
+    }
+  };
```
jstmpelowycz, line 33: you could implement it the following way
```
const todosToSet = initialTodos.filter(todo => {
  switch (...) {
    case ...:
      return ...; // boolean
  }
});

setTodos(todosToSet);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,90 @@
...
+  const [initialTodos, setInitialTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState<string>('');
+
+  useEffect(() => {
+    getTodos().then(setInitialTodos);
+  }, []);
+
+  const selectHandler = (selectionKind: string) => {
+    switch (selectionKind) {
+      case 'all':
```
jstmpelowycz, line 20: use enum

jstmpelowycz (changes requested):

PR #972, opened 2022-08-07
solution

nastia-sydorchuk:
[DEMO LINK](https://nastia-sydorchuk.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
```
ArtemTopolskyi, line 18: you don't need to dublicate data in states. You should only save all todos => then calculate visible todos dynamically

src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [shownTodo, setShownTodo] = useState<number>(0);
```
ArtemTopolskyi, line 20: don't use 0, as empty todo id, use null

src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+  const [selectValue, setSelectValue] = useState<string>('all');
```
ArtemTopolskyi, line 21: use enum for filters

ArtemTopolskyi (changes requested):

ArturErmolenko (approved):
Well done 🔥

PR #977, opened 2022-08-12
Solution task React dynamic list of TODOs

Tania-Troshchuk:
[DEMO LINK](https://tania-troshchuk.github.io/react_dynamic-list-of-todos/)


Tania-Troshchuk:
Hie, when I have been doing this task, i haven`t fantasy to name value, then I pass as props in component ))
```
setFilter={setFilter}
filter={filter} 
```
It`s ok to name values that way or better use other name?

ArturErmolenko:
> Hie, when I have been doing this task, i haven`t fantasy to name value, then I pass as props in component ))
> 
> ```
> setFilter={setFilter}
> filter={filter} 
> ```
> 
> It`s ok to name values that way or better use other name?

Its Ok 👍

ArturErmolenko (approved):
Great job 👍

PR #981, opened 2022-08-18
solution

oleksandr-kotliarov:
[DEMO LINK](https://oleksandr-kotliarov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
 /* eslint-disable max-len */
-import React from 'react';
+import {
+  FC,
+  useState,
+  useEffect,
+  useCallback,
+  useMemo,
+} from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+
+  const visibleTodos = useMemo(() => (
+    todos.filter(todo => {
+      const queryFilter = todo.title.toLowerCase().includes(query.toLowerCase());
+
+      switch (filterBy) {
+        case 'active':
+          return queryFilter && todo.completed === false;
+
+        case 'completed':
```
BudnikOleksii, line 42: You can use enum for switch, but it works too

BudnikOleksii (approved):
Good job! 
You should hide the clear button when the query is clear, check tests😉
2) should hide clear button when seach query becomes empty

src/App.tsx
VETALEK: agree, better use enum

VETALEK (changes requested):
GJ 👍 
commented

VETALEK (approved):
🚀

PR #999, opened 2022-08-23
Solution: 90%

rodiontseva:
[DEMO LINK](https://rodiontseva.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,100 +1,92 @@
-import React from 'react';
+import classNames from 'classnames';
+import { FC } from 'react';
+import { Todo } from '../../types/Todo';
 
-export const TodoList: React.FC = () => (
-  <table className="table is-narrow is-fullwidth">
-    <thead>
-      <tr>
-        <th>#</th>
-        <th>
-          <span className="icon">
-            <i className="fas fa-check" />
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+interface Props {
+  todos: Todo[],
+  setSelectedTodo: (arg0: Todo) => void,
+  openedTodoModal: boolean,
+  setOpenedTodoModal: (arg0: boolean) => void,
+  setLoading: (arg0: boolean) => void,
+}
 
-    <tbody>
-      <tr data-cy="todo" className="">
-        <td className="is-vcentered">1</td>
-        <td className="is-vcentered" />
-        <td className="is-vcentered is-expanded">
-          <p className="has-text-danger">delectus aut autem</p>
-        </td>
-        <td className="has-text-right is-vcentered">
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,92 @@
...
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => handleClick(todo)}
+              >
+                <span className="icon">
+                  <i className={
+                    classNames(
+                      'far fa-eye', { 'fa-eye-slash': openedTodoModal },
+                    )
```
Oleksandr-Pustovit, line 81: This condition is same for all todos. This way instead of slashing one eye on the place of opened modal, you will add slash to icons of all todos.
[your solution](https://prnt.sc/flOGksR3kFb4)
[example](https://prnt.sc/bqY_AVN02juF)

Oleksandr-Pustovit (changes requested):
Good job! 👍 
There are sme comments.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,72 @@
...
+  const {
+    selectedTodo, setOpenedTodoModal, loading, setLoading,
+  } = props;
+  const [selectedUser, setSelectedUser] = useState<User>();
+
+  useEffect(() => {
+    getUser(Number(selectedTodo?.userId))
+      .then(setSelectedUser);
+    setLoading(false);
+  }, [selectedTodo]);
```
Oleksandr-Pustovit, line 24: same as previous comment. 
set loading will work as soon as getUser is initialised and wil not wait until it finishes downloading data.
as a result you will ahve modal opened before data is downloaded [there should be user](https://prnt.sc/i1hcuZI07onL)

Oleksandr-Pustovit (changes requested):

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,80 @@
...
-            </p>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                type="button"
+                className="delete"
+                data-cy="modal-close"
+                onClick={() => {
+                  setSelectedTodoId(null);
+                  setOptionForFilter(OptionForFilterTodos.All);
+                }}
```
Oleksandr-Pustovit, line 51: I dont think you have to reset Filtering option here.
```suggestion
                data-cy="modal-close"
                onClick={() => {
                  setSelectedTodoId(null);
                }}
```

Oleksandr-Pustovit (approved):
🔥

Ihor-Karpyn (changes requested):
Good job, let's fix few issues

Ihor-Karpyn (approved):
GJ

PR #1016, opened 2022-08-24
task is ready

opanasiukvlad:
[DEMO](https://opanasiukvlad.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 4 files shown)


ClaudiaKalnitskaya (approved):
Все ок!

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +38,36 @@ export const TodoModal: React.FC = () => {
...
+                ? <strong className="has-text-success">Done</strong>
+                : <strong className="has-text-danger">Planned</strong>}
 
               {' by '}
 
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
+              <a href={user?.email}>
+                {user?.name}
               </a>
```
ivan-kadykalo, line 71: What will be in case when there is no user?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +38,36 @@ export const TodoModal: React.FC = () => {
...
+                showModal(false);
+                selectTodo(null);
+              }}
             />
           </header>
 
           <div className="modal-card-body">
             <p className="block" data-cy="modal-title">
-              quis ut nam facilis et officia qui
+              {selectedTodo?.title}
```
ivan-kadykalo, line 59: What will be in case when there is no selectedTodo?

src/App.tsx
```diff
@@ -17,18 +54,35 @@ export const App: React.FC = () => {
...
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!todosAreLoaded && <Loader />}
+              <TodoList
+                todos={filteredTodos}
+                showModal={setShowModal}
+                selectedTodo={selectedTodo}
+                selectTodo={setSelectedTodo}
+              />
```
ivan-kadykalo, line 72: Probably would be better to show only one of them in the same time 
```suggestion
              {!todosAreLoaded 
                ? <Loader />
                : <TodoList
                  todos={filteredTodos}
                  showModal={setShowModal}
                  selectedTodo={selectedTodo}
                  selectTodo={setSelectedTodo}
                />
              }
              
```

ivan-kadykalo (changes requested):

src/components/TodoModal/TodoModal.tsx
opanasiukvlad: I simplified that part and changed the condition for rendering the whole block to the user as a param.

src/App.tsx
opanasiukvlad: I added the condition, thank you!

ivan-kadykalo (approved):
🚀

PR #1036, opened 2022-09-19
Solution

Dmytro-Matsehora:
[DEMO LINK](https://Dmytro-Matsehora.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,62 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { Todo } from '../../types/Todo';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  todos: Todo[];
+  value: string;
+  setValue: (arg: string) => (string) | void;
+  text: string;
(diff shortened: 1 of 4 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,69 @@ export const TodoList: React.FC = () => (
...
+                    onClick={() => {
+                      selectTodo(id);
+                    }}
+                  >
+                    <span className="icon">
+                      <i className="far fa-eye" />
+                    </span>
+                  </button>
+                )}
+            </td>
```
tiserett, line 90: you should think about simplifying this code, try to return different values from function or set differrent class names according to the (selectedTodoId === id) condition

tiserett (changes requested):
even if the todo is already completed, in modal window text is red, must be green

tiserett (changes requested):
good, but still one issue that I haven't mentioned yet, you should check the way your modal window loading, and how it should actually work

tiserett (approved):
great job 🚀

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {text !== '' && (
+            <>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
danylo-onishchuk, line 58: no pls

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,31 +43,40 @@ export const TodoModal: React.FC = () => {
               className="modal-card-title has-text-weight-medium"
               data-cy="modal-header"
             >
-              Todo #2
+              Todo #
+              {todoId}
             </div>
 
             {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
danylo-onishchuk, line 50: fix too

danylo-onishchuk (changes requested):

danylo-onishchuk (approved):

PR #1049, opened 2022-09-27
Solution implemented

OlehMy:
[DEMO LINK](https://OlehMy.github.io/react_dynamic-list-of-todos/)

I think failed tests because I changed values inside the select

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+  userId: 0,
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoader, setIsLoader] = useState(true);
+  const [selectedTodo, setSelectedTodo] = useState<Todo>(defaultTodo);
+  const [selectValue, setSelectValue] = useState('');
+  const [filterValue, setFilterValue] = useState('');
+
```
oksana-logos-frontend, line 26: filterValue should be 'all' by default

src/components/TodoList/TodoList.tsx
```diff
@@ -16,21 +28,54 @@ export const TodoList: React.FC = () => (
...
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
-      <tr data-cy="todo" className="has-background-info-light">
+      {todos.map(todo => (
+        <tr
```
oksana-logos-frontend, line 32: you can use here destruction

oksana-logos-frontend (approved):
Good job)

src/App.tsx
OlehMy: Do you mean selectValue?

oksana-logos-frontend: > Do you mean selectValue?

Yes, sorry....my mistake))))

OlehMy: Yes, you're right. But in my solution I use `""` to display all todos)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,30 @@
-export const TodoFilter = () => (
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
dellachaise, line 1: Try to add `aria-label` attribute to buttons all over the project to avoid using `eslint-disable`.

src/components/TodoList/TodoList.tsx
dellachaise: Agreed 👍 

Need to use destructuring.

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+  const hundleChange = (value: string) => {
+    setFilterValue(value);
+  };
+
+  const todosForRender
+    = todos
+      .filter(({ completed }) => (
+        selectValue ? String(completed) === selectValue : String(completed) !== selectValue
+      ))
```
dellachaise, line 54: Good solution 👍 
But better create a separate variable and use `switch case` to make code more readable.

dellachaise (changes requested):
Good job 👍 

Just add fixes according to the comments.

src/components/TodoFilter/TodoFilter.tsx
OlehMy: Great advice. Thanks))

src/App.tsx
OlehMy: no creativity whatsoever 😢

dellachaise: Creativety it is great 👍  

But your code still should be clear and easy to read for other programmers.
In the real project it is great when you use the ternary operator in some easy cases, but for some more complicated things should write easy to read code.

dellachaise (approved):
Well done 👍

PR #1053, opened 2022-09-27
Solution of react_dinamic-list-of-todos

OleksiiPustovit:
[DEMO LINK](https://oleksiipustovit.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,39 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todoId, setTodoId] = useState<number>(0);
+  const [loader, setLoader] = useState<boolean>(false);
+  const [filterByValue, setFilterByValue] = useState<string>('all');
+  const [filterByText, setFilterByText] = useState<string>('');
+
+  useEffect(() => {
+    getTodos()
+      .then(response => {
+        setTodos(response);
+        setLoader(true);
+      });
+  }, []);
+
+  const filteredTodos = todos.filter(todo => {
+    switch (filterByValue) {
+      case 'active':
+        return !todo.completed;
+
+      case 'completed':
+        return todo.completed;
+
+      default:
+        return todo;
+    }
+  }).filter(todo => {
+    return todo.title.toLowerCase().includes(filterByText.toLowerCase());
+  });
+
   return (
     <>
       <div className="section">
@@ -17,18 +48,36 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                todos={todos}
+                value={filterByValue}
+                text={filterByText}
+                setValue={setFilterByValue}
+                setText={setFilterByText}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              { !loader ? (
+                <Loader />
+              ) : (
+                <TodoList
+                  todos={filteredTodos}
+                  selectedTodoId={todoId}
+                  selectedTodo={(todo) => setTodoId(todo)}
+                />
+              )}
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {todoId !== 0 && (
+        <TodoModal
+          todoId={todoId}
+          todos={todos}
+          selectedTodo={(todo) => setTodoId(todo)}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,73 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { ChangeEvent } from 'react';
+import { Todo } from '../../types/Todo';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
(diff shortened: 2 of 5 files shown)


lizapolokhina (approved):
looks great:)

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todoId, setTodoId] = useState<number>(0);
+  const [loader, setLoader] = useState<boolean>(false);
```
dellachaise, line 16: Would be logically correct - if loader `true` - I see it, if `false` - no loader.
What do you think about it?

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todoId, setTodoId] = useState<number>(0);
```
dellachaise, line 15: No need to write primitive types, the same in line 18.
```suggestion
  const [todoId, setTodoId] = useState(0);
```

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+  const [loader, setLoader] = useState<boolean>(false);
+  const [filterByValue, setFilterByValue] = useState<string>('all');
```
dellachaise, line 17: Better create enum for filter type and also use it in `swith case`.

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+  const [filterByText, setFilterByText] = useState<string>('');
+
+  useEffect(() => {
+    getTodos()
+      .then(response => {
```
dellachaise, line 22: Let’s use `async await` all over the code. It gives you a simpler way to work with asynchronous promise-based code

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+      case 'active':
+        return !todo.completed;
+
+      case 'completed':
+        return todo.completed;
+
+      default:
+        return todo;
+    }
+  }).filter(todo => {
```
dellachaise, line 39: Now you twice filter todos. Can you filter just one time?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,73 @@
...
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={value}
+            onChange={handleChangeValue}
+          >
+            <option value="all">All</option>
```
dellachaise, line 36: You could import to this file the previously created enum and use it.

src/App.tsx
```diff
@@ -17,18 +48,36 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                todos={todos}
```
dellachaise, line 52: For what purpose did you need to pass `todos` as props in `TodoFilter`? When did you use it there?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,93 @@
...
-          </button>
-        </td>
-      </tr>
+      <tbody>
+        {todos.map(({ id, title, completed }) => {
+          const handleSelect = () => {
+            return selectedTodoId === id
+              ? selectedTodo(0)
+              : selectedTodo(id);
+          };
```
dellachaise, line 37: move this code in this file after line 15.

dellachaise (changes requested):
Great job 👍 

Just add improvements according to the comments.

The completed tasks should be green and the 
Screenshot 2022-10-01 at 00 16 51
uncompleted - red.

src/App.tsx
OleksiiPustovit: I completely agree))

OleksiiPustovit:
Thanks for the feedback, made changes according to the comments))

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,72 @@
...
+  setValue,
+  text,
+  setText,
+}) => {
+  const handleChangeValue = (event: ChangeEvent<HTMLSelectElement>) => {
+    setValue(event.currentTarget.value as FilterType);
+  };
+
+  const handleChangeText = (event:
+  ChangeEvent<HTMLInputElement>) => {
```
AlinaMatuschak, line 22:
```suggestion
  const handleChangeText = (
    event:  ChangeEvent<HTMLInputElement>,
  ) => {
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,72 @@
...
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+        <span
+          className="icon is-right"
+          style={{ pointerEvents: 'all' }}
+        >
+          {text !== '' && (
+            <>
```
AlinaMatuschak, line 58: do you need fragment here?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,95 @@
...
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map(({ id, title, completed }) => (
+          <tr
+            data-cy="todo"
+            key={id}
+            className=""
```
AlinaMatuschak, line 44:
```suggestion
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,40 @@
...
+  selectedTodo,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [todoOpen, setTodoOpen] = useState(true);
+  const currentTodo = todos.find(todo => todo.id === todoId);
+
+  useEffect(() => {
+    if (currentTodo) {
+      getUser(currentTodo.userId)
+        .then(response => setUser(response));
```
AlinaMatuschak, line 25: Let’s use async await. It gives you a simpler way to work with asynchronous promise-based code

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,40 @@
...
+import React, { useState, useEffect } from 'react';
 import { Loader } from '../Loader';
+import { getUser } from '../../api';
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+
+type Props = {
+  todoId: number;
+  todos: Todo[];
+  selectedTodo: (argument: number) => number | void;
```
AlinaMatuschak, line 10:
```suggestion
  selectTodo: (argument: number) => number | void;
```

AlinaMatuschak (approved):
Good work. But make sure that you fix all comments )

PR #1055, opened 2022-09-27
solution

Kulchevych:
[DEMO LINK](https://Kulchevych.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,52 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+
+enum GroupBy {
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(todos[0]);
+
+  useEffect(() => {
+    getTodos().then(response => {
+      setTodos(response);
+      setVisibleTodos(response);
+      setIsLoaded(true);
+    });
+  }, []);
+
+  const filterTodos = (groupBy: string, query: string) => {
+    setVisibleTodos(
+      todos.filter(todo => {
+        switch (groupBy) {
+          case GroupBy.All:
+            return true;
+          case GroupBy.Active:
+            return todo.completed === false;
+          case GroupBy.Completed:
+            return todo.completed === true;
+          default:
+            return true;
+        }
+      }).filter(todo => (
+        todo.title.toLowerCase().includes(query.toLocaleLowerCase())
+      )),
+    );
+  };
+
+  const setTodo = (todo: Todo | null) => {
+    setSelectedTodo(todo);
+  };
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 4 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,77 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={handleChange}
+            value={groupBy}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
max-kravchenko, line 45: if you set here enum values as values, I guess all your tests will pass

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,77 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query.trim().length > 0 ? (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
max-kravchenko, line 65: to get rid of this comment you could use aria-label with some descriptive text for the button

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,75 @@
...
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
-
-              {' by '}
-
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
-              </a>
-            </p>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
max-kravchenko, line 42: same comment as above about aria-label

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,75 @@
...
+              <button
+                type="button"
+                className="delete"
+                data-cy="modal-close"
+                onClick={() => {
+                  setTodo(null);
+                }}
```
max-kravchenko, line 49: I guess you would need to create a separate method here

max-kravchenko (approved):
Great job!

ps. Some ideas below as what to change to pass all tests

src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { GroupBy } from './types/GroupBy';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(todos[0]);
```
AlinaMatuschak, line 18: by default shouldn’t be selected any todo. And here will be undefined (todos is empty on the first render)
```suggestion
  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
```

src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos().then(response => {
+      setTodos(response);
+      setVisibleTodos(response);
+    }).finally(() => setIsLoading(false));
```
AlinaMatuschak, line 26: Let’s use async await. It gives you a simpler way to work with asynchronous promise-based code

src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
+            return !todo.completed;
+          case GroupBy.Completed:
+            return todo.completed;
+
+          default:
+            return true;
+        }
+      }).filter(todo => (
+        todo.title.toLowerCase().includes(query.toLocaleLowerCase())
+      )),
```
AlinaMatuschak, line 43: now you twice filter todos. Can you filter just one time?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,78 @@
...
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            <button
+              aria-label="query-delete"
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={resetQuery}
+            />
+          </span>
+        ) : ''}
```
AlinaMatuschak, line 74:
```suggestion
        {!!query.trim() && (
          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
            <button
              aria-label="query-delete"
              data-cy="clearSearchButton"
              type="button"
              className="delete"
              onClick={resetQuery}
            />
          </span>
        )}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,77 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { getUser } from '../../api';
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
 import { Loader } from '../Loader';
 
-export const TodoModal: React.FC = () => {
+type Props = {
+  todo: Todo | null;
```
AlinaMatuschak, line 8: Don’t need to render modal if you don't have selected todo
```suggestion
  todo: Todo;
```

src/types/GroupBy.ts
```diff
@@ -0,0 +1,5 @@
+export enum GroupBy {
```
AlinaMatuschak, line 1: 🚀

AlinaMatuschak (approved):

PR #1079, opened 2022-09-29
solution

lizapolokhina:
[DEMO LINK](https://lizapolokhina.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,49 @@
 /* eslint-disable max-len */
-import React from 'react';
+import { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todo, setTodo] = useState<Todo | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState('all');
+  const [query, setQuery] = useState('');
+  const [load, setLoad] = useState<Todo[]>(todos);
+
+  const lowerCompare = (str: string) => (
+    str.toLowerCase().includes(query.toLowerCase())
+  );
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos);
+  }, []);
+
+  useEffect((() => {
+    switch (filterBy) {
+      case 'all':
+        setLoad(todos.filter(({ title }) => lowerCompare(title)));
+        break;
+      case 'active':
+        setLoad(todos
+          .filter(({ completed, title }) => !completed && lowerCompare(title)));
+        break;
+      case 'completed':
+        setLoad(todos
+          .filter(({ completed }) => completed)
+          .filter(({ title }) => lowerCompare(title)));
+        break;
+      default:
+        throw new Error('Warning!');
+    }
+  }), [todos, filterBy, query]);
+
   return (
     <>
       <div className="section">
@@ -17,18 +52,30 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                filterBy={filterBy}
+                setFilterBy={setFilterBy}
+                query={query}
+                setQuery={setQuery}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!todos.length ? (
+                <Loader />
+              ) : (
+                <TodoList
+                  todos={load}
+                  setSelectedTodo={setTodo}
+                  selectedTodo={todo}
+                />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      <TodoModal todo={todo} setTodo={setTodo} />
     </>
   );
 };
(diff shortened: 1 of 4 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="clearSearchButton"
+            type="button"
+            className="delete"
+            onClick={() => setQuery('')}
+          />
```
OleksiiPustovit, line 60:
```suggestion
        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
          <button
            data-cy="clearSearchButton"
            type="button"
            className="delete"
            onClick={() => setQuery('')}
            aria-label="Some text"
          />
```

OleksiiPustovit: You can not use disabling the linter, but add an attribute to the button))

OleksiiPustovit (approved):
Good job 👍
But I have one suggestion, you can read about it by following this link.

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/control-has-associated-label.md

src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
+    str.toLowerCase().includes(query.toLowerCase())
+  );
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos);
+  }, []);
+
+  useEffect((() => {
+    switch (filterBy) {
```
dellachaise, line 29: To make the code more understandable, better create a separate variable  for filtered todos
for example:
```suggestion
   const filteredTodos = todos
    .filter(({ destructure }) => {
      switch (filterBy) {
        //code
      }
    });
```
and then `setLoad(filteredTodos)`.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filterBy}
+            onChange={handleStatus}
+          >
+            <option value="all">All</option>
```
dellachaise, line 33: You could import here the previously created enum and use it.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          <button
+            data-cy="clearSearchButton"
+            type="button"
+            className="delete"
+            onClick={() => setQuery('')}
```
dellachaise, line 58: Better create a separate method instead of creating an inline event handler(all over the code).

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,50 @@ export const TodoList: React.FC = () => (
...
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map((todo) => (
+        <tr
+          key={todo.id}
```
dellachaise, line 33: use destructure.

dellachaise (changes requested):
Great job 👍 

Just add changes according to comments 👇

src/components/TodoList/TodoList.tsx
lizapolokhina: I tried to use the destructure here, but it didn't quite work, because we need to refer to the todo object in several places

src/App.tsx
```diff
@@ -1,14 +1,55 @@
 /* eslint-disable max-len */
```
AlinaMatuschak, line 1: don't disable eslint

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { FilterTypes } from './types/FilterTypes';
 
 export const App: React.FC = () => {
+  const [todo, setTodo] = useState<Todo | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState('all');
+  const [query, setQuery] = useState('');
+  const [load, setLoad] = useState<Todo[]>(todos);
```
AlinaMatuschak, line 18:
```suggestion
  const [loadedTodos, setLoadedTodos] = useState<Todo[]>(todos);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,68 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filterBy}
+            onChange={handleStatus}
+          >
+            <option value={FilterTypes.all}>All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
AlinaMatuschak, line 38: use FilterTypes enum

AlinaMatuschak (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState(FilterTypes.all);
+  const [query, setQuery] = useState('');
+  const [loadedTodos, setLoadedTodos] = useState<Todo[]>(todos);
+
+  const compareLower = (title: string) => (
+    title.toLowerCase().includes(query.toLowerCase())
+  );
+
+  const isLoading = !todos.length;
```
AlinaMatuschak, line 23: What if the server will return an empty array?

AlinaMatuschak: need create state isLoading with default value true and set it to false when we get a response from the server

AlinaMatuschak (approved):

PR #1086, opened 2022-09-29
Version 0.1

David-Shahmatov:
[DEMO LINK](https://David-Shahmatov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,8 +7,55 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { useState } from 'react';
+import { useEffect } from 'react';
+
+export function getFilteredTodo(
+  todos: Todo[],
+  filterType: string,
+  query: string,
+) {
+  const filterBy = todos.filter((todo) => {
+    switch (filterType) {
+      case 'completed':
+        return todo.completed;
+
+      case 'active':
+        return !todo.completed;
+
+      default:
+        return todo;
+    }
+  });
+
+  return filterBy.filter(({ title }) => (
+    title.toLocaleLowerCase().includes(query.toLocaleLowerCase())
+  ));
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedUserId, setSelectedUserId] = useState(0);
+  const [selectedTodo, setSelectedTodo] = useState(0);
+  const [isLoading, setIsLoading] = useState(true);
+  const [filteredBy, setFilteredBy] = useState('all');
+  const [query, setQuery] = useState('');
+
+  useEffect(() => {
+    const getTodosFromServer = async () => {
+      const receivedTodos = await getTodos();
+
+      setTodos(receivedTodos);
+      setIsLoading(false);
+    };
+
+    getTodosFromServer();
+  }, []);
+
+  const filteredTodos = getFilteredTodo(todos, filteredBy, query);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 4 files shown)


artem-kodesnikov (approved):
Good!)

src/App.tsx
```diff
@@ -7,28 +7,96 @@ import { TodoList } from './components/TodoList';
...
+        return !todo.completed;
+
+      default:
+        return todo;
+    }
+  });
+
+  return filterBy.filter(({ title }) => (
+    title.toLocaleLowerCase().includes(query.toLocaleLowerCase())
+  ));
```
dellachaise, line 35: Now you twice filter todos. Can you filter just one time?

src/App.tsx
```diff
@@ -7,28 +7,96 @@ import { TodoList } from './components/TodoList';
...
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedUserId, setSelectedUserId] = useState(0);
+  const [selectedTodo, setSelectedTodo] = useState(0);
+  const [isLoading, setIsLoading] = useState(true);
+  const [filteredBy, setFilteredBy] = useState('all');
```
dellachaise, line 43: Create `enam` for filter options and use it here and in filter too.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,53 @@
...
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filteredBy}
+            onChange={(event) => setFilteredBy(event.target.value)}
+          >
+            <option value="all">All</option>
```
dellachaise, line 19: use here values from previously created `enum`.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,53 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {query && (
+            // eslint-disable-next-line jsx-a11y/control-has-associated-label
```
dellachaise, line 41: Using `eslint-disable` is a bag practice. Try to avoid it here and in test tasks. Find the way to solve it without `eslint-disable`.

Do the same all over the code.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,97 @@
...
+              {completed && (
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              )}
+            </td>
+            <td className={classNames('is-vcentered',
+              {
+                'is-expanded': !completed,
+              })}
```
dellachaise, line 48:
```suggestion
            <td className={classNames(
              'is-vcentered',
              { 'is-expanded': !completed }
            )}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,97 @@
...
+            >
+              <p className={
+                !completed
+                  ? 'has-text-danger'
+                  : 'has-text-success'
+              }
```
dellachaise, line 54: Use `classNames` library.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,97 @@
...
+              {selectedTodoId === id
+                ? (
+                  <button
+                    data-cy="selectButton"
+                    className="button"
+                    type="button"
+                    onClick={() => {
+                      selectedTodo(0);
+                      selectedUserId(0);
+                    }}
```
dellachaise, line 69: Create separate methods instead of creating inline event handlers **all over the code**.

dellachaise (changes requested):
Good start 👍 

Just need to add improvements according to the comments.

PR #1091, opened 2022-09-29
React dynamic list of TODOs

rboiar:
- [DEMO LINK](https://rboiar.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,43 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [queryFilter, setQueryFilter] = useState('');
+  const [selectStatus, setSelectStatus] = useState('all');
+  const [todoId, setTodoId] = useState(0);
+
+  useEffect(() => {
+    getTodos()
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+        setIsLoading(false);
+      });
+  }, []);
+
+  const filterBy = todos.filter((todo) => {
+    switch (selectStatus) {
+      case 'active':
+        return !todo.completed;
+
+      case 'completed':
+        return todo.completed;
+
+      default:
+        return true;
+    }
+  });
+
+  const visibleTodos = filterBy.filter(todoItem => {
+    return todoItem.title
+      .toLocaleLowerCase()
+      .includes(queryFilter.toLocaleLowerCase());
+  });
+
   return (
     <>
       <div className="section">
@@ -17,18 +52,37 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                todos={todos}
+                selectStatus={selectStatus}
+                setSelectStatus={setSelectStatus}
+                query={queryFilter}
+                setQuery={setQueryFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading
+                ? (
+                  <Loader />
+                ) : (
+                  <TodoList
+                    todos={visibleTodos}
+                    selectedTodoId={todoId}
+                    selectedTodo={todo => setTodoId(todo)}
+                  />
+                )}
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {todoId && (
+        <TodoModal
+          todos={visibleTodos}
+          todoId={todoId}
+          selectedTodo={todo => setTodoId(todo)}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,69 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { ChangeEvent } from 'react';
+import { Todo } from '../../types/Todo';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  todos: Todo[],
+  selectStatus: string,
+  query: string,
+  setSelectStatus: (arg: string) => (string) | void;
+  setQuery: (value: string) => (string) | void;
+};
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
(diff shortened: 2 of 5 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,90 @@
...
-              <strong className="has-text-danger">Planned</strong>
-
-              {' by '}
-
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
-              </a>
-            </p>
+              <p className="block" data-cy="modal-user">
+                {currentTodo?.completed ? (
```
Ohiiko, line 69: ? (   should be moved to a new line

src/App.tsx
```diff
@@ -17,18 +52,37 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                todos={todos}
```
Ohiiko, line 56: you don't use it in the filter

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,69 @@
...
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  todos: Todo[],
```
Ohiiko, line 6: you don't use it in the filter

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,83 @@
...
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              )}
+            </td>
+            <td className="is-vcentered is-expanded">
+              <p className={completed
+                ? 'has-text-success'
+                : 'has-text-danger'}
+              >
```
Ohiiko, line 53: better use classNames

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,83 @@
...
+                    ? selectedTodo(0)
+                    : selectedTodo(id);
+                }}
+              >
+                <span className="icon">
+                  {selectedTodoId === id
+                    ? (
+                      <i className="far fa-eye-slash" />
+                    ) : (
+                      <i className="far fa-eye" />
```
Ohiiko, line 73: better use classNames

Ohiiko (approved):
Good job. Fix small bugs.

src/components/TodoFilter/TodoFilter.tsx
AndriiRut: yeah! also we can't change props! So you cant pass an array to a child and mutate the array in it. all changes must be done in component where state/var/data was created!

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,69 @@
...
+            onChange={handleSelectStatus}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
+          </select>
+        </span>
+      </p>
+
+      <p className="control is-expanded has-icons-left has-icons-right">
```
AndriiRut, line 43: this part looks like an independent component

src/components/TodoList/TodoList.tsx
AndriiRut: ternary operator in className === reject

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,90 @@
...
-              className="modal-card-title has-text-weight-medium"
-              data-cy="modal-header"
-            >
-              Todo #2
-            </div>
+      {!user
+        ? (
+          <Loader />
+        ) : (
+          <div className="modal-card">
```
AndriiRut, line 43: if this is modal, create component with it

AndriiRut (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,49 @@
 /* eslint-disable max-len */
```
dellachaise, line 1: Delete `eslint-disable`. Using `eslint-disable` is a bag practice. Try to avoid it here and in test tasks. Find a way to solve it without `eslint-disable`.

src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [queryFilter, setQueryFilter] = useState('');
+  const [selectStatus, setSelectStatus] = useState('all');
```
dellachaise, line 17: Better create enum for filter type and also use it in `swith case`.

src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
+  const [todoId, setTodoId] = useState(0);
```
dellachaise, line 18:
```suggestion
  const [todoId, setTodoId] = useState(null);
```

src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then(todosFromServer => {
```
dellachaise, line 22: Let’s use `async await` all over the code. It gives you a simpler way to work with asynchronous promise-based code

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,67 @@
...
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
dellachaise, line 1: Using `eslint-disable` is a bag practice. Try to avoid it here and in test tasks. Find the way to solve it without `eslint-disable`.

Try to add `aria-label` attribute to buttons all over the project to avoid using `eslint-disable`.
More information  [here]( https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/control-has-associated-label.md)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,67 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={selectStatus}
+            onChange={handleSelectStatus}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
dellachaise, line 36: Do not rely on the unknown string, make constants for this. Better crate enum for filter options. 
When you would write test tasks, using enum in similar case would be a ➕ and could help you to get a job offer 😉

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,87 @@
...
+            <td className="has-text-right is-vcentered">
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => {
+                  return selectedTodoId === id
+                    ? selectedTodo(0)
+                    : selectedTodo(id);
+                }}
```
dellachaise, line 70: Create separate methods instead of creating inline event handlers, do always the same in all your projects and test tasks.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,91 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
dellachaise, line 1: Delete `eslint-disable`.

dellachaise (approved):
Good job 👍 

But need some improvements. Pay attention to the comments and follow the same rules when you would write test tasks, it is really important.

src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
+      default:
+        return true;
+    }
+  });
+
+  const visibleTodos = filterBy.filter(todoItem => {
+    return todoItem.title
+      .toLocaleLowerCase()
+      .includes(queryFilter.toLocaleLowerCase());
+  });
```
dellachaise, line 45: now you twice filter todos. Can you filter just one time?

PR #1092, opened 2022-09-30
solution

ShmelyovOleg:
[DEMO LINK](https://ShmelyovOleg.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +6,41 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loader, setLoader] = useState(false);
+  const [todoId, setTodoId] = useState(0);
+  const [statusFilter, setStatusFilter] = useState('all');
+  const [textFilter, setTextFilter] = useState('');
+
+  useEffect(() => {
+    getTodos()
+      .then(response => {
+        setTodos(response);
+        setLoader(true);
+      });
+  }, []);
+
+  const filtered = todos.filter(todoo => {
+    switch (statusFilter) {
+      case 'active':
+        return !todoo.completed;
+
+      case 'completed':
+        return todoo.completed;
+
+      default:
+        return true;
+    }
+  });
+
+  const visibleTodos = filtered.filter(todoo => {
+    return todoo.title.toLowerCase().includes(textFilter.toLowerCase());
+  });
+
   return (
     <>
       <div className="section">
@@ -17,18 +49,37 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                todos={todos}
+                value={statusFilter}
+                setValue={setStatusFilter}
+                text={textFilter}
+                setText={setTextFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!loader
+                ? (<Loader />)
+                : (
+                  <TodoList
+                    todos={visibleTodos}
+                    selectedTodoId={todoId}
+                    selectTodo={(todo) => setTodoId(todo)}
+                  />
+                )}
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {todoId !== 0
+        && (
+          <TodoModal
+            todoId={todoId}
+            todos={visibleTodos}
+            selectTodo={(todo) => setTodoId(todo)}
+          />
+        )}
     </>
   );
 };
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loader, setLoader] = useState(false);
+  const [todoId, setTodoId] = useState(0);
+  const [statusFilter, setStatusFilter] = useState('all');
```
dellachaise, line 16: Better create enum for filter type and also use it in `swith case`.

src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
+        return todoo.completed;
+
+      default:
+        return true;
+    }
+  });
+
+  const visibleTodos = filtered.filter(todoo => {
+    return todoo.title.toLowerCase().includes(textFilter.toLowerCase());
+  });
```
dellachaise, line 42: Now you twice filter todos. Can you filter just one time?

src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loader, setLoader] = useState(false);
+  const [todoId, setTodoId] = useState(0);
+  const [statusFilter, setStatusFilter] = useState('all');
+  const [textFilter, setTextFilter] = useState('');
+
+  useEffect(() => {
+    getTodos()
+      .then(response => {
```
dellachaise, line 21: Let’s use `async await` all over the code and in future test tasks too. It gives you a simpler way to work with asynchronous promise-based code.

src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loader, setLoader] = useState(false);
```
dellachaise, line 14: I think initially better set the loader to `true` because firstly we have no todos and need to show the loader(it would be kind of logical). What do you think?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={value}
+            onChange={handleChangeValue}
+          >
+            <option value="all"> All </option>
+            <option value="active"> Active </option>
+            <option value="completed"> Completed </option>
```
dellachaise, line 37: You could import to this file the previously created enum and use it here.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
         />
-      </span>
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
dellachaise, line 55: Better create a class instead of using `style`.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
+          {text !== '' && (
+            <>
+              <button
+                data-cy="clearSearchButton"
+                type="button"
+                className="delete"
+                aria-label="Clear search query"
```
dellachaise, line 62: 👍

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,86 @@
...
+            >
+              <td className="is-vcentered">{id}</td>
+              <td className="is-vcentered">
+                {todo.completed
+                  && (
+                    <span className="icon" data-cy="iconCompleted">
+                      <i className="fas fa-check" />
+                    </span>
+                  )}
+              </td>
```
dellachaise, line 53: Better create a separate component for the todo item.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,40 @@
...
+  const [cardOpen, setCardClose] = useState(true);
+  const [user, setUser] = useState<User | null>(null);
+  const currentTodo = todos.find(
+    (todo: Todo) => todo.id === todoId,
+  );
+
+  useEffect(() => {
+    if (currentTodo) {
+      getUser(currentTodo.userId)
+        .then(response => {
```
dellachaise, line 23: Here the same.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,31 +43,37 @@ export const TodoModal: React.FC = () => {
...
-            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
             <button
               type="button"
               className="delete"
               data-cy="modal-close"
+              aria-label="Select todo"
+              onClick={() => {
+                setCardClose(false);
+                selectTodo(0);
+              }}
```
dellachaise, line 57: Create separate methods instead of creating inline event handlers.

dellachaise (changes requested):
Good job 👍 

Just need to add some improvements according to the comments.

And the same way you should do when you will write your future test tasks(must be taken into account comments from this PR and others). It would make your code clear and will help to write better test task.

dellachaise (changes requested):
See the comments ☝️

PR #1104, opened 2022-10-05
Solution

vyurkov:
[DEMO LINK](https://vyurkov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -8,7 +8,55 @@ import { TodoFilter } from './components/TodoFilter';
...
+import { getTodos } from './api';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState([] as Todo[]);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [filter, setFilter] = useState({ status: '', query: '' });
+  const [modalTodo, setModalTodo] = useState({} as Todo);
+
+  // eslint-disable-next-line no-console
+  console.log(modalTodo);
```
TarasHoliuk, line 21: delete it

src/App.tsx
```diff
@@ -17,18 +65,22 @@ export const App: React.FC = () => {
...
+              {!isLoaded
+                ? <Loader />
+                : <TodoList todos={visibleTodos} onTodoSelect={setModalTodo} selectedTodo={modalTodo} />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {Object.keys(modalTodo).length > 0 && <TodoModal todo={modalTodo} onModalChange={setModalTodo} />}
```
TarasHoliuk, line 83: better set `null` by default instead of an empty object and then check here will look like `{modalTodo && <TodoModal propExample={yourProps} />}`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,60 @@
...
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+type Props = {
+  filter: { status: string, query: string },
+  onFilterChange: (filter: { status: string, query: string }) => void,
```
TarasHoliuk, line 3:
```suggestion
  onFilterChange(filter: { status: string, query: string }): void,
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,38 @@
...
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo,
+  onModalChange: (todo: Todo) => void,
+};
+
+export const TodoModal: React.FC<Props> = ({ todo, onModalChange }) => {
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [user, setUser] = useState({} as User);
```
TarasHoliuk, line 14:
```suggestion
  const [user, setUser] = useState<User | null>(null);
```

TarasHoliuk (changes requested):
Good job! I've left a few suggestions - check it, please

nazarmatsevych (approved):

PR #1120, opened 2022-10-28
Develop

Maksym845:
- [DEMO LINK](https://Maksym845.github.io/react_dynamic-list-of-todos/)

I have no idea why tests are not passing(

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,7 +8,7 @@ import { getTodos } from './api';
 import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
-import { Values } from './types/Enum';
+import { SelectTypes } from './types/Enum';
 import { TodoFilter } from './components/TodoFilter';
 
 export const App: React.FC = () => {
@@ -34,19 +34,21 @@ export const App: React.FC = () => {
 
   const filtered = todosList;
 
-  const search = (query: string, select: Values) => {
-    if (select === Values.COMPLETED && todosList) {
-      return [...todosList].filter(todo => {
-        return todo.title.toLowerCase().includes(query.toLowerCase()) && todo.completed;
+  const search = (query: string, select: SelectTypes) => {
+    const lowercasedQuery = query.toLowerCase();
+
+    if (select === SelectTypes.COMPLETED && todosList) {
+      return todosList.filter(todo => {
+        return todo.title.toLowerCase().includes(lowercasedQuery) && todo.completed;
       });
     }
 
-    if (select === Values.ACTIVE && todosList) {
-      return [...todosList].filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()) && !todo.completed);
+    if (select === SelectTypes.ACTIVE && todosList) {
+      return todosList.filter(todo => todo.title.toLowerCase().includes(lowercasedQuery) && !todo.completed);
     }
 
-    if (select === Values.ALL && todosList) {
-      return [...todosList].filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()));
+    if (select === SelectTypes.ALL && todosList) {
+      return todosList.filter(todo => todo.title.toLowerCase().includes(lowercasedQuery));
     }
 
     return null;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,13 +12,9 @@ import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
   const [todosList, setTodosList] = useState<Todo[] | null>(null);
-  const [visibleTodos, setVisibleTodos] = useState<Todo[] | null>(null);
   const [userId, setUserId] = useState(0);
-  const [selectedTodo, setSelectedTodo] = useState<Todo>();
-
-  const selectTodo = (todo: Todo) => {
-    setSelectedTodo(todo);
-  };
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[] | null>(null);
 
   const getTodosFromApi = async () => {
     const todos = await getTodos().then(todosApi => todosApi);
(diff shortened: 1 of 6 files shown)


Oleksandra-Shevchenko (approved):
GJ 👍

src/App.tsx
```diff
@@ -1,34 +1,105 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[] | null>(null);
+  const [userId, setUserId] = useState(0);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[] | null>(null);
+
+  const getTodosFromApi = async () => {
```
Andr1yK, line 19: Do u need to create this fn on each re-render?

src/App.tsx
```diff
@@ -1,34 +1,105 @@
...
+    const todos = await getTodos().then(todosApi => todosApi);
+
+    if (todos !== undefined) {
+      setTodosList(() => {
+        return todos;
+      });
```
Andr1yK, line 25:
```suggestion
      setTodosList(todos);
```

src/App.tsx
```diff
@@ -1,34 +1,105 @@
...
+    return null;
+  };
+
+  const filtered = todosList;
+
+  const search = (query: string, select: string) => {
+    const filter = filterQeury(query);
+
+    if (select === 'completed' && filter) {
+      return filter.filter(todo => todo.completed);
```
Andr1yK, line 50: let's use one filter

src/App.tsx
```diff
@@ -1,34 +1,105 @@
...
+                search={search}
+                setVisibleTodos={setVisibleTodos}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {(filtered
+                && (
```
Andr1yK, line 79: OMG, it's strange, let's use ternary :)

src/App.tsx
```diff
@@ -1,34 +1,105 @@
...
+                ))
+                || <Loader />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {userId !== 0
+        && selectedTodo
```
Andr1yK, line 94:
```suggestion
      {userId && selectedTodo
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
+            onChange={(event) => {
+              setSelected(event.target.value);
+              const todos = search(text, event.target.value);
+
+              setVisibleTodos(todos);
+            }}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
Andr1yK, line 30: Create enum with those values

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={selected}
+            onChange={(event) => {
+              setSelected(event.target.value);
+              const todos = search(text, event.target.value);
+
+              setVisibleTodos(todos);
```
Andr1yK, line 25: do u need this state, as for me it only creates more rerenders😅

src/components/TodoList/TodoList.tsx
```diff
@@ -14,87 +28,61 @@ export const TodoList: React.FC = () => (
...
+              ? (
+                <td className="is-vcentered">
+                  <span className="icon" data-cy="iconCompleted">
+                    <i className="fas fa-check" />
+                  </span>
+                </td>
+              )
+              : (
+                <td className="is-vcentered" />
+              )}
```
Andr1yK, line 54:
```suggestion
              : <td className="is-vcentered" />
            }
```

Andr1yK (changes requested):

src/components/TodoFilter/TodoFilter.tsx
Maksym845: Sorry, I didn`t understand your tip, what exactly state is unneeded?

Andr1yK: You can use useMemo instead of using setState to remove a couple of redundant renderers, but it's optional

src/App.tsx
```diff
@@ -1,61 +1,52 @@
...
-  };
-
   const filtered = todosList;
 
-  const search = (query: string, select: string) => {
-    const filter = filterQeury(query);
-
-    if (select === 'completed' && filter) {
-      return filter.filter(todo => todo.completed);
+  const search = (query: string, select: Values) => {
```
Andr1yK, line 37: Why exactly the `value`? What is this meaning? Let's rename it

src/App.tsx
```diff
@@ -1,61 +1,52 @@
...
+    if (select === Values.COMPLETED && todosList) {
+      return [...todosList].filter(todo => {
+        return todo.title.toLowerCase().includes(query.toLowerCase()) && todo.completed;
```
Andr1yK, line 40: Do u need to do lowercasing on each iteration?

src/App.tsx
```diff
@@ -1,61 +1,52 @@
...
+      });
     }
 
-    if (select === 'active' && filter) {
-      return filter.filter(todo => !todo.completed);
+    if (select === Values.ACTIVE && todosList) {
+      return [...todosList].filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()) && !todo.completed);
```
Andr1yK, line 45: image

Andr1yK (changes requested):

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
-import { TodoFilter } from './components/TodoFilter';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { SelectTypes } from './types/Enum';
+import { TodoFilter } from './components/TodoFilter';
 
 export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[] | null>(null);
```
Anna-Tkachenko, line 15: let's store here an empty array instead of null if there is no todos

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
+  const [userId, setUserId] = useState(0);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [visibleTodos, setVisibleTodos] = useState<Todo[] | null>(null);
```
Anna-Tkachenko, line 18: same here, will be better to store an empty array instead of null

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
+
+  const getTodosFromApi = useCallback(async () => {
+    const todos = await getTodos().then(todosApi => todosApi);
```
Anna-Tkachenko, line 21: then after await function is redundant

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
+
+    if (todos !== undefined) {
+      setTodosList(todos);
+      setVisibleTodos(() => {
+        return todos;
+      });
```
Anna-Tkachenko, line 27: why did you use functional set state here?

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
+    }
+  }, []);
+
+  useEffect(() => {
+    getTodosFromApi();
+  }, []);
+
+  const filtered = todosList;
```
Anna-Tkachenko, line 35: what's the reason of creating this variable?

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
+      });
+    }
+
+    if (select === SelectTypes.ACTIVE && todosList) {
+      return todosList.filter(todo => todo.title.toLowerCase().includes(lowercasedQuery) && !todo.completed);
+    }
+
+    if (select === SelectTypes.ALL && todosList) {
+      return todosList.filter(todo => todo.title.toLowerCase().includes(lowercasedQuery));
+    }
```
Anna-Tkachenko, line 52: if you need to do some logic depending on the value of one variable, will be better to use switch / case instead of if / else

src/components/TodoList/TodoList.tsx
```diff
@@ -14,87 +28,59 @@ export const TodoList: React.FC = () => (
...
+                className="button"
+                type="button"
+                onClick={() => {
+                  setUserId(userId);
+                  selectTodo(todo);
+                }}
+              >
+                <span className="icon">
+                  <i className={cn('far', {
+                    'fa-eye': selected?.id !== id,
```
Anna-Tkachenko, line 74: you checked is current todo selected several times, let's create separate boolean variable inside the callback with the result of this check and will use it

Anna-Tkachenko (changes requested):

PR #1126, opened 2022-10-31
Implemented the dynamic list of todos

anastasiia-nahaiska:
[DEMO LINK](https://anastasiia-nahaiska.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,80 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-
-import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { TodoList } from './components/TodoList';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+export enum FilterBy {
+  ALL = 'all',
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
+
+const getFilteredTodos = (todos: Todo[], filerBy: FilterBy, query: string) => {
+  const filteredTodos = todos.filter(({ title }) => (
+    title.toLowerCase().includes(query.toLowerCase())
+  ));
+
+  if (filerBy !== FilterBy.ALL) {
+    switch (filerBy) {
+      case FilterBy.ACTIVE:
+        return filteredTodos.filter(({ completed }) => !completed);
+
+      case FilterBy.COMPLETED:
+        return filteredTodos.filter(({ completed }) => completed);
+
+      default:
+        break;
+    }
+  }
+
+  return filteredTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [filerBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+
+  useEffect(() => {
+    const loadTodos = async () => {
+      const todosFromApi = await getTodos();
+
+      setTodos(todosFromApi);
+      setIsLoaded(true);
+    };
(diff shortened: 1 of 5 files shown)


AntonVaida (approved):

src/App.tsx
```diff
@@ -1,14 +1,104 @@
...
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
+
+const getFilteredTodos = (todos: Todo[], filerBy: FilterBy, query: string) => {
+  const filteredTodos = todos.filter(({ title }) => (
+    title.toLowerCase().includes(query.toLowerCase())
+  ));
+
+  if (filerBy !== FilterBy.ALL) {
```
Anna-Tkachenko, line 28: you can add this case inside switch / case functionality instead of creating separate if statement

src/App.tsx
```diff
@@ -1,14 +1,104 @@
...
+  }, []);
+
+  const selectFilterType = useCallback((
+    event: React.ChangeEvent<HTMLSelectElement>,
+  ) => {
+    setFilterBy(event.target.value as FilterBy);
+  }, []);
+
+  const applyQuery = useCallback(
+    debounce(setAppliedQuery, 1000),
```
Anna-Tkachenko, line 83: 🔥

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,50 @@ export const TodoList: React.FC = () => (
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          data-cy="todo"
+          className={cn({
+            'has-background-info-light': selectedTodo?.id === todo.id,
```
Anna-Tkachenko, line 35: you check is current todo selected several times, create separate boolean variable for this value and use it where you need it

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,100 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+export enum FilterBy {
+  ALL = 'all',
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
```
Denchepornyuk, line 21: it shouldn't be in the App file

src/App.tsx
```diff
@@ -1,14 +1,100 @@
...
+
+const getFilteredTodos = (todos: Todo[], filerBy: FilterBy, query: string) => {
+  const filteredTodos = todos.filter(({ title }) => (
+    title.toLowerCase().includes(query.toLowerCase())
```
Denchepornyuk, line 25: move `query.toLowerCase()` from filter, don't do it on each filter iteration

src/App.tsx
```diff
@@ -1,14 +1,100 @@
...
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [appliedqQuery, setAppliedQuery] = useState('');
+  const [filerBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+  const [hasLoadingError, setHasLoadingError] = useState(false);
+
+  useEffect(() => {
+    const loadTodos = async () => {
+      try {
+        const todosFromApi = await getTodos();
```
Denchepornyuk, line 52: what if getTodos return an error? You should catch an error here

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,47 @@
...
+    userId,
+    completed,
+  } = todo;
+  const [user, setUser] = useState<User | null>(null);
+  const [hasLoadingError, setHasLoadingError] = useState(false);
+
+  useEffect(() => {
+    const loadUser = async () => {
+      try {
+        const userFromApi = await getUser(userId);
```
Denchepornyuk, line 29: what if getUser return an error? You should catch an error here

Denchepornyuk (changes requested):

Denchepornyuk (changes requested):
I don't see changes from previous review

anastasiia-nahaiska:
> I don't see changes from previous review

Check now, please
maybe something went wrong while deploy was doing(

Denchepornyuk (approved):

PR #1134, opened 2022-11-01
React_dynamic-list-of-todos

DenysKleimenov:
[DEMO LINK](https://DenysKleimenov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


vladimir-dublya (approved):

src/App.tsx
```diff
@@ -17,18 +52,34 @@ export const App: React.FC = () => {
...
+                query={query}
+                setQuery={setQuery}
+                setFilter={setFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0 && <Loader />}
```
Anna-Tkachenko, line 63: will be better to handle it with additional boolean variable, that will define if data was loaded
for now if you'll receive an empty array from the server, user will see infinite loader on the page

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,88 @@
...
-);
+      <tbody>
+        {todos.map(todo => {
+          const { id, title, completed } = todo;
+
+          return (
+            <tr
+              data-cy="todo"
+              className={cn(
+                { 'has-background-info-light': todo.id === selectedTodoId },
```
Anna-Tkachenko, line 44: you check if current todo is selected 2 times, let's create separate variable for this value before return statement in the map callback

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,37 @@
...
+    }
+  }, []);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!user ? (
```
Anna-Tkachenko, line 34: same here about user loading

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,37 @@
...
+};
+
+export const TodoModal: React.FC<Props> = ({
+  selectedTodoId,
+  setSelectedTodoId,
+  todos,
+}) => {
+  const [user, setUser] = useState<User>();
+
+  const selectedTodo = todos.find(todo => todo.id === selectedTodoId);
```
Anna-Tkachenko, line 21: will be better to store null here if you don't find todo add show corresponding message about it

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 🔥

PR #1135, opened 2022-11-01
Dynamic list of todos was implemented

Yevheniya-Deryka:
[DEMO LINK](https://Yevheniya-Deryka.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,12 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState, useCallback } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+
+import { getTodos } from './api';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,77 @@
...
+
+  const filterTodos = () => {
+    let filteredTodos;
+
+    switch (status) {
+      case Status.Active:
+        filteredTodos = todos.filter(todo => !todo.completed);
+        break;
+      case Status.Completed:
+        filteredTodos = todos.filter(todo => todo.completed);
```
nikitaovch99, line 58: You could just set in cases a callback by witch you will filter to not write filtering everytime but thats ok

nikitaovch99 (approved):
Left some comments. Also I can write a Word in search but if I Put spaces in the end and start it wont find it. But everything else is good.

Yevheniya-Deryka:
> Left some comments. But everything else is good.

yep! already fixed the point with spaces in query) thanks!

src/App.tsx
silchyn, line 58: `filteredTodos = todos.filter(todo =>` repeats

silchyn (changes requested):

silchyn (approved):
Cool!

PR #1137, opened 2022-11-01
Implement getting todos from server and rendering them

imaksweb:
- [DEMO LINK](https://imaksweb.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


vmamchur (approved):
Everything looks good to me.
GJ 👍🏻

src/App.tsx
```diff
@@ -17,18 +66,35 @@ export const App: React.FC = () => {
...
+                query={query}
+                setQuery={setQuery}
+              />
             </div>
 
-            <div className="block">
-              <Loader />
-              <TodoList />
-            </div>
+            {!todos.length
```
Anna-Tkachenko, line 77: will be better to handle loader with separate boolean variable
for now if you receive an empty array from server, user will see infinite loader

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,68 @@ export const TodoList: React.FC = () => (
...
+                  <button
+                    data-cy="selectButton"
+                    className="button"
+                    type="button"
+                    onClick={() => selectTodo(0)}
+                  >
+                    <span className="icon">
+                      <i className="far fa-eye-slash" />
+                    </span>
+                  </button>
```
Anna-Tkachenko, line 76: you can also create one button, pass isActive value to button handler and call needed setState according to those value

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,36 @@
...
+    }
+  }, []);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!todo || !user ? (
```
Anna-Tkachenko, line 33: same here, will be better to handle it with separate variable for loading

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 🔥

PR #1139, opened 2022-11-01
dynamic todo list using context

TheVovchik:
**[DEMO LINK](https://thevovchik.github.io/react_dynamic-list-of-todos/)**

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useContext, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,26 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { UserContext } from './components/UserContext';
 
 export const App: React.FC = () => {
+  const {
+    todos,
+    setTodos,
+    selectedTodoId,
+  } = useContext(UserContext);
+
+  const getTodosFromServer = async () => {
+    const newTodos = await getTodos();
+
+    setTodos(newTodos);
+  };
+
+  useEffect(() => {
+    getTodosFromServer();
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -21,14 +39,15 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0
+                ? <Loader />
+                : <TodoList />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodoId && <TodoModal />}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,80 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { FC, useContext, useCallback } from 'react';
+import { UserContext } from '../UserContext';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+function debounce(f:(finalQuery: string) => void, delay: number) {
+  let timerId: NodeJS.Timeout;
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+  return (arg: string) => {
+    clearTimeout(timerId);
+    timerId = setTimeout(f, delay, arg);
+  };
+}
+
+export const TodoFilter: FC = () => {
+  const {
+    setSortType,
+    query,
+    setQuery,
+    setAppliedQuery,
+  } = useContext(UserContext);
+  const changeSortType = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const { value } = event.target;
+
+    setSortType(value);
+  };
+
+  const applyQuery = useCallback(debounce(setAppliedQuery, 1000),
+    []);
+  const handleSearch = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    applyQuery(event.target.value);
+  };
+
+  const clearSearchBar = () => {
+    setQuery('');
+    setAppliedQuery('');
+  };
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={changeSortType}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
+          </select>
+        </span>
+      </p>
+
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          placeholder="Search..."
+          value={query}
+          onChange={handleSearch}
         />
-      </span>
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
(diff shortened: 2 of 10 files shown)


src/components/UserContext/UserContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+  setSelectedTodoId: () => {},
+  userInfo: null,
+  setUserInfo: () => {},
+  sortType: 'all',
+  setSortType: () => {},
+  query: '',
+  setQuery: () => {},
+  appliedQuery: '',
+  setAppliedQuery: () => {},
+});
```
grafit-off, line 37: Wow, it's really big one 😮

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,80 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
grafit-off, line 68: FYI, you can use aria-label attribute to avoid this one

src/components/UserContext/UserContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const UserProvider: React.FC = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+  const [userInfo, setUserInfo] = useState<User | null>(null);
+  const [sortType, setSortType] = useState('all');
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+
```
grafit-off, line 47: Looks convenient to have such data in one place

grafit-off (approved):
Good usage of context! 🔥

src/components/UserContext/UserContext.tsx
TheVovchik: probably we can split this context on  more smaller contexts, but don't know what is better

src/components/TodoFilter/TodoFilter.tsx
TheVovchik: I googled FYI, it's For Your Information?))) =)

grafit-off: @TheVovchik it's like for informational purposes only and doesn't require any action)

src/App.tsx
```diff
@@ -21,14 +39,15 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0
```
Anna-Tkachenko, line 42: will be awesome to handle loader with separate variable
for now if you receive an empty array from server, user will see infinite loader

Anna-Tkachenko: (not important, just suggestion)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,80 @@
...
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+import { FC, useContext, useCallback } from 'react';
+import { UserContext } from '../UserContext';
+
+function debounce(f:(finalQuery: string) => void, delay: number) {
+  let timerId: NodeJS.Timeout;
+
+  return (arg: string) => {
```
Anna-Tkachenko, line 7: will be great to handle all available arguments, that can be passed to the debounce function, for now you can't reuse it for the function with other argument

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,65 @@
...
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+  const getVisibleTodos = (currentTodos: Todo[]) => {
+    switch (sortType) {
+      case 'all':
```
Anna-Tkachenko, line 25: will be better to create enum for these values

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,65 @@
...
-);
+          </th>
+          <th>Title</th>
+          <th> </th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {visibleTodos.map(todo => (
+          <React.Fragment key={todo.id}>
```
Anna-Tkachenko, line 58: 🔥

src/components/UserContext/UserContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+  setSortType: (type: string) => void,
+  query: string,
+  setQuery: (newQuery: string) => void,
+  appliedQuery: string,
+  setAppliedQuery: (finalQuery: string) => void,
+};
+
+export const UserContext = React.createContext<Context>({
+  todos: [],
+  setTodos: () => {},
```
Anna-Tkachenko, line 24: it's not great to move todos and todos handler to User Context
will be better to create separate TodoContext for this purpose

Anna-Tkachenko (approved):

src/components/UserContext/UserContext.tsx
TheVovchik: will try to fix it on weekend

src/App.tsx
TheVovchik: done

src/components/TodoFilter/TodoFilter.tsx
TheVovchik: don't know why but only this option work

src/components/UserContext/UserContext.tsx
```diff
@@ -0,0 +1,137 @@
...
+  getTodosFromServer: () => void,
+  changeSortType: (event: React.ChangeEvent<HTMLSelectElement>) => void,
+  applyQuery: ((arg: string) => void),
+  handleSearch: (event: React.ChangeEvent<HTMLInputElement>) => void,
+  clearSearchBar: () => void,
+  selectCurrentTodo: (id: number, userId: number) => void
+};
+
+export const UserContext = React.createContext<Context>({
+  todos: [],
```
Anna-Tkachenko, line 30: userContext shouldn't know anything about todo info and function for managing todos
let's create 2 separate context: one for user data and functions for handling each changes, related to user
and second one for the same with todo data

in this way will be easier to add management function for separate models (for user and for todo), and it'll allow use todo or user provider separately of each other

src/components/utils/debounce.ts
```diff
@@ -0,0 +1,8 @@
+export function debounce(f:(finalQuery: string) => void, delay: number) {
+  let timerId: NodeJS.Timeout;
+
+  return (arg: string) => {
```
Anna-Tkachenko, line 4: will be awesome to accept any arguments here for making debounce the universal one
you can use arguments keyword for this purpose

src/types/SortTypes.ts
```diff
@@ -0,0 +1,5 @@
+export enum SortTypes {
```
Anna-Tkachenko, line 1: 🔥

src/components/UserContext/UserContext.tsx
```diff
@@ -0,0 +1,137 @@
...
+});
+
+export const UserProvider: React.FC = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedUser, setSelectedUser] = useState(0);
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+  const [userInfo, setUserInfo] = useState<User | null>(null);
+  const [sortType, setSortType] = useState<SortTypes>(SortTypes.ALL);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
```
Anna-Tkachenko, line 57: variables that can be changeable a lot of times will be better to move to the component, where you handle them
in context you should store not all variables for your app, just the main one, the main data of your app (like todo and user for this one)

other variables, related to the query and to the info, was data loaded or no, you can move to the state of the needed component

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,30 @@
...
+    getUserFromServer,
+    handleClose,
+  } = useContext(UserContext);
+
+  useEffect(() => {
+    getUserFromServer();
+  }, []);
+
+  const currentTodo: Todo | undefined = todos
+    .find(todo => todo.id === selectedTodoId);
```
Anna-Tkachenko, line 21: let's return null if we don't find todo

Anna-Tkachenko (changes requested):

src/components/utils/debounce.ts
TheVovchik: have some problems to write rigth types for debounce, hope I successed

src/components/UserContext/UserContext.tsx
TheVovchik: hope I understand :)

src/components/UserContext/UserContext.tsx
TheVovchik: In my case I throw query and applied query from TodoFilter to TodoList and they connect only through Context, will try to connect them in other way tomorrow, now just separate User and Todo in two Contexts

src/components/TodoModal/TodoModal.tsx
TheVovchik: changed 👍

Denchepornyuk (approved):
good job 👍

PR #1140, opened 2022-11-01
implemented react-dynamic-list-of-todos

nikitaovch99:
[DEMO LINK](https://nikitaovch99.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+
+type Props = () => Promise<void>;
```
Yevheniya-Deryka, line 13: I think name Props usually is for type that the component gets. So maybe will be better to chose some other name here

src/App.tsx
```diff
@@ -17,18 +71,32 @@ export const App: React.FC = () => {
...
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todosAreLoading
+              && <Loader />}
+              <TodoList
+                todos={visiableTodos}
+                onTodoSelected={onTodoSelected}
+                selectedTodo={selectedTodo}
+              />
```
Yevheniya-Deryka, line 89: I think will be better to render TodoList only if you already have false in todosAreLoading.
now you have an empty table while getting todos from server.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,43 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { Loader } from '../Loader';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  selectedTodo: Todo | null;
```
Yevheniya-Deryka, line 8: it seems to me you're rendering this component only if you have selectedTodo in App, so it should work fine without this null option in type

Yevheniya-Deryka (approved):
GJ! 
Left few comments

src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+  const handleSetStatus = (value: string) => {
+    setStatus(value);
+  };
+
+  const doesTitleMatch = (todo: Todo) => (
+    todo.title.toLowerCase().includes(query.toLowerCase().trim()));
+
+  const doesStatusMatch = (todo: Todo) => {
+    switch (status) {
+      case 'completed':
```
Anna-Tkachenko, line 53: will be better to create enum for these values

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,82 @@
...
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map((todo) => (
+          <tr
+            data-cy="todo"
+            className={classNames({
+              'has-background-info-light': selectedTodo?.id === todo.id,
```
Anna-Tkachenko, line 38: you check several times is current todo selected, create separate boolean variable for this value and use it where you need it

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+type GetTodos = () => Promise<void>;
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosAreLoading, setTodosAreLoading] = useState(true);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [status, setStatus] = useState('');
+  const getTodosFromServer: GetTodos = async () => {
+    const todosFromServer = await getTodos();
```
Denchepornyuk, line 23: what if getTodos return an error? You should catch an error here

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,43 @@
...
+  {
+    selectedTodo,
+    onResetSelectedTodo,
+  },
+) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [userIsLoading, setUserIsLoading] = useState(true);
+
+  const getUsersFromServer: GetUsers = async (userId) => {
+    const usersFromServer = await getUser(userId);
```
Denchepornyuk, line 23: what if getUser return an error? You should catch an error here

Denchepornyuk (changes requested):

PR #1142, opened 2022-11-02
dynamic list of todos implemented

artem-musii:
[DEMO LINK](https://artem-musii.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,7 +15,7 @@ export const App: React.FC = () => {
   const [loadedTodos, setLoadedTodos] = useState<Todo[]>([]);
   const [isShowClicked, setIsShowClicked] = useState(false);
   const [selectedUserId, setSelectedUserId] = useState(0);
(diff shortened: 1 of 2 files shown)


Vladosik1337 (approved):
Nice✨

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,56 @@ export const TodoList: React.FC = () => (
...
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          data-cy="todo"
+          key={todo.id}
+          className={classNames(
+            {
+              'has-background-info-light': todo.id === selectedTodo?.id,
```
Anna-Tkachenko, line 37: you check if current todo is selected several times, will be better to create separate boolean variable for this value and use it where you need it

Anna-Tkachenko (approved):
Good job!

PR #1146, opened 2022-11-02
Develop

artem-nasadchuk:
[DEMO LINK](https://artem-nasadchuk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,76 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { GroupBy } from '../../types/GroupBy';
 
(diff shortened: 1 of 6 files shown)


illia-sosnytskyi (approved):
nice work

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,76 @@
...
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span
+          className="icon is-right"
+          style={{ pointerEvents: 'all' }}
+        >
+          {query && (
+            <>
```
Anna-Tkachenko, line 62: react fragment is redundant here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,93 @@
...
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map(({ id, title, completed }) => (
+          <tr
+            data-cy="todo"
+            className={cn({
+              'has-background-info-light': selectedTodoId === id,
```
Anna-Tkachenko, line 35: you check is current todo selected several times, create separate boolean variable for this value and use it in places where you need it

Anna-Tkachenko (approved):
Good job!

PR #1148, opened 2022-11-02
dynamic list of todos using react was implemented

dsydorenko534:
[DEMO LINK](https://dsydorenko534.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+  id: 0,
+  title: 'no title',
+  completed: false,
+  userId: 0,
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoader, setIsLoader] = useState(true);
+  const [selectedTodo, setSelectedTodo] = useState<Todo>(defaultTodo);
```
iPanda177, line 22:
```suggestion
export const App: React.FC = () => {
  const [todos, setTodos] = useState<Todo[]>([]);
  const [isLoader, setIsLoader] = useState(true);
  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
```

iPanda177: do you really need to create default todo? or you can just put null as a default value?)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,33 +1,55 @@
...
+  hundleChange,
+}) => (
   <form className="field has-addons">
     <p className="control">
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
+          data-cy="statusSelect"
+          value={selectValue}
+          onChange={(event) => hundleSelect(event.target.value)}
```
iPanda177, line 22: better to move it to separate method

iPanda177 (changes requested):
Cool!
Just a few things to think about)

src/App.tsx
dsydorenko534: I can't do that because my selectedTodo and todo might return null and It'll be a mistake. However as far as I'm concerned I consider this to be a better practice than using null

Andr1yK: > I'm concerned I consider this to be a better practice than using null

I do not agree with this, if you have a shelf and you took everything from it, will you have a default item there? Or will you have nothing there?

Andr1yK: It is the same here, if you do not have a selected task, then you have nothing there and definitely not some default object

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+    setSelectValue(value);
+  };
+
+  const hundleChange = (value: string) => {
+    setFilterValue(value);
+  };
+
+  const todosForRender
+    = todos
+      .filter(({ completed }) => {
```
Andr1yK, line 51:
```suggestion
    = todos.filter(({ completed }) => {
```

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+        switch (selectValue) {
+          case 'active':
```
Andr1yK, line 53: add 'active' and 'completed' to enum

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+            return !completed;
+
+          case 'completed':
+            return completed;
+
+          default:
+            return true;
+        }
+      })
+      .filter(({ title }) => (
```
Andr1yK, line 63: I think it is possible to combine these two filters and reduce the complexity

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,42 @@
...
+
+    getUserAsync();
+  }, [todo.id]);
+
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {isLoader ? (
```
Andr1yK, line 39:
```suggestion
      {loading ? (
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +45,42 @@ export const TodoModal: React.FC = () => {
...
+              {`Todo #${todo.id}`}
             </div>
 
-            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
             <button
+              aria-label="close"
               type="button"
               className="delete"
               data-cy="modal-close"
+              onClick={() => onSelect(defaultTodo)}
```
Andr1yK, line 56:
```suggestion
              onClick={() => onSelect(null)}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +45,42 @@ export const TodoModal: React.FC = () => {
...
+                className={classNames(
+                  { 'has-text-danger': !todo.completed },
+                  { 'has-text-success': todo.completed },
+                )}
+              >
+                {
+                  todo.completed
+                    ? 'Done'
+                    : 'Planned'
+                }
```
Andr1yK, line 77:
```suggestion
                {todo.completed
                    ? 'Done'
                    : 'Planned'
                }
```

Andr1yK (changes requested):
image

What is this 0 doing here?🤨

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+    const todosFromAPI = await getTodos();
+
+    setTodos(todosFromAPI);
+  }
+
+  useEffect(() => {
+    getTodosFromServer();
+  }, []);
+
+  async function selectUser(todoId: number) {
```
Anna-Tkachenko, line 31: can we accept user id here instead of todo id?

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+    const filteredTodos = todos.filter(todo => todo.id === todoId);
+    const { userId } = filteredTodos[0];
+    const userFromServer = await getUser(userId);
+
+    setSelectedUser(userFromServer);
+  }
+
+  const selectTodo = (todoId: number) => {
+    const filteredTodos = todos.filter(todo => todo.id === todoId);
```
Anna-Tkachenko, line 40: if you need to find some element, use find method instead of the filter one

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+
+  const resetTodo = () => {
+    setSelectedTodo(null);
+    setSelectedUser(null);
+  };
+
+  const filteredTodos = todos
+    .filter(todo => {
+      switch (field) {
+        case 'active':
```
Anna-Tkachenko, line 54: let's create enum for possible field values

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+        case 'completed':
+          return todo.completed;
+          break;
+
+        default:
+          return true;
+      }
+    })
+    .filter(todo => todo.title.toLowerCase()
+      .includes(query.toLowerCase()));
```
Anna-Tkachenko, line 66: will be better to create separate variable for storing query in lower case before calling filter method

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,88 @@
...
+                <button
+                  data-cy="selectButton"
+                  className="button"
+                  type="button"
+                  value={id}
+                  onClick={handleClick}
+                >
+                  <span className="icon">
+                    <i className={classNames('far', {
+                      'fa-eye': id !== selectedTodo?.id,
```
Anna-Tkachenko, line 75: you check is current todo selected several times, let's create separate boolean variable inside the callback for storing this value

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+  };
+
+  useEffect(() => {
+    getTodosFromServer();
+  }, []);
+
+  const filterTodos = () => {
+    const todosMatchedQuery = todos.filter(({ title }) => {
+      const titleToLowerCase = title.toLowerCase();
+      const queryToLowerCase = query.toLowerCase();
```
Anna-Tkachenko, line 34: will be better to move this variable outside of the filter method

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,86 @@
...
+                <button
+                  data-cy="selectButton"
+                  className="button"
+                  type="button"
+                  onClick={() => onSelect(todo)}
+                >
+                  <span className="icon">
+                    <i
+                      className={classNames('far', {
+                        'fa-eye-slash': selectedTodo?.id === id,
```
Anna-Tkachenko, line 73: you checked is current todo selected several times, will be better to create separate variable for storing this value inside the callback

Anna-Tkachenko (approved):

PR #1151, opened 2022-11-02
React dynamic list of todos

glugovsky:
[DEMO LINK](https://glugovsky.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,63 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+
+import { getTodos } from './api';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { Todo } from './types/Todo';
+import { SortBy } from './types/SortBy';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  const [query, setQuery] = useState('');
+  const [sortBy, setSortBy] = useState<SortBy>(SortBy.All);
+
+  const selectTodo = useCallback((todo: Todo | null) => {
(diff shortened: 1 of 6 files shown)


VovaDzhuryk (approved):
GJ!

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+        default:
+          return true;
+      }
+    });
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
```
OleksiiBlezniuk, line 53: Would be nice to use async/await syntax instead of methods chaining

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+        setIsLoaded(true);
+      })
+      .catch(error => {
+        throw new Error(error);
+      });
```
OleksiiBlezniuk, line 58: Is there any sense to throw an error if this error is already thrown? Let's show some error message to user instead. In most cases user will not open console to check if there is an error (users usually don't know what is this at all)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,45 @@
...
+
+  useEffect(() => {
+    getUser(userId)
+      .then((userFromServer) => {
+        setUser(userFromServer);
+        setIsUserLoaded(true);
+      })
+      .catch(error => {
+        throw new Error(error);
+      });
```
OleksiiBlezniuk, line 35: same here about error

OleksiiBlezniuk (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+  useEffect(() => {
+    const getTodosFromServer = async () => {
+      try {
+        const todosFromServer = await getTodos();
+
+        setTodos(todosFromServer);
+        setIsLoaded(true);
+      } catch (error) {
+        // eslint-disable-next-line no-alert
+        alert('Error: todos could not be loaded');
```
OleksiiBlezniuk, line 59: okay, but would be nice to show some kind of styled message on UI (for example some red text instead of list)

but it's not critical 🙂

src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+          return completed;
+
+        case SortBy.Active:
+          return !completed;
+
+        default:
+          return true;
+      }
+    });
+  };
```
OleksiiBlezniuk, line 48: Probably there is no need in this function, you could just save filtered todos to constant

OleksiiBlezniuk (approved):

PR #1153, opened 2022-11-02
Develop

Dikey945:
[DEMO LINK](https://Dikey945.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,21 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  // const [loaded, setLoaded] = useState(false);
+
+  useEffect(() => {
+    getTodos()
+      .then(todosFromServer => (
+        setTodos(todosFromServer)
+      ));
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +30,23 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter todos={todos} setTodo={setTodos} />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0 && <Loader />}
+              <TodoList
+                todos={todos}
+                selectedTodo={selectedTodo}
+                setSelectedTodo={setSelectedTodo}
+              />
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal />
+      )}
     </>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,6 +13,8 @@ import { Todo } from './types/Todo';
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filterSelection, setFilterSelection] = useState('all');
+  const [query, setQuery] = useState('');
   // const [loaded, setLoaded] = useState(false);
 
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,45 @@
...
+  //   email
+  // } = user
+
+  useEffect(() => {
+    getUser(userId)
+      .then(userFromServer => {
+        setUser(userFromServer);
+        setIsUserLoaded(true);
+      });
+  }, []);
```
Anna-Borodaienko, line 36: I'm not sure, but - why is this logic here? You choose todo and it's user on the main page, here you just show this result

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
+  filterSelection,
+  setFilterSelection,
+  query,
+  setQuery,
+}) => {
+  const handleFilterSelection
+    = (event: React.ChangeEvent<HTMLSelectElement>) => {
+      event.preventDefault();
+      setFilterSelection(event.target.value);
+    };
```
Anna-Borodaienko, line 20: I didn't do prevent default here. Please, explain me, what do you prevent?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,45 @@
...
+  const {
+    id,
+    title,
+    completed,
+    userId,
+  } = selectedTodo;
+  // const {
+  //   name
+  //   email
+  // } = user
```
Anna-Borodaienko, line 28: delete it

src/components/TodoModal/TodoModal.tsx
Dikey945: For me, this logic is easier and the user used  only in this component, so I don't see the need to lift state up

src/components/TodoFilter/TodoFilter.tsx
Dikey945: agree here it redundant

src/App.tsx
```diff
@@ -17,18 +56,31 @@ export const App: React.FC = () => {
...
+                setFilterSelection={setFilterSelection}
+                query={query}
+                setQuery={setQuery}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0 && <Loader />}
```
Anna-Tkachenko, line 68: will be great to handle it with separate variable, that will be responsible for the data loading
for now if you'll receive an empty array from server, user will see infinite loader

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,59 @@ export const TodoList: React.FC = () => (
...
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+        const handleButtonClick = () => (
+          setSelectedTodo(todo)
+        );
```
Anna-Tkachenko, line 40: will be better to move it before the return statement

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,59 @@ export const TodoList: React.FC = () => (
...
+                  <i className="fas fa-check" />
+                </span>
+              )}
+            </td>
+            <td className="is-vcentered is-expanded">
+              <p
+                className={classNames(completed
+                  ? 'has-text-success'
+                  : 'has-text-danger')}
+              >
```
Anna-Tkachenko, line 61: in this case you shouldn't use classnames library
let's specify correct object here, and will add needed class depending on some condition
```suggestion
                className={classNames({
                  'has-text-success': completed,
                  'has-text-danger': !completed,
                }>
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,59 @@ export const TodoList: React.FC = () => (
...
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={handleButtonClick}
+              >
+                <span className="icon">
+                  <i className={classNames('far', {
+                    'fa-eye-slash': id === selectedTodo?.id,
+                    'fa-eye': id !== selectedTodo?.id,
```
Anna-Tkachenko, line 75: you check if current tab is selected several times, will be better to create separate boolean variable and store here the result of current check
and use this variable where you need it

src/App.tsx
Anna-Tkachenko: (not important)

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,57 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filterSelection, setFilterSelection] = useState('all');
+  const [query, setQuery] = useState('');
+  const [isTodosLoaded, setIsTodosLoaded] = useState(false);
+
+  useEffect(() => {
+    const getTodosFormServer = async () => {
+      const todosFromServer = await getTodos();
```
Denchepornyuk, line 22: what if getTodos return an error? You should catch an error here

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,41 @@
...
+  const [isUserLoaded, setIsUserLoaded] = useState(false);
+  const {
+    id,
+    title,
+    completed,
+    userId,
+  } = selectedTodo;
+
+  useEffect(() => {
+    getUser(userId)
```
Denchepornyuk, line 27: what if getTodos return an error? You should catch an error here

Denchepornyuk (changes requested):

Denchepornyuk (changes requested):
I don't see changes from previous review

Dikey945:
> I don't see changes from previous review

Need more sleep))) Forgot to push 🤦‍♂️

Denchepornyuk (approved):

PR #1160, opened 2022-11-03
implemented filter and selector, accepted data from server

SerhiiRomaniukWD:
[DEMO LINK](https://serhiiromaniukwd.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { FC, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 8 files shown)


SerhiiRomaniukWD:
I was left alone, I wrote to the group several times, but without results, that's why I'm without review

src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+    setSelectedTodo(todo);
+  };
+
+  const visibleTodos = todos?.filter(todo => {
+    const titleLowerCase = todo.title.toLowerCase();
+    const queryLowerCase = query.toLowerCase();
+    const isTitleIncludeQuery = titleLowerCase.includes(queryLowerCase);
+
+    switch (category) {
+      case 'completed':
```
Anna-Tkachenko, line 49: will be better to create enum for possible values

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,59 @@ export const TodoList: React.FC = () => (
...
+                type="button"
+                onClick={() => {
+                  if (todo) {
+                    chooseTodo(todo);
+                  }
+                }}
+              >
+                <i className={cn('far', {
+                  'fa-eye': selectedTodo?.id !== todo.id,
+                  'fa-eye-slash': selectedTodo?.id === todo.id,
```
Anna-Tkachenko, line 76: you checked is current todo selected several times, let's create boolean variable inside the map callback that will store this value

Anna-Tkachenko (approved):

PR #1162, opened 2022-11-04
started the project

ArtemHulkoDev:
[DEMO LINK](https://ArtemHulkoDev.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,4 +24,4 @@ works as expected;
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_dynamic-list-of-todos/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://ArtemHulkoDev.github.io/react_dynamic-list-of-todos/) and add it to the PR description.
--- a/src/components/TodoModal/TodoModal.tsx
+++ b/src/components/TodoModal/TodoModal.tsx
@@ -2,11 +2,13 @@ import React from 'react';
 import { Loader } from '../Loader';
 
 export const TodoModal: React.FC = () => {
+  const valid = true;
+
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {valid ? (
         <Loader />
       ) : (
         <div className="modal-card">


Oleksandra-Shevchenko (approved):
👍

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+  const [query, setQuery] = useState('');
+  const [todoId, setTodoId] = useState(0);
+  const [userId, setUserId] = useState(0);
+  const [todoTitle, setTodoTitle] = useState('');
+  const [completed, setCompleted] = useState(false);
+
+  const filteredTodos = () => {
+    const selectedTodos = todos.filter(todo => {
+      switch (selected) {
+        case 'active':
```
Anna-Tkachenko, line 25: let's create enum for available options here

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+
+        case 'completed':
+          return todo.completed;
+
+        default:
+          return todo;
+      }
+    });
+
+    return selectedTodos.filter(todo => todo.title.toLowerCase().match(query.toLowerCase()));
```
Anna-Tkachenko, line 36: will be better to create separate variable for storing query in lower case before calling filter method

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +33,62 @@ export const TodoList: React.FC = () => (
...
+            {todo.completed && (
+              <span className="icon" data-cy="iconCompleted">
+                <i className="fas fa-check" />
+              </span>
+            )}
+          </td>
+          <td className="is-vcentered is-expanded">
+            <p className={todo.completed
+              ? 'has-text-success'
+              : 'has-text-danger'}
```
Anna-Tkachenko, line 53: use classnames library if you need to set class depending on some check

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +33,62 @@ export const TodoList: React.FC = () => (
...
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => {
+                  setTodoId(todo.id);
+                  setTodoTitle(todo.title);
+                  setCompleted(todo.completed);
+                  setUserId(todo.userId);
+                }}
```
Anna-Tkachenko, line 82: can we just store the full selected todo in state instead of storing each separate value?

Anna-Tkachenko, line 82: can we just store the full selected todo in state instead of storing each separate value?

Anna-Tkachenko (changes requested):

src/App.tsx
ArtemHulkoDev: done

src/App.tsx
ArtemHulkoDev: done

src/components/TodoList/TodoList.tsx
ArtemHulkoDev: done

src/components/TodoList/TodoList.tsx
ArtemHulkoDev: done

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+
+enum Selected {
+  Active = 'active',
+  Completed = 'completed',
+}
```
OleksiiBlezniuk, line 15:
```suggestion
enum Selected {
  Active = 'active',
  Completed = 'completed',
  All = 'all',
}
```

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(true);
+  const [selected, setSelected] = useState('all');
```
OleksiiBlezniuk, line 20:
```suggestion
  const [selected, setSelected] = useState(Enum.All);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,70 @@ export const TodoList: React.FC = () => (
...
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          data-cy="todo"
+          className=""
```
OleksiiBlezniuk, line 34: do you really need to set empty string here?

OleksiiBlezniuk (approved):

PR #1169, opened 2022-11-06
implemented RDLT

VovaDzhuryk:
[DEMO LINK](https://VovaDzhuryk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -17,18 +68,35 @@ export const App: React.FC = () => {
...
+                  <Loader />
+                </div>
+              ) : (
+                <div className="block">
+                  <TodoList
+                    todos={filterTodos()}
+                    selectedTodoId={selectedTodoId}
+                    selectTodo={handleTodoSelect}
+                  />
+                </div>
```
glugovsky, line 91: You are a little bit repetitive here, both `Loader` and `TodoList` are inside `div.block`, so you could move conditional rendering inside of this block

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,61 @@ export const TodoList: React.FC = () => (
...
+            {!todo.completed
+              ? (
+                <td className="is-vcentered" />
+              ) : (
+                <td className="is-vcentered">
+                  <span className="icon" data-cy="iconCompleted">
+                    <i className="fas fa-check" />
+                  </span>
+                </td>
+              )}
```
glugovsky, line 52: `td.is-vcentered` is rendered in any case, you should probably only use conditional rendering for the inner span
```suggestion
            <td className="is-vcentered">
              {todo.completed && (
                <span className="icon" data-cy="iconCompleted">
                  <i className="fas fa-check" />
                </span>
              )}
            </td>
```

glugovsky (approved):

src/App.tsx
```diff
@@ -8,7 +8,58 @@ import { TodoFilter } from './components/TodoFilter';
...
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [dataIsLoaded, setDataIsLoaded] = useState(false);
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+  const [todoStatus, setTodoStatus] = useState<TodoStatus>(TodoStatus.All);
+  const [query, setQuery] = useState('');
+
+  const loadTodos = async () => {
+    const todosFromServer = await getTodos();
```
OleksiiBlezniuk, line 25: what if there will be an error?

src/App.tsx
```diff
@@ -8,7 +8,58 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+    setTodos(todosFromServer);
+    setDataIsLoaded(true);
+  };
```
OleksiiBlezniuk, line 29: Well, better to use useCallback here, I think we don't want to create this function on every render

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,38 @@
...
+  todo: Todo | null;
+  selectTodo: (todoId: number) => void
+};
+
+export const TodoModal: React.FC<Props> = ({ todo, selectTodo }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [userIsLoaded, setUserIsLoaded] = useState(false);
+
+  const loadUser = async (userId: number) => {
+    const userFromServer = await getUser(userId);
```
OleksiiBlezniuk, line 17: what if there will be an error?

OleksiiBlezniuk (changes requested):
🥲 

Screenshot 2022-11-07 at 18 33 49

PR #1199, opened 2022-12-06
solution

pushkalov:
https://pushkalov.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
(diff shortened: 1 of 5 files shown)


Anastasiia-Svintsova (changes requested):
Fix comments please

src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
+      switch (filterBy) {
+        case 'active':
+          return !todoItem.completed;
+
+        case 'completed':
+          return todoItem.completed;
+
+        case 'all':
+        default:
+          return todoItem;
```
Anastasiia-Svintsova, line 47: Create an enum for filter types and use it everywhere instead of `active`, `completed`, and `all`

Anastasiia-Svintsova (changes requested):

PR #1202, opened 2022-12-07
Develop

andriy-shymkiv:
[DEMO LINK](https://andriy-shymkiv.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,65 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
-import { Loader } from './components/Loader';
+// import { TodoModal } from './components/TodoModal';
+// import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+
+import { getTodos } from './api';
+// import { User } from './types/User';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState<string>('');
+  const [selectedOption, setSelectedOption] = useState(0);
+  // const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    const fetchResponse = async () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -8,7 +8,56 @@ import { TodoFilter } from './components/TodoFilter';
...
+        return todo;
+
+      case 'Active':
+        return !todo.completed;
+
+      case 'Completed':
+        return todo.completed;
+
+      default:
+        return 0;
```
iryna-lelyk, line 46: Better to return just `todo` by default.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,96 @@
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      <tbody>
+        {todos.map(todo => (
+          <tr
+            data-cy="todo"
+            className=""
```
iryna-lelyk, line 38: When todo, you're mapping is equal to selectedTodo, you should add class `has-background-info-light`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,51 @@
...
+
+    getUserFromServer();
+  }, []);
+
+  const findClickedTodoBySelectedId = useCallback((todoId: number) => {
+    const findedTodo = todos.find(todo => todo.id === todoId);
+
+    return findedTodo;
+  }, []);
+
```
iryna-lelyk, line 43: You can save selected todo, not only id of it. And if you've saved whole todo, you would have access to userId also

iryna-lelyk (approved):
Good job!

src/App.tsx
```diff
@@ -8,7 +8,56 @@ import { TodoFilter } from './components/TodoFilter';
...
 import { Loader } from './components/Loader';
 
+import { Todo } from './types/Todo';
+
+import { getTodos } from './api';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [userId, setUserId] = useState<number>(0);
+  const [query, setQuery] = useState<string>('');
```
mykola-koval, line 18: don't need the generic here, actually
it is automatically calculated based on the default value
```suggestion
  const [query, setQuery] = useState('');
```

src/components/TodoList/TodoList.tsx
mykola-koval: agree, add it

PR #1204, opened 2022-12-08
react_dynamic-list-of-todos

Blenemy:
[DEMO LINK](https://Blenemy.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,14 +14,17 @@ export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [status, setStatus] = useState('all');
   const [query, setQuery] = useState('');
-  const [isLoading, setIsLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
   const [selectedTodo, setSelectedTodo] = useState<Todo | undefined>();
 
   useEffect(() => {
-    getTodos().then(result => {
-      setTodos(result);
-      setIsLoading(true);
-    });
+    getTodos()
+      .then(result => {
+        setTodos(result);
+    })
+      .finally(() => {
+        setIsLoading(false);
+      });
   }, []);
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -17,18 +56,38 @@ export const App: React.FC = () => {
...
-              <TodoList />
+              {!isLoading ? (
+                <Loader />
+              ) : (
+                <TodoList
+                  todos={visibleTodos}
+                  selectedTodo={selectedTodo}
+                  setSelectedTodo={setSelectedTodo}
+                />
+              )}
```
qoa11a, line 76: Broken logic. Loader should be shown if isLoading is true.

src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+  const [status, setStatus] = useState('all');
+  const [query, setQuery] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | undefined>();
+
+  useEffect(() => {
+    getTodos().then(result => {
+      setTodos(result);
+      setIsLoading(true);
+    });
```
qoa11a, line 24: You should setIsLoading to false when the loading is done. And you should do it in .finally.

qoa11a (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
+  const [selectedTodo, setSelectedTodo] = useState<Todo | undefined>();
+
+  useEffect(() => {
+    getTodos()
+      .then(result => {
+        setTodos(result);
+    })
+      .finally(() => {
+        setIsLoading(false);
+      });
```
qoa11a, line 27:
```suggestion
    getTodos()
      .then(result => {
        setTodos(result);
      })
      .finally(() => {
        setIsLoading(false);
      });
```

qoa11a (approved):

PR #1209, opened 2022-12-08
dynamicTodoList

sh1gatsu:
DEMO LINK: https://sh1gatsu.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,61 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosStatus, setTodosStatus] = useState<boolean | null>(null);
+  const [todosQuery, setTodosQuery] = useState('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos().then(setTodos);
+  }, []);
+
+  const setterOfSelectedTodo = (pickedTodo: Todo | null) => {
+    setSelectedTodo(pickedTodo);
+  };
+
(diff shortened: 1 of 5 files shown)


Illia-Skladnik (approved):
lgtm

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosStatus, setTodosStatus] = useState<boolean | null>(null);
```
mykola-koval, line 15: let's use enum instead of boolean | null ?
```suggestion
  const [todosStatus, setTodosStatus] = useState<boolean | null>(null);
```

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+  const [todosQuery, setTodosQuery] = useState('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos().then(setTodos);
+  }, []);
+
+  const setterOfSelectedTodo = (pickedTodo: Todo | null) => {
+    setSelectedTodo(pickedTodo);
+  };
```
mykola-koval, line 25: why do you need this function if you can use `setSelectedTodo` right away?

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+  if (todosStatus !== null) {
+    filteredTodos = filteredTodos
+      .filter(todoStatus => todoStatus.completed === todosStatus);
+  }
+
+  if (todosQuery.trim() !== '') {
+    const lowerQuery = todosQuery.toLocaleLowerCase().trim();
+
+    filteredTodos = filteredTodos
+      .filter(todoQuery => todoQuery.title.toLocaleLowerCase().trim().includes(lowerQuery));
```
mykola-koval, line 56: i guess you don't have to trim todos title here, no strong opinion tho

sh1gatsu: > i guess you don't have to trim todos title here, no strong opinion tho

lets leave it, i like to add some spaces before entering smth ;)

src/App.tsx
sh1gatsu: > why do you need this function if you can use `setSelectedTodo` right away?

my mistake, its bad naming of it, this is Lifting state up function, thats why i need to use it like this

mykola-koval (approved):

PR #1211, opened 2022-12-09
First solution

taraselnik:
[DEMO LINK](https://taraselnik.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,40 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoadedTodos, setIsLoadedTodos] = useState(false);
+  const [checkedTodo, setCheckedTodo] = useState<Todo | null>(null);
+  const [inputState, setInputState] = useState('');
+  const [toggleFilter, setToggleFilter] = useState<boolean | null>(null);
+
+  let filteredTodos = [...todos];
+
+  const fetchData = useCallback(async () => {
+    const data = await getTodos();
+
+    setTodos(data);
+    setIsLoadedTodos(true);
+  }, []);
+
+  useEffect(() => {
+    fetchData();
+  }, []);
+
+  if (inputState && inputState.length > 0) {
+    const normInputState = inputState.toLocaleLowerCase().trim();
+
+    filteredTodos = filteredTodos.filter(todo => todo.title.toLocaleLowerCase()
+      .includes(normInputState));
+  }
+
+  if (toggleFilter !== null) {
+    filteredTodos = filteredTodos.filter(todo => todo.completed === toggleFilter);
+  }
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


taraselnik:
image

src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoadedTodos, setIsLoadedTodos] = useState(false);
+  const [checkedTodo, setCheckedTodo] = useState<Todo | null>(null);
+  const [inputState, setInputState] = useState('');
+  const [toggleFilter, setToggleFilter] = useState<boolean | null>(null);
```
masha-girya, line 18: you can set here 'all' select option by default. Also, name toggle doesn`t suitable in that case, because here is not a toggle but options in select. So good name will be 'option, setOption' :)

src/App.tsx
```diff
@@ -17,18 +49,34 @@ export const App: React.FC = () => {
...
+                inputState={inputState}
+                setInputState={setInputState}
+                setToggleFilter={setToggleFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!isLoadedTodos
```
masha-girya, line 60: it can be checked by todos.length, so 'isLoadedTodos' variable will be not needded anymore

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,78 @@
...
+import React from 'react';
+
+type Props = {
+  inputState: string
+  setInputState: (inputState: string) => void
+  setToggleFilter: (toggleFilter: boolean | null) => void
+};
+
+export const TodoFilter: React.FC<Props> = ({
+  inputState, setInputState, setToggleFilter,
```
masha-girya, line 10:
```suggestion
  inputState, 
  setInputState, 
  setToggleFilter,
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,78 @@
...
+}) => {
+  const getToggleValue = (value: string) => {
+    switch (value) {
```
masha-girya, line 13: you can do switch case right into filter in App file

masha-girya: if you will do like i suggested, then toggleValue and toggleFilter will unit, because they refer to the same logic

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,78 @@
...
+      default:
+        setToggleFilter(null);
+        break;
+    }
+  };
+
+  return (
+    <form
+      className="field has-addons"
+      onSubmit={e => e.preventDefault()}
```
masha-girya, line 35: nice mention!

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,78 @@
...
+    >
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
```
masha-girya, line 40: i thenk, here could be value={toggle}, where be defaul toggle is 'all' option

src/components/TodoList/TodoList.tsx
```diff
@@ -16,7 +28,53 @@ export const TodoList: React.FC = () => (
...
+              type="button"
+              onClick={() => onSetCheckedTodo(todo)}
+            >
+              <span className="icon">
+                <i className={classNames(
+                  'far',
+                  {
+                    'fa-eye': todo !== checkedTodo,
+                    'fa-eye-slash': todo === checkedTodo,
+                  },
```
masha-girya, line 68:
```suggestion
                <i className={  todo === checkedTodo 
                  ? 'far fa-eye-slash' 
                  : 'far fa-eye' }
```

masha-girya (changes requested):
i leave some comments, I think there could be less variables in state, so I believe you can fix it)

src/App.tsx
mykola-koval: totally agree,
let's use an enum for that
```suggestion
  const [filter, setFilter] = useState<FilterType>(FilterType.None);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,80 @@
...
+  setInputState,
+  setToggleFilter,
+}) => {
+  const getToggleValue = (value: string) => {
+    switch (value) {
+      case 'active':
+        setToggleFilter(false);
+        break;
+
+      case 'completed':
```
mykola-koval, line 20: use enum values

mykola-koval (approved):

PR #1227, opened 2022-12-12
list of todos created

Esceype:
- [DEMO LINK](https://Esceype.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,100 +1,92 @@
 import React from 'react';
+import { Todo } from '../../types/Todo';
 
-export const TodoList: React.FC = () => (
-  <table className="table is-narrow is-fullwidth">
-    <thead>
-      <tr>
-        <th>#</th>
-        <th>
-          <span className="icon">
-            <i className="fas fa-check" />
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+type Props = {
+  todos: Todo[],
+  onTodoUserId: (todoUserId: number | null) => void,
+  modalStatus: boolean,
+  onModalStatus: (modalStatus: boolean) => void,
+  onSelectedTodo: (selectedTodo: Todo | null) => void,
+  selectedTodo: Todo | null,
+};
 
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,92 @@
...
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map(todo => (
+          <tr
+            data-cy="todo"
+            className=""
```
referee1988, line 42: In the sample page, the active todo in the todos table under the modal is given the class 'has-background-info-light' and is lightly highlighted in blue.
download

But on your page, the active todo under the modal window does not change the class.
download

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,92 @@
...
+                  <td className="is-vcentered is-expanded">
+                    <p className="has-text-success">{todo.title}</p>
+                  </td>
+                </>
+              ) : (
+                <>
+                  <td className="is-vcentered" />
+                  <td className="is-vcentered is-expanded">
+                    <p className="has-text-danger">{todo.title}</p>
+                  </td>
```
referee1988, line 63: Lines 54-56 and 61-63 are the same except for the class. For conditional class definition it is better to use classNames

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,92 @@
...
+              )}
+            <td className="has-text-right is-vcentered">
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => {
+                  onTodoUserId(todo.userId);
+                  onModalStatus(!modalStatus);
+                  onSelectedTodo(todo);
```
referee1988, line 74: Lots of logic in the onClick event. It's better to put it in a separate handler function.

referee1988 (changes requested):
Good job, but there are a few remarks.

referee1988 (approved):
Great job!

Olexandr6 (approved):

PR #1236, opened 2022-12-13
the react dynamic-list-of-todos is done

Anastasiya145:
[DEMO LINK](https://Anastasiya145.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -8,7 +8,76 @@ import { TodoFilter } from './components/TodoFilter';
...
+    setTodosLoaded(false);
+
+    setTodos(reponse);
+  }
+
+  useEffect(() => {
+    LoadTodos();
+  }, []);
+
+  const onChangeFilterType = (type: SortType) => {
```
TarasHoliuk, line 41: Actually, seems like now we don't need this wrapper at all. you can pass `setFilterType` to `onChangeFilterType` prop

`onChangeFilterType={setFilterType}`

src/App.tsx
```diff
@@ -8,7 +8,76 @@ import { TodoFilter } from './components/TodoFilter';
...
+    const filteredByFilter = [...todos].filter(todo => {
+      switch (type) {
+        case SortType.Active:
+          return todo.completed === false;
+
+        case SortType.Completed:
+          return todo.completed === true;
+
+        default:
+          return todo;
```
TarasHoliuk, line 55:
```suggestion
          return true;
```
The filter's callback should return a boolean value. `todo` will be transformed to `boolean` automatically (always `true`), but it's better to do it explicitly.

src/App.tsx
```diff
@@ -8,7 +8,76 @@ import { TodoFilter } from './components/TodoFilter';
...
+    });
+
+    let visibleGoods = filteredByFilter;
+
+    if (input) {
+      visibleGoods = filteredByFilter.filter(todo => todo
+        .title.toLocaleLowerCase().includes(query.toLowerCase().trim()));
+    }
+
+    return visibleGoods;
```
TarasHoliuk, line 66: you can use ternary here in this way:

```
return input
  ? filteredByFilter.filter(// your callback)
  : filteredByFilter;
```

src/App.tsx
```diff
@@ -8,7 +8,76 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+  const visibleTodos = useMemo(() => {
+    return getFilteredGoods({ filterType, query });
+  }, [filterType, query, todos]);
+
+  const handleClick = (id: number | null) => {
+    setSelectedTodo(id);
+  };
+
+  const onChangeInput = (str: string) => {
```
TarasHoliuk, line 77: the same - we can pass `setQuery(str);` to `onChangeInput` prop

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,18 +34,23 @@ export const TodoFilter = () => (
...
       </span>
 
       <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
-        />
+        {valueInput && (
```
TarasHoliuk, line 45: I'd move `span` wrapper inside this condition as well (at least it'll fix one test)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,36 @@
...
+  selectedTodo: Todo,
+};
+
+export const TodoModal: React.FC<Props> = ({
+  handleClick,
+  selectedTodo,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  async function LoadUser() {
```
TarasHoliuk, line 19:
```suggestion
  async function loadUser() {
```

TarasHoliuk (changes requested):

Anastasiya145 (commented):
Thanks for reviewing! I have changed all comments except 1st one. The onChangeFilterType doesn't work as you proposed above. So I left it

src/App.tsx
```diff
@@ -8,7 +8,70 @@ import { TodoFilter } from './components/TodoFilter';
...
+  };
+
+  function getFilteredGoods({ filterType: type, query: input }: ReorderOptions) {
+    const filteredByFilter = [...todos].filter(todo => {
+      switch (type) {
+        case SortType.Active:
+          return todo.completed === false;
+
+        case SortType.Completed:
+          return todo.completed === true;
```
AlisaVlasova, line 52:
```suggestion
        case SortType.Active:
          return !todo.completed;

        case SortType.Completed:
          return todo.completed;
```

AlisaVlasova (approved):

PR #1251, opened 2022-12-30
Develop

UmizDemud:
[DEMO LINK](https://UmizDemud.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


naziamoff (approved):
I have a couple of minor suggestions
But that is a great job, I like the way you styled it as well 👍

src/components/TodoList/TodoList.tsx
```diff
@@ -46,9 +46,10 @@ export const TodoList: React.FC<Props> = ({
               )}
           </td>
           <td className="is-vcentered is-expanded">
-            <p className={todo.completed
-              ? 'has-text-success'
-              : 'has-text-danger'}
+            <p className={classNames(
+              { 'has-text-success': todo.completed },
+              { 'has-text-danger': todo.completed },
```
meforoma, line 51: condition is the same :)

src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+        return todos
+          .filter(todo => !todo.completed
+              && todo.title.toLowerCase().includes(cleanQuery));
+      case 'completed':
+        return todos
+          .filter(todo => todo.completed
+              && todo.title.toLowerCase().includes(cleanQuery));
+      default:
+        return todos
+          .filter(todo => todo.title.toLowerCase().includes(cleanQuery));
```
meforoma, line 39: feels too much of repeats
this seems to be valid - see if that fits your implementation
```suggestion
    return todos.filter(({ title, completed }) => (
      title.toLowerCase().includes(cleanQuery)
      && completed === (filterType === 'active')
    ))
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,51 @@ export const TodoList: React.FC = () => (
...
+              type="button"
+              onClick={() => onSelect(todo.id)}
+            >
+              <span className="icon">
+                <i className={classNames(
+                  'far',
+                  { 'fa-eye': selectedTodo !== todo.id },
+                  { 'fa-eye-slash': selectedTodo === todo.id },
+                )}
+                />
```
meforoma, line 70: to my preference ternary is more concise
```suggestion
                <i className={classNames(
                  'far',
                  selectedTodo === todo.id
                    ? 'fa-eye-slash'
                    : 'fa-eye'
                )}
                />
```

meforoma (changes requested):
love your entire implementation 👍🏼, usage of a modal, styling.
One bug and one suggestion to test and you're all good.

src/App.tsx
UmizDemud: I actually thought along these lines, but decided to keep the switch, in case I'd want to extend this or make changes in the future. However I'll just simplify because it is like 1/3rd of the code

meforoma (approved):
see one final suggestion - otherwise good job 👍🏼

PR #1275, opened 2023-01-18
solution

nikachu404:


src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodoId, setSelectedTodoId] = useState<number>(0);
```
vadiimvooo, line 15: It's not necessary to write type number here.

vadiimvooo (approved):

src/App.tsx
solaryasha: I totally agree with @vadiimvooo

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+        return todo.completed && isQueryMatchTitle;
+
+      case 'active':
+        return !todo.completed && isQueryMatchTitle;
+
+      case 'all':
+      default:
+        return todo && isQueryMatchTitle;
+    }
+  });
```
solaryasha, line 46: Let's wrap this results with `useMemo`

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+  useEffect(() => {
+    getTodos()
+      .then(setTodos);
+  }, []);
+
+  const closeModal = useCallback(() => (
+    setSelectedTodoId(0)
+  ), []);
+
+  const visibleTodos = todos.filter(todo => {
```
solaryasha, line 30: It would be great if we don't filter the array, if no query passed

src/App.tsx
```diff
@@ -17,18 +53,36 @@ export const App: React.FC = () => {
...
-              <Loader />
-              <TodoList />
+              {!todos.length
+                ? <Loader />
+                : (
+                  <TodoList
+                    todos={visibleTodos}
+                    selectedTodo={selectedTodo}
+                    onSelectTodo={(id: number) => setSelectedTodoId(id)}
+                  />
```
solaryasha, line 72: Don't use inliners
You have two options:

1. pass `setState` callback, this function never changes, so you don't have useless re-renders.
2. create a specific function that takes value and triggers `setState` function with that value. But you need to wrap this function with useCallback (and you will get render optimization) - that is what you actually did and it's okay

If you pass `(value) => setState(value)`- it will cause re-renders of the TodoList

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,28 @@
...
+import React from 'react';
+
+type Props = {
+  filter: string;
+  query: string;
+  onFilterChange: (filter: string) => void;
+  onQueryChange: (query: string) => void;
+};
+
+export const TodoFilter: React.FC<Props> = (
```
solaryasha, line 10: What about `React.memo` ?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,78 @@
...
-    <tbody>
-      <tr data-cy="todo" className="">
-        <td className="is-vcentered">1</td>
-        <td className="is-vcentered" />
-        <td className="is-vcentered is-expanded">
-          <p className="has-text-danger">delectus aut autem</p>
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
+export const TodoList: React.FC<Props> = (
```
solaryasha, line 11: What about `React.memo` ?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,81 @@
...
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+import { getUser } from '../../api';
+
+type Props = {
+  selectedTodo: Todo;
+  onCloseModal: () => void;
+};
+
+export const TodoModal: React.FC<Props> = ({ selectedTodo, onCloseModal }) => {
```
solaryasha, line 13: What about the `React.memo` ?

src/App.tsx
solaryasha: Also, you could try to apply a debounce for filtering todos by query (it's not required, but it would be fun)

solaryasha (changes requested):

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,83 @@
...
+      completed,
+      id,
+      title,
+      userId,
+    } = selectedTodo;
+    const [user, setUser] = useState<User | null>(null);
+
+    useEffect(() => {
+      getUser(userId)
+        .then(setUser);
```
valerii-derkach, line 25: what about errors?

src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
+  const [query, setQuery] = useState('');
+  const [debouncedQuery, setDebouncedQuery] = useState('');
+
+  const selectedTodo = useMemo(() => (
+    todos.find(todo => todo.id === selectedTodoId)
+  ), [selectedTodoId]);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos);
```
valerii-derkach, line 29: what about errors?

valerii-derkach (changes requested):
check your debounce functionality to see if it works as we expect?

valerii-derkach (approved):

PR #1282, opened 2023-01-19
solution

victoria-kovalenko:
[DEMO LINK](https://victoria-kovalenko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
+/* eslint-disable no-nested-ternary */
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+      .then(todo => {
+        setTodos(todo);
+      });
+  }, []);
+
+  const preaperedSearchQuery = searchQuery.toLowerCase();
+
+  const visibleTodo = todos.filter(todo => {
+    const searchingTodo = todo.title.toLowerCase()
+      .includes(preaperedSearchQuery);
```
romanblyashuk94, line 29:
```suggestion
    const isSearchQueryMatched = todo.title.toLowerCase()
      .includes(preaperedSearchQuery);
```

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+        case 'completed':
+          return todo.completed === false;
+        default:
+          break;
+      }
+
+      return value;
+    };
+
+    const filteredTodo = swithCase(filter);
```
romanblyashuk94, line 48: it's something strange.

you create by every iteration swithCase function accepts some magic "value"

after this you call this function and pass selected filter option to it

after this you put result of this function to "filterTodo" variable, but this function returns not todo, but boolean value.

i dont understand nothing.

lets refactor this.

you dont need to create any functions here

just write switch case here that return boolean value for every todo, based on todo.completed and isSearchQueryMatched values

src/App.tsx
```diff
@@ -17,18 +58,40 @@ export const App: React.FC = () => {
...
       </div>
 
-      <TodoModal />
+      {todoId !== 0 && (
+        <TodoModal
+          todoId={todoId}
+          todos={todos}
+          selectTodo={(value: number) => {
+            setTodoId(value);
+          }}
```
romanblyashuk94, line 92: Todo modal no need all todos and selected todo id.

just find selected Todo and pass it to modal
```suggestion
        <TodoModal
          selectedTodo={selectedTodo}
          onClose={(value: number) => setTodoId(0)}
```

romanblyashuk94 (changes requested):

PR #1289, opened 2023-01-20
Solution

vrima1221:
[DEMO LINK](https://vrima1221.github.io/react_dynamic-list-of-todos/)


NadiiaKoch (changes requested):
Everything is okay! Just use React.memo, useCallback, useMemo in this task.

PR #1315, opened 2023-01-23
React dynamic list of todos

OlexandraKorol:
[DEMO LINK](https://OlexandraKorol.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
 import 'bulma/css/bulma.css';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+        case 'completed':
+          return filterByTitle && completed;
+          break;
+
+        case 'all':
+        default:
+          return filterByTitle;
+          break;
+      }
+    });
```
romanblyashuk94, line 56: it would be better to move filter logic to helper function

romanblyashuk94 (approved):
good job 👍

src/App.tsx
Ihor-Karpyn: Agree

Ihor-Karpyn (approved):
Good job

PR #1316, opened 2023-01-23
dynamic lists with debounce

nurdanokcu:
[DEMO LINK](https://nurdanokcu.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,44 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+  const [preparedTodos, setPreparedTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isLoadingFailed, setIsLoadingFailed] = useState(false);
+
+  const getTodosFromServer = async () => {
+    setIsLoading(true);
+    try {
+      const todosFromServer = await getTodos();
+
+      setIsLoading(false);
+      setTodos(todosFromServer);
+      setPreparedTodos(todosFromServer);
+    } catch (error: unknown) {
+      setIsLoadingFailed(true);
+      setIsLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    getTodosFromServer();
+  }, []);
+
+  const selectTodo = (todoId: number) => {
+    setSelectedTodoId(todoId);
+  };
+
+  const closeModal = () => {
+    setSelectedTodoId(0);
+  };
+
+  const selectedTodo = todos.find(todo => todo.id === selectedTodoId) as Todo;
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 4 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,109 @@
...
+
+    if (debouncedQuery) {
+      filteredTodos = todosToFilter.filter(
+        todo => todo.title.toLowerCase().includes(debouncedQuery.toLowerCase()),
+      );
+    }
+
+    if (status !== 'all') {
+      filteredTodos = filteredTodos.filter(
+        todo => todo.completed === (status === 'completed'),
```
Lun4trick, line 35: haha very smart, but maybe a switch is more readable here. :)

Lun4trick (approved):
Good Job!

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+
+  const selectTodo = (todoId: number) => {
+    setSelectedTodoId(todoId);
+  };
+
+  const closeModal = () => {
+    setSelectedTodoId(0);
+  };
+
+  const selectedTodo = todos.find(todo => todo.id === selectedTodoId) as Todo;
```
naziamoff, line 46: why do you need type conversion here?

src/App.tsx
```diff
@@ -17,18 +53,40 @@ export const App: React.FC = () => {
...
-              <TodoList />
+              {isLoading && (
+                <Loader />
+              )}
+
+              {isLoadingFailed && (
+                <p>No todos were loaded!</p>
+              )}
+
+              {!isLoading && !isLoadingFailed && (
```
naziamoff, line 71: seems quite wordy for a positive result 😄 
I'd say:
```suggestion
              {areTodosReady && (
```

src/components/TodoFilter/TodoFilter.tsx
naziamoff: yup, I'd create enum + helper with switch for this

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,109 @@
...
+    event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>,
+  ) => {
+    if (event.target.name === 'title') {
+      setQuery(event.target.value);
+      applyQuery(event.target.value);
+    }
+
+    if (event.target.name === 'status') {
+      setStatus(event.target.value);
+    }
```
naziamoff, line 56: comparing for the exact string match usually looks hardcoded in the if conditions

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,49 @@ export const TodoList: React.FC = () => (
...
+          <td className="is-vcentered">{id}</td>
+          <td className="is-vcentered">
+            {completed && (
+              <span className="icon" data-cy="iconCompleted">
+                <i className="fas fa-check" />
+              </span>
+            )}
+          </td>
+          <td className="is-vcentered is-expanded">
+            <p className={cn({
```
naziamoff, line 48: those two conditions are mutually exclusive, so we can assign one of them as the default one, and one as conditional. And it will switch itself this way

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,49 @@ export const TodoList: React.FC = () => (
...
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => selectTodo(id)}
+            >
+              <span className="icon">
+                <i className={cn('far', {
+                  'fa-eye': id !== selectedTodoId,
+                  'fa-eye-slash': id === selectedTodoId,
```
naziamoff, line 66: same here

naziamoff (changes requested):
I really like your approach - the code is clear and ergonomic 🔥 
Still, there's some space for improvements

src/App.tsx
nurdanokcu: I added it because the result of the find method could be undefined and I am passing selectedTodo to the TodoModal component as a prop. It gives me an error when I pass it .
 If todo is listed in the list we know that it exists this is why we are sure that we will find it  so I decided to go with type conversion.

src/components/TodoList/TodoList.tsx
nurdanokcu: Hi Nazar I added it this way: 
```
className={cn('far fa-eye', {
                  'fa-eye-slash': id === selectedTodoId,
                })}
```
but it fails me in tests. Because tests are checking that I far-eye class shouldn' t exist when far-eye-slash class is there.

src/App.tsx
naziamoff: @nurdanokcu I guess this should work if you make it empty array if result of `find` is undefined

naziamoff (approved):
amazing job 🔥

PR #1328, opened 2023-01-30
react dynamic list of todos

Mohameddahi:
[DEMO LINK](https://mohameddahi.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+    return todos.find(todo => todo.id === id) as Todo;
+  };
+
+  const selectTodoId = (id: number) => {
+    setTodoId(id);
+  };
+
+  const selectUserId = (id: number) => {
+    setUserId(id);
+  };
```
meforoma, line 30: these are generally redundant
unless you pursue to shorten the onClick prop drilling
and yeah, you do

then you can at least single-line such :)
```suggestion
  const selectTodoId = (id: number) => { setTodoId(id) };
  const selectUserId = (id: number) => { setUserId(id) };
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,7 +30,76 @@ export const TodoList: React.FC = () => (
...
+                >
+                  <span className="icon">
+                    <i className="far fa-eye" />
+                  </span>
+                </button>
+              </td>
+            </tr>
+          )
+      ))}
+      {/* <tr data-cy="todo" className="">
```
meforoma, line 102: don't keep commented code

src/components/TodoList/TodoList.tsx
```diff
@@ -16,7 +30,76 @@ export const TodoList: React.FC = () => (
...
+              </td>
+              <td className="has-text-right is-vcentered">
+                <button
+                  data-cy="selectButton"
+                  className="button"
+                  type="button"
+                  onClick={() => {
+                    onSelectTodo(todo.id);
+                    onSelectUser(todo.userId);
+                  }}
```
meforoma, line 92: structure is the same, props differ
i believe you can DRY this code

meforoma (changes requested):
nice job 👏🏼 
effective i could approve it already.
But let's clean up the commented code and try to DRY the `<tr` element

src/components/TodoList/TodoList.tsx
Mohameddahi: yes absolutely right  , thanks for review!

PR #1337, opened 2023-02-02
Develop

Grayfox147:
[DEMO LINK](https://Grayfox147.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -68,6 +68,7 @@ export const App: React.FC = () => {
               <TodoList
                 todos={visibleTodos}
                 onSetSelectedTodo={setSelectedTodo}
+                selectedTodo={selectedTodo}
               />
             </div>
           </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 /* eslint-disable max-len */
 import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
+    getTodos().then(setAllTodos)
+
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
+
+  const visibleTodos = allTodos
+    .filter((todo) => {
+      const trimmed = query.toLocaleLowerCase().trim();
```
UmizDemud, line 32: You should move this line to line 29 (Before the `const visibleTodos`). It is the same trimmed query, but it is calculated multiple times in the loop

src/App.tsx
```diff
@@ -17,18 +54,33 @@ export const App: React.FC = () => {
...
+              />
             </div>
-
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && (
+                <Loader />
+              )}
+              <TodoList
```
UmizDemud, line 68: Not important in this project, but I'll write it. 

Probably better to render the list only if it is not loading. Something like; `isLoading ? ( ... ) : <TodoList />`
But again, it isn't important

UmizDemud (approved):
Approving but do fix that first issue as the icing on the cake. Would go a long way

src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [allTodos, setAllTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [query, setQuery] = useState('');
+  const [filter, setSelectedFilter] = useState('all');
```
naziamoff, line 19:
```suggestion
  const [selectedFilter, setSelectedFilter] = useState('all');
```

src/App.tsx
naziamoff: agree with @UmizDemud

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,83 @@
...
+                  onSetSelectedTodo(todo);
+                }}
+              >
+                <span className="icon">
+                  <i
+                    className={classNames(
+                      'far',
+                      {
+                        'fa-eye': todo !== selectedTodo,
+                        'fa-eye-slash': todo === selectedTodo,
```
naziamoff, line 71: basically those are about pretty much the same condition. You can assign one of them as default and another one in classnames. It is gonna toggle depending on the selected condition

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,84 @@
...
-      ) : (
-        <div className="modal-card">
-          <header className="modal-card-head">
-            <div
-              className="modal-card-title has-text-weight-medium"
-              data-cy="modal-header"
-            >
-              Todo #2
-            </div>
+      {!user
```
naziamoff, line 33: start booleans with positive

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,84 @@
...
+                }}
+              />
+            </header>
+            <div className="modal-card-body">
+              <p className="block" data-cy="modal-title">
+                {selectedTodo.title}
+              </p>
+
+              <p className="block" data-cy="modal-user">
+                {/* <strong className="has-text-success">Done</strong> */}
```
naziamoff, line 61:
```suggestion
```

naziamoff (approved):
not bad. Still got a couple of suggestions for you

src/App.tsx
Grayfox147: I understand! it is done! :D Thanks Umut

src/App.tsx
Grayfox147: Great! it is done :D, thanks guys!

src/App.tsx
Grayfox147: Done!

src/components/TodoModal/TodoModal.tsx
Grayfox147: Done!

src/components/TodoModal/TodoModal.tsx
Grayfox147: Done!

PR #1339, opened 2023-02-03
Develop

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,19 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todo, setTodo] = useState<Todo>();
+  // const [todos] = useState<Todo[]>();
+  const [appTodo, setAppTodo] = useState<Todo[]>([]);
+  const [check, setCheck] = useState(false);
+  const [loader, setLoader] = useState(true);
+
+  setTimeout(() => {
+    setLoader(false);
+  }, 500);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


mmahfuzi:
Dear reviewer, if you take a look at the test failures, you would see the reason of all failures(except one) is "timing out" and I think it's due to my different implementation!

src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
 
 export const App: React.FC = () => {
+  const [todo, setTodo] = useState<Todo>();
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [appTodo, setAppTodo] = useState<Todo[]>([]);
+  const [check, setCheck] = useState(false);
+  const [selectedTodoId, setSelectedTodoId] = useState<number>(0);
+
+  useEffect(() => {
+    selectTodo(selectedTodoId);
```
Grayfox147, line 21: Ah I understand you re render because your selectedTodo uses the getTodo function again, I see,.

src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+  });
+
+  useEffect(() => {
+    getTodos().then(setTodos);
+  }, []);
+
+  useEffect(() => {
+    setTodos(todos);
```
Grayfox147, line 29: Also I believe you dont need one, basically once you bring the todos all of them and set them in your state, you can use them to filter them etc, and then use the result for the todolist, I believe you dont need to use the useEffect again. I am not sure, although I did not need it.

src/App.tsx
```diff
@@ -17,18 +37,19 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter onFilter={setAppTodo} />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length
```
Grayfox147, line 44: I believe that we need to hide the loader inside the condition in case we dont have our information loaded , but the table can sit outside, basically if it has no information we will be able to see the table header empty. and the loader on top

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,91 @@
...
+
+type Props = {
+  onFilter: React.Dispatch<React.SetStateAction<Todo[]>>,
+};
+
+export const TodoFilter:React.FC<Props> = ({ onFilter }) => {
+  const [query, setQuery] = useState('');
+
+  const customFilter = useCallback(async (input: string) => {
+    if (input === 'all') {
```
Grayfox147, line 16: really interesting approach! never saw it before!! good job!

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,80 @@
...
+  todo: Todo,
+  onCheck: (isSelected: boolean) => void,
+  onSelect: (selectedTodoId: number) => void,
+};
+
+export const TodoModal: React.FC<Props> = ({ todo, onCheck, onSelect }) => {
+  const [user, setUser] = useState<User>();
+  const [loader, setLoader] = useState(true);
+
+  setTimeout(() => {
```
Grayfox147, line 18: interesting approach setting the loader inside

Grayfox147 (approved):
Overall quite different approach than mine, also different uses of the useEffect and the diynamic used with the fetching functions really impressive. I am not sure if the tests that you are not passing might be affected by the many fetching of Todos through the components, but overall really interesting approach.

Grayfox147 (changes requested):
I believe there are some improvements to do before, that will help us to pass the tests.

PR #1343, opened 2023-02-05
done

alaaashkar:
[DEMO LINK](https://alaaashkar.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,66 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import './app.scss';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+  const [selectedOption, setSelectedOption] = useState('all');
+  const [query, setQuery] = useState('');
+
+  const getTodosFromServer = async () => {
+    const todosFromServer = await getTodos();
+
+    setTodos(todosFromServer);
+  };
+
+  useEffect(() => {
+    getTodosFromServer();
+  }, []);
+
+  const selectTodo = (todoId: number) => {
+    setSelectedTodoId(todoId);
+  };
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+  const [selectedOption, setSelectedOption] = useState('all');
```
naziamoff, line 16: Selected option for what? Probably just `filterBy` and `setFilterBy`?

src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
+
+  let visibleTodos = [...todos];
+
+  if (selectedOption === 'completed') {
+    visibleTodos = visibleTodos.filter(todo => todo.completed);
+  }
+
+  if (selectedOption === 'active') {
+    visibleTodos = visibleTodos.filter(todo => !todo.completed);
+  }
```
naziamoff, line 55: create eNum for that. You should do that when working with multiple string values

src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
+
+  const selectOption = (option: string) => {
+    setSelectedOption(option);
+  };
+
+  const handleQuery = (value: string) => {
+    setQuery(value);
+  };
+
+  let visibleTodos = [...todos];
```
naziamoff, line 47: seems like this should be a state

src/App.tsx
```diff
@@ -17,18 +69,35 @@ export const App: React.FC = () => {
...
+                onHandleQuery={handleQuery}
+                query={query}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0
+                ? (<Loader />)
```
naziamoff, line 82:
```suggestion
                ? <Loader />
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,49 @@ export const TodoList: React.FC = () => (
...
+          })}
+        >
+          <td className="is-vcentered">{todo.id}</td>
+          {!todo.completed && <td className="is-vcentered" />}
+          {todo.completed && (
+            <td className="is-vcentered">
+              <span className="icon" data-cy="iconCompleted">
+                <i className="fas fa-check" />
+              </span>
+            </td>
```
naziamoff, line 46: the same element is repeated twice. Why not just conditionally render its children?

naziamoff (changes requested):
not bad, still got a work to do

PR #1347, opened 2023-02-07
Solution

RomeoKs:
[DEMO LINK](https://RomeoKs.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,57 @@
...
+      .then(data => {
+        setTodos(data);
+        setFilteredTodos(data);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
+
+  const onSelectTodos = (event: ChangeEvent<HTMLSelectElement>) => {
```
naziamoff, line 42: [naming](https://medium.com/swlh/how-to-better-name-your-functions-and-variables-e962a4ef335b)

naziamoff, line 42: really good idea to use `switch` here 🔥 
Will be even better if you create a enum for the filter type

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,56 @@
...
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+export const TodoFilter: React.FC<Props> = ({
+  query,
+  onInputQuery,
+  onSelectTodos,
+  onClearInput,
+}) => {
+  return (
```
naziamoff, line 16: seems you can return right away here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,140 @@
...
+              {todo.completed && (
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              )}
+            </td>
+            <td className="is-vcentered is-expanded">
+              <p className={
+                classNames(
+                  todo.completed ? 'has-text-success' : 'has-text-danger',
```
naziamoff, line 42: no ternaries in classnames library

naziamoff (approved):
LGTM 🔥 
Got a couple of suggestions

src/components/TodoList/TodoList.tsx
RomeoKs: Ooopsie....

src/App.tsx
```diff
@@ -1,14 +1,89 @@
...
+    setSearchActive(true);
+  };
+
+  const selectTodo = (todo: Todo) => {
+    setSelectedTodo(todo);
+  };
+
+  const closePopUp = () => {
+    setSelectedTodo(null);
+  };
```
naziamoff, line 35: I don't think you need those handlers, why not change state directly?

src/App.tsx
```diff
@@ -1,14 +1,89 @@
...
+      case 'active':
+        setFilteredTodos(todos.filter(todo => !todo.completed));
+        break;
+      case 'completed':
+        setFilteredTodos(todos.filter(todo => todo.completed));
+        break;
+      default:
+        setFilteredTodos(todos);
+    }
+  };
```
naziamoff, line 85: it is better to create enum for that and operate with it. String values are really bad options

naziamoff (approved):

PR #1355, opened 2023-02-13
Dynamic todo list was realised

anna-inozemets:
[DEMO LINK](https://anna-inozemets.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,69 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
+import { Todo } from './types/Todo';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+
+export const enum SelectByCompleted {
+  ALL = 'all',
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [selectByCompleted, setSelectByCompleted] = useState<string>(SelectByCompleted.ALL);
+  const [query, setQuery] = useState('');
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+
+  const loadTodos = async () => {
+    const loadedTodos = await getTodos();
+
+    setTodos(loadedTodos);
+    setIsLoaded(true);
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  let visibleTodos = [...todos];
+
+  switch (selectByCompleted) {
+    case SelectByCompleted.ACTIVE:
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+    case SelectByCompleted.COMPLETED:
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+    case SelectByCompleted.ALL:
+    default:
+      visibleTodos = [...todos];
+      break;
+  }
+
+  if (query) {
+    const normalizedQuery = query
+      .split(' ')
+      .map(word => word.trim().toLowerCase())
+      .join(' ');
+
+    visibleTodos = visibleTodos.filter(todo => todo.title.toLowerCase().includes(normalizedQuery));
+  }
+
+  const selectedTodo = todos.find(todo => todo.id === selectedTodoId) || null;
+
+  const handleCloseModal = () => {
+    setSelectedTodoId(0);
+  };
+
   return (
     <>
       <div className="section">
@@ -17,18 +72,32 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                onSelectByCompletedChange={(option) => setSelectByCompleted(option)}
+                query={query}
+                onQuery={(newQuery) => setQuery(newQuery)}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoaded
+                ? (
+                  <TodoList
+                    todos={visibleTodos}
+                    selectedTodoId={selectedTodoId}
+                    onSlelectedTodoId={(selectedId) => setSelectedTodoId(selectedId)}
+                  />
+                ) : (
+                  <Loader />
+                )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal todo={selectedTodo} onCloseModal={handleCloseModal} />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,8 +1,25 @@
-export const TodoFilter = () => (
+type Props = {
+  onSelectByCompletedChange: (option: string) => void,
+  query: string,
+  onQuery: (newQuery: string) => void,
+};
+
+export const TodoFilter: React.FC<Props> = (
+  {
+    onSelectByCompletedChange,
+    query,
+    onQuery,
+  },
+) => (
   <form className="field has-addons">
     <p className="control">
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
+          data-cy="statusSelect"
+          onChange={(event) => {
+            onSelectByCompletedChange(event.target.value);
+          }}
+        >
           <option value="all">All</option>
           <option value="active">Active</option>
           <option value="completed">Completed</option>
@@ -16,19 +33,24 @@ export const TodoFilter = () => (
         type="text"
         className="input"
         placeholder="Search..."
+        value={query}
+        onChange={(event) => onQuery(event.target.value)}
       />
       <span className="icon is-left">
         <i className="fas fa-magnifying-glass" />
       </span>
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
-        />
-      </span>
+      {query && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="clearSearchButton"
+            type="button"
+            className="delete"
+            onClick={() => onQuery('')}
+          />
+        </span>
+      )}
     </p>
   </form>
 );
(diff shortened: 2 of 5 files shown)


NVyrsta (approved):
Good job 👍

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +29,49 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr data-cy="todo" key={todo.id}>
```
NVyrsta, line 33: In the initial demo selected todo has class="has-background-info-light"

Screenshot_2

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
 import '@fortawesome/fontawesome-free/css/all.css';
 
+import { Todo } from './types/Todo';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+
+export const enum SelectByCompleted {
```
Anna-Tkachenko, line 13: will be better to name it somehow related for `todo status`, for now it's not clear, for what variable or property we can use this enum

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+  ALL = 'all',
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [selectByCompleted, setSelectByCompleted] = useState<string>(SelectByCompleted.ALL);
```
Anna-Tkachenko, line 22: add corresponding type here, it's not string if you added enum with limited number of values

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+  const [query, setQuery] = useState('');
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+
+  const loadTodos = async () => {
+    const loadedTodos = await getTodos();
```
Anna-Tkachenko, line 27: will be better to handle possible errors that can be caused while processing request to the server

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+    case SelectByCompleted.COMPLETED:
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+    case SelectByCompleted.ALL:
+    default:
+      visibleTodos = [...todos];
+      break;
+  }
```
Anna-Tkachenko, line 50: let's create separate function that will transform todos in the needed format

src/App.tsx
```diff
@@ -17,18 +72,32 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                onSelectByCompletedChange={(option) => setSelectByCompleted(option)}
```
Anna-Tkachenko, line 76: do you need to create wrappers here while passing the link to the function? can you just pass `setSelectByCompleted` function here?

src/App.tsx
```diff
@@ -17,18 +72,32 @@ export const App: React.FC = () => {
...
+                query={query}
+                onQuery={(newQuery) => setQuery(newQuery)}
```
Anna-Tkachenko, line 78: same here

src/App.tsx
```diff
@@ -17,18 +72,32 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoaded
+                ? (
+                  <TodoList
+                    todos={visibleTodos}
+                    selectedTodoId={selectedTodoId}
+                    onSlelectedTodoId={(selectedId) => setSelectedTodoId(selectedId)}
```
Anna-Tkachenko, line 88: same here about wrapper

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,25 @@
-export const TodoFilter = () => (
+type Props = {
+  onSelectByCompletedChange: (option: string) => void,
+  query: string,
+  onQuery: (newQuery: string) => void,
+};
+
+export const TodoFilter: React.FC<Props> = (
```
Anna-Tkachenko, line 7: how does it work? I don't see any import in this file 🐝

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,19 @@
 import React from 'react';
 
-export const TodoList: React.FC = () => (
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[],
+  onSlelectedTodoId: (selectedId:number) => void,
```
Anna-Tkachenko, line 8:
```suggestion
  onSelectedTodoId: (selectedId:number) => void,
```

src/components/TodoList/TodoList.tsx
Anna-Tkachenko: great point! check @NVyrsta comment

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +29,49 @@ export const TodoList: React.FC = () => (
...
+            ? (
+              <td className="is-vcentered">
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              </td>
+            )
+            : (
+              <td className="is-vcentered" />
+            )}
```
Anna-Tkachenko, line 45: looks like we need to render only inner span according to the `completed` property, not whole `td`

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +29,49 @@ export const TodoList: React.FC = () => (
...
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => onSlelectedTodoId(todo.id)}
+            >
+              <span className="icon">
+                <i className={classNames(
+                  'far',
+                  { 'fa-eye': selectedTodoId !== todo.id },
+                  { 'fa-eye-slash': selectedTodoId === todo.id },
```
Anna-Tkachenko, line 67: you checked is current todo selected several times, let's create separate variable for it

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,32 @@
...
+type Props = {
+  todo: Todo | null,
+  onCloseModal: () => void,
+};
+
+export const TodoModal: React.FC<Props> = ({ todo, onCloseModal }) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  const loadUser = async () => {
+    const loadedUser = await getUser(todo?.userId || 0);
```
Anna-Tkachenko, line 16: will be better to handle possible errors here

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+
+  const loadTodos = async () => {
+    try {
+      const loadedTodos = await getTodos();
+
+      setTodos(loadedTodos);
+      setIsLoaded(true);
+    } catch (error) {
+      throw new Error(`There is an error with loading data: ${error}`);
```
Anna-Tkachenko, line 29: if you throw an error here, that app will crash and user can't reload the data or work with your app
let's think about other way for handling the error here
what can we do, if api call will throw an error? how we can show it for the user without app crashing?

src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+    }
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
+  const visibleTodos = preparedTodos(todos, todoStatus, query);
```
Anna-Tkachenko, line 37: do you pass this array as an props to some component? can we optimize somehow rendering of child components?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,77 @@
...
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  onTodoStatus: (option: TodoStatus) => void,
```
Anna-Tkachenko, line 5: let's check the variable name, for now it's not clear what's the main action of this function
is it called on `TodoStatus` change, or is it change `TodoStatus`?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,36 @@
...
+export const TodoModal: React.FC<Props> = ({ todo, onCloseModal }) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  const loadUser = async () => {
+    try {
+      const loadedUser = await getUser(todo?.userId || 0);
+
+      setUser(loadedUser);
+    } catch (error) {
+      throw new Error(`There is an error with loading data: ${error}`);
```
Anna-Tkachenko, line 21: same here about errors processing, let's think how we can handle it without throwing new error

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+  const [isError, setIsError] = useState(false);
+
+  const loadTodos = async () => {
+    try {
+      const loadedTodos = await getTodos();
+
+      setTodos(loadedTodos);
+      setIsLoaded(true);
+    } catch (error) {
+      setIsError(true);
```
Anna-Tkachenko, line 31: should we change `isLoaded` variable in case of any error?

src/components/ErrorMessage/ErrorMeassage.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import 'bulma/css/bulma.css';
+import './ErrorMessage.scss';
+
+export const ErrorMeassage: React.FC = () => (
```
Anna-Tkachenko, line 5:
```suggestion
export const ErrorMessage: React.FC = () => (
```

src/utils/preparedTodos.ts
```diff
@@ -0,0 +1,35 @@
+import { TodoStatus } from '../types/TodoStatus';
+import { Todo } from '../types/Todo';
+
+export function preparedTodos(
```
Anna-Tkachenko, line 4: function name should contain some action
`preparedTodos` is good name for the variable, which will store the final array
let's think more about function naming

src/utils/preparedTodos.ts
```diff
@@ -0,0 +1,35 @@
...
+  switch (status) {
+    case TodoStatus.ACTIVE:
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+    case TodoStatus.COMPLETED:
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+    case TodoStatus.ALL:
+    default:
+      visibleTodos = [...todos];
```
Anna-Tkachenko, line 20: do we need to make additional copy here?

Anna-Tkachenko (approved):

PR #1357, opened 2023-02-14
Added the task solution

oleksusov:
[DEMO LINK](https://oleksusov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,46 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+
+import { Todo } from './types/Todo';
+
+enum FilterValues {
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedItem, setSelectedItem] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [filter, setFilter] = useState('all');
+
+  useEffect(() => {
+    getTodos()
+      .then(todosFromServer => setTodos(todosFromServer));
+  }, []);
+
+  let visibleTodos = todos.filter(todo => {
+    switch (filter) {
+      case FilterValues.Active:
+        return !todo.completed;
+
+      case FilterValues.Completed:
+        return todo.completed;
+
+      default:
+        return true;
+    }
+  });
+
+  if (query) {
+    visibleTodos = visibleTodos.filter(todo => {
+      return todo.title.toLowerCase().includes(query.toLowerCase());
+    });
+  }
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -17,18 +55,35 @@ export const App: React.FC = () => {
...
+                setQuery={setQuery}
+                filter={filter}
+                setFilter={setFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length > 0 ? (
```
OleziO, line 67: Replace with `todos.length`, it will be the same as `todos.length > 0`.

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,53 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
```
OleziO, line 32: Add `key` attribute.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +33,35 @@ export const TodoModal: React.FC = () => {
...
             <p className="block" data-cy="modal-user">
               {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              {selectedItem.completed ? (
+                <strong className="has-text-success">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
 
               {' by '}
```
OleziO, line 61: Better is to wrap this text on `<span>` tag.

OleziO (approved):

src/App.tsx
```diff
@@ -1,14 +1,52 @@
...
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedItem, setSelectedItem] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [filter, setFilter] = useState('all');
```
anstsot, line 24: it would be great to use `FilterValues` as type here and check all usages below
```suggestion
  const [filter, setFilter] = useState<FilterValues>(FilterValues.All);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,67 @@
...
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  query: string,
+  setQuery: (query: string) => void,
+  filter: string,
+  setFilter: (filter: string) => void,
```
anstsot, line 7: you can still use created enum `FilterValues` but it would be better to place it somewhere separately (not in App.tsx)
```suggestion
  filter: FilterValues,
  setFilter: (filter: FilterValues) => void,
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,67 @@
...
+            name="filterSelector"
+            data-cy="statusSelect"
+            value={filter}
+            onChange={(event => {
+              setFilter(event.target.value);
+            })}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
anstsot, line 30: same with `FilterValues` here
```suggestion
            <option value={FilterValues.All}>All</option>
            <option value={FilterValues.Active}>Active</option>
            <option value={FilterValues.Completed}>Completed</option>
```

anstsot (changes requested):

nazarmatsevych (approved):

PR #1358, opened 2023-02-14
solution

Malva37:
[DEMO LINK](https://Malva37.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable max-len */
-import React from 'react';
+// import React, { useCallback, useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +8,54 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [filterBy, setFilterBy] = useState('');
+
+  let visibleTodos = [...todos];
+
+  if (query) {
+    const preparedQuery = query.trim().toLowerCase();
+
+    visibleTodos = visibleTodos.filter(todo => todo.title.toLowerCase().includes(preparedQuery));
+  }
+
+  switch (filterBy) {
+    case 'active':
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+    case 'completed':
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+    case 'all':
+    default:
+      visibleTodos = [...visibleTodos];
+      break;
+  }
+
+  const fetchData = async () => {
+    const data = await getTodos();
+
+    setTodos(data);
+  };
+
+  useEffect(() => {
+    fetchData();
+  }, []);
+
+  const showTodo = (todo: Todo) => {
+    setSelectedTodo(todo);
+  };
+
+  const closeTodo = () => {
+    setSelectedTodo(null);
+  };
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+    case 'completed':
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+    case 'all':
+    default:
+      visibleTodos = [...visibleTodos];
+      break;
+  }
```
anna-inozemets, line 39: I guess, wil be better to create a separate function for this part of code

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+
+  const fetchData = async () => {
+    const data = await getTodos();
```
anna-inozemets, line 42: I think. will be better to add handling errors here, if server won't be work properly

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,32 @@
...
+type Props = {
+  closeTodo: () => void,
+  selectedTodo: Todo | null,
+};
+
+export const TodoModal: React.FC<Props> = ({ selectedTodo, closeTodo }) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  const fetchData = async () => {
+    const currentUser = await getUser(selectedTodo?.userId || 0);
```
anna-inozemets, line 16: I guess, we should catch an error here, too)

anna-inozemets (approved):
Great solution!)

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+
+const filteredTodos = (todos:Todo[], query:string, filterBy: string) => {
```
Anna-Tkachenko, line 14: let's use enum for `filterBy` property instead of simple string

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
+  switch (filterBy) {
+    case 'active':
+      preparedTodos = preparedTodos.filter(todo => !todo.completed);
+      break;
+    case 'completed':
+      preparedTodos = preparedTodos.filter(todo => todo.completed);
+      break;
+    case 'all':
+    default:
+      preparedTodos = [...preparedTodos];
```
Anna-Tkachenko, line 32: do you need to make one more copy here?

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
+  return preparedTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [filterBy, setFilterBy] = useState('');
+
+  // let visibleTodos = [...todos];
```
Anna-Tkachenko, line 45: delete redundant and unused comments

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
+  const visibleTodos = filteredTodos(todos, query, filterBy);
+
+  const fetchData = async () => {
+    try {
+      const data = await getTodos();
+
+      setTodos(data);
+    } catch (error) {
+      if (error instanceof Error) {
+        alert(error.message);
```
Anna-Tkachenko, line 74: will be better to handle errors with some message for the user on the page instead of using `alert`

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
+    fetchData();
+  }, []);
+
+  const showTodo = (todo: Todo) => {
+    setSelectedTodo(todo);
+  };
+
+  const closeTodo = () => {
+    setSelectedTodo(null);
+  };
```
Anna-Tkachenko, line 91: don't forget to use advanced hooks to optimize components rendering

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,88 @@
...
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map((todo) => (
+          <tr
+            data-cy="todo"
+            className={cn({
+              'has-background-info-light': selectedTodo?.id === todo.id,
```
Anna-Tkachenko, line 36: you checked is current todo selected several times, let's create separate boolean variable for it

src/types/Todo.ts
```diff
@@ -1,6 +1,12 @@
+// import { User } from './User';
+
 export interface Todo {
   id: number;
   title: string;
   completed: boolean;
   userId: number;
 }
+
+// export interface FullTodo extends Todo {
```
Anna-Tkachenko, line 10: delete redundant comments

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+
+  const visibleTodos = filteredTodos(todos, query, filterBy);
+
+  const fetchData = async () => {
+    try {
+      const data = await getTodos();
+
+      setTodos(data);
+      setIsDataLoading(false);
+      setIsLoadingError(false);
```
Anna-Tkachenko, line 55: don't repeat yourself: you set `false` value to `isDataLoading` field for both cases
can we move it somewhere for calling it only once?

Anna-Tkachenko (approved):
Good job! ⭐️

PR #1369, opened 2023-02-16
solution

Oleksandr-Maly:
[DEMO LINK](https://oleksandr-maly.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,44 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completedStatus, setCompletedStatus] = useState('');
+  const [query, setQuery] = useState('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    const getTodosFromServer = async () => {
+      const todosFromServer = await getTodos();
+
+      setTodos(todosFromServer);
+    };
+
+    getTodosFromServer();
+  }, []);
+
+  let visibleTodos = [...todos];
+
+  if (completedStatus === 'active') {
+    visibleTodos = todos.filter(todo => !todo.completed);
+  }
+
+  if (completedStatus === 'completed') {
+    visibleTodos = todos.filter(todo => todo.completed);
+  }
+
+  if (query.trim()) {
+    visibleTodos = visibleTodos.filter(todo => {
+      const normalizedTitle = todo.title.toLowerCase();
+      const normalizedQuery = query.toLowerCase().trim();
+
+      return normalizedTitle.includes(normalizedQuery);
+    });
+  }
+
   return (
     <>
       <div className="section">
@@ -17,18 +53,31 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                status={completedStatus}
+                setStatus={setCompletedStatus}
+                query={query}
+                setQuery={setQuery}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+
+              {todos.length
+                ? (
+                  <TodoList
+                    todos={visibleTodos}
+                    setSelectedTodo={setSelectedTodo}
+                    selectedTodoId={selectedTodo?.id || 0}
+                  />
+                )
+                : <Loader />}
+
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {selectedTodo && <TodoModal setSelectedTodo={setSelectedTodo} selectedTodo={selectedTodo} />}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,70 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+type Props = {
+  status: string,
+  setStatus: (a:string) => void,
+  query:string;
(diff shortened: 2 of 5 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,84 @@
...
+
+  useEffect(() => {
+    const getUserFromServer = async () => {
+      const UserFromServer = await getUser(userId);
+
+      setUser(UserFromServer);
+    };
+
+    getUserFromServer();
+  }, [selectedTodo]);
```
max-rozzhalovets, line 33: Good, but I'm pretty sure that you don't need `selectedTodo` in dependency array. Modal in this case only mount and unmount and never update, so you don't need any dependencies in your array.
```suggestion
  useEffect(() => {
    const getUserFromServer = async () => {
      const UserFromServer = await getUser(userId);

      setUser(UserFromServer);
    };

    getUserFromServer();
  }, []);
```

max-rozzhalovets (approved):
Very good!

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completedStatus, setCompletedStatus] = useState('');
+  const [query, setQuery] = useState('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
```
denys-cheporniuk, line 17: It's better to save only id instead of object

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+
+  useEffect(() => {
+    const getTodosFromServer = async () => {
+      const todosFromServer = await getTodos();
+
+      setTodos(todosFromServer);
+    };
```
denys-cheporniuk, line 24: move this function outside the useEffect

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+
+  let visibleTodos = [...todos];
+
+  if (completedStatus === 'active') {
+    visibleTodos = todos.filter(todo => !todo.completed);
+  }
+
+  if (completedStatus === 'completed') {
+    visibleTodos = todos.filter(todo => todo.completed);
+  }
```
denys-cheporniuk, line 37: this should be switch case and would be better to create separate function for all this logic

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,70 @@
...
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+  const handleInput = (event:React.ChangeEvent<HTMLInputElement>) => (
+    setQuery(event.target.value)
+  );
+
+  const handleReset = () => {
+    setQuery('');
+    setStatus('all');
```
denys-cheporniuk, line 24: create enum for statuses and use enum in all places

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,102 @@
...
+              </td>
+
+              <td className="has-text-right is-vcentered">
+                {id === selectedTodoId
+                  ? (
+                    <button
+                      data-cy="selectButton"
+                      className="button"
+                      type="button"
+                      onClick={() => handleUnSelectTodo()}
```
denys-cheporniuk, line 78:
```suggestion
                      onClick={handleUnSelectTodo}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,102 @@
...
+                    </button>
+                  )
+                  : (
+                    <button
+                      data-cy="selectButton"
+                      className="button"
+                      type="button"
+                      onClick={() => handleSelectTodo(todo)}
+                    >
+                      <span className="icon"><i className="far fa-eye" /></span>
```
denys-cheporniuk, line 92: don't write tags in one line, every tag should be on the new line

src/App.tsx
```diff
@@ -17,18 +53,31 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <Loader />
-              <TodoList />
+
+              {todos.length
+                ? (
+                  <TodoList
+                    todos={visibleTodos}
+                    setSelectedTodo={setSelectedTodo}
```
denys-cheporniuk, line 70: don't pass function for updating state in child component, create function wrapper for it, fix in all places

src/components/TodoModal/TodoModal.tsx
denys-cheporniuk, line 33: do you need selectedTodo in array here?

denys-cheporniuk (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+import { TodoStatus } from './types/TodoStatus';
+import { FilterTodos } from './types/FilterTodos';
+
+const filterTodos:FilterTodos = (todos, query, completedStatus) => {
+  let prepearedTodos = [...todos];
+
+  if (query.trim()) {
+    prepearedTodos = prepearedTodos.filter(todo => {
+      const normalizedTitle = todo.title.toLowerCase();
+      const normalizedQuery = query.toLowerCase().trim();
```
Anna-Tkachenko, line 21: for now you transform `query` on each callback call
will be better to move it outside of the `filter` method

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [completedStatus, setCompletedStatus] = useState<TodoStatus>(TodoStatus.All);
+  const [query, setQuery] = useState('');
+  const [selectedTodoId, setSelectedTodoId] = useState(0);
+
+  const getTodosFromServer = async () => {
+    const todosFromServer = await getTodos();
```
Anna-Tkachenko, line 48: if there will be any errors while creating api request - your app will be broken, let's fix it

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,104 @@
...
+              data-cy="todo"
+              className={classNames({
+                'has-background-info-light': selectedTodoId,
+              })}
+            >
+              <td className="is-vcentered">{id}</td>
+
+              <td className="is-vcentered">
+                {completed
+                  && (
```
Anna-Tkachenko, line 54:
```suggestion
                {completed && (
```

Anna-Tkachenko (changes requested):

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,103 @@
...
+    id,
+    completed,
+  } = selectedTodo;
+
+  const getUserFromServer = async () => {
+    setIsUserLoadingError(false);
+    setUserIsLoading(true);
+
+    try {
+      const UserFromServer = await getUser(userId);
```
Anna-Tkachenko, line 32:
```suggestion
      const userFromServer = await getUser(userId);
```

Anna-Tkachenko (approved):
Good job!

PR #1378, opened 2023-02-16
react_dynamic-list-of-todos

rafmamedov:
[DEMO LINK](https://rafmamedov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,73 @@
...
+          return !todo.completed;
+
+        default:
+          return true;
+      }
+    });
+
+  const handleGetUser = async (todo: Todo) => {
+    setIsUserFetching(true);
+    const selectedUser = await getUser(todo.userId);
```
slavanosatskyi, line 48: Don't forget to process errors. What if getUser throws an error?

src/App.tsx
```diff
@@ -1,14 +1,73 @@
...
+import { Todo } from './types/Todo';
+import { User } from './types/User';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isTodosFetching, setIsTodosFetching] = useState(true);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [user, setUser] = useState<User | null>(null);
+  const [isUserFetching, setIsUserFetching] = useState(false);
+  const [isModalVisible, setIsModalVisible] = useState(false);
```
slavanosatskyi, line 20: the less state you have the better. isModalVisible can be substituted by selectedTodo. When it is null - modal is hidden, otherwise it is visible

slavanosatskyi (changes requested):

src/App.tsx
rafmamedov: Implemented this, thank you for the review!

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+  const handleShowModal = async (todo: Todo) => {
+    setIsUserFetching(true);
+    setSelectedTodo(todo);
+    try {
+      const selectedUser = await getUser(todo.userId);
+
+      setUser(selectedUser);
+    } catch (error) {
+      // eslint-disable-next-line
+      console.error(error);
```
nazarmatsevych, line 54: You can show a message to user instead of console.log the error

nazarmatsevych (approved):
👍

PR #1382, opened 2023-02-17
Solution

max-rozzhalovets:


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,48 @@
...
+    const fetchUser = async () => {
+      setIsLoading(true);
+      const fetchedUser = await getUser(userId);
+
+      setUser(fetchedUser);
+      setIsLoading(false);
+    };
+
+    fetchUser();
+  }, []);
```
ihor-soloviov, line 39: here the same with try catch

src/App.tsx
```diff
@@ -1,14 +1,42 @@
...
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [typeOfTodos, setTypeOfTodos] = useState('all');
```
denys-cheporniuk, line 14: create enum for typeOfTodos and use it here

src/App.tsx
```diff
@@ -1,14 +1,42 @@
...
+    const fetchTodos = async () => {
+      setIsLoading(true);
+      const todosFromServer = await getTodos();
+
+      setTodos(todosFromServer);
+      setIsLoading(false);
+    };
+
+    fetchTodos();
+  }, []);
```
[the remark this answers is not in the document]
denys-cheporniuk: agree try catch should be here

src/App.tsx
```diff
@@ -1,14 +1,42 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      setIsLoading(true);
+      const todosFromServer = await getTodos();
+
+      setTodos(todosFromServer);
+      setIsLoading(false);
+    };
```
denys-cheporniuk, line 27: move fetchTodos outside the useEffect

src/App.tsx
```diff
@@ -1,14 +1,42 @@
...
+    fetchTodos();
+  }, []);
+
+  const groupedTodos = typeOfTodos === 'all'
+    ? todos
+    : todos.filter(({ completed }) => (typeOfTodos === 'completed' ? completed : !completed));
+
+  const filteredTodos = filter === ''
+    ? groupedTodos
+    : groupedTodos.filter(({ title }) => title.toLowerCase().includes(filter.toLowerCase()));
```
denys-cheporniuk, line 38: create separate function for this logic and use switch case for typeOfTodos
Also, move filter.toLowerCase() to a separate variable, don't do it on each filter iteration

src/App.tsx
```diff
@@ -17,18 +45,30 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {
+                isLoading && <Loader />
+              }
+              <TodoList
+                todos={filteredTodos}
+                onButtonClick={setTodoToShow}
```
denys-cheporniuk, line 62: don't pass function for updating state in child component, create wrapper for it, fix in all places

src/App.tsx
```diff
@@ -17,18 +45,30 @@ export const App: React.FC = () => {
...
+                todoToShow={todoToShow}
+              />
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {
+        todoToShow && <TodoModal todo={todoToShow} onButtonClick={setTodoToShow} />
```
denys-cheporniuk, line 70: don't write code in one line, move TodoModal on the new line of code

src/components/TodoModal/TodoModal.tsx
denys-cheporniuk: agree

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,48 @@
...
+  useEffect(() => {
+    const fetchUser = async () => {
+      setIsLoading(true);
+      const fetchedUser = await getUser(userId);
+
+      setUser(fetchedUser);
+      setIsLoading(false);
+    };
+
+    fetchUser();
```
denys-cheporniuk, line 38: move fetchUser outside the useEffect

denys-cheporniuk (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { TodosType } from './types/TodosType';
 
 export const App: React.FC = () => {
+  const [typeOfTodos, setTypeOfTodos] = useState<string>('all');
```
Anna-Tkachenko, line 15: will be better to use enum instead of simple string for `typeOfTodos` variable

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+
+  const fetchTodos = async () => {
+    try {
+      setIsLoading(true);
+      const todosFromServer = await getTodos();
+
+      setTodos(todosFromServer);
+    } catch {
+      // eslint-disable-next-line
+      console.log('Error occurred in App fetch')
```
Anna-Tkachenko, line 45: will be better to show some message on the page about the error

src/App.tsx
```diff
@@ -17,18 +79,37 @@ export const App: React.FC = () => {
...
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {
+                isLoading && <Loader />
+              }
+              {
+                !isLoading && (
```
Anna-Tkachenko, line 96: ternary operator usage will be better here

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,59 @@ export const TodoList: React.FC = () => (
...
+                  data-cy="selectButton"
+                  className="button"
+                  type="button"
+                  onClick={() => onCloseButton(todo)}
+                >
+                  <span className="icon">
+                    <i
+                      className={
+                        classNames({
+                          'far fa-eye': todoToShow?.id !== id,
```
Anna-Tkachenko, line 72: you checked it several times, will be better to create separate boolean variable

Anna-Tkachenko (approved):

PR #1383, opened 2023-02-17
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable max-len */
 import React, { useEffect, useState, useCallback } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';


src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filterBy, setFilterBy] = useState('all');
```
denys-cheporniuk, line 16: create enum for filterBy and use enum here

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+  const [query, setQuery] = useState('');
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const data = await getTodos();
+
+      setTodos(data);
+    };
```
denys-cheporniuk, line 24: move fetchData outside the useEffect

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+        return [];
+      case 'active':
+        return todo.completed === false;
+      case 'completed':
+        return todo.completed === true;
+    }
+  });
+
+  if (query) {
+    visibleTodos = [...visibleTodos].filter(todo => todo.title.toLocaleLowerCase().includes(query.toLocaleLowerCase()));
```
denys-cheporniuk, line 58: create a variable for query.toLocaleLowerCase(), don't do it on each filter iteration

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,79 @@
...
-          <p className="has-text-danger">delectus aut autem</p>
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
+export const TodoList: React.FC<Props> = React.memo(({
+  todos,
+  openModal,
+  selectedTodo,
+}) => {
+  return (
```
denys-cheporniuk, line 16: you can avoid return here

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,35 @@
...
+  openModal,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const data = await getUser(selectedTodo.userId);
+
+      setUser(data);
+    };
```
denys-cheporniuk, line 23: move fetchData outside the useEffect

denys-cheporniuk (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,77 @@
...
+
+  const changeQuery = useCallback((value: string) => {
+    setQuery(value);
+  }, []);
+
+  const resetQuery = useCallback(() => {
+    setQuery('');
+  }, []);
+
+  let visibleTodos = todos.filter(todo => {
```
Anna-Tkachenko, line 57: let's move todos preparing to the separate function

src/App.tsx
```diff
@@ -17,18 +80,34 @@ export const App: React.FC = () => {
...
+                  />
+                )}
+
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo && <TodoModal selectedTodo={selectedTodo} openModal={openModal} />}
```
Anna-Tkachenko, line 109:
```suggestion
      {selectedTodo && (
        <TodoModal
          selectedTodo={selectedTodo}
          openModal={openModal}
        />
      )}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,50 @@ export const TodoList: React.FC = () => (
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          data-cy="todo"
+          className={cn({
+            'has-background-info-light': selectedTodo,
```
Anna-Tkachenko, line 35: check the condition that you used for setting this class
do we need to set this classname for each todo if `selectedTodo` is not null?

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,50 @@ export const TodoList: React.FC = () => (
...
+          <td className="has-text-right is-vcentered">
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => openModal(todo)}
+            >
+              <span className="icon">
+                <i className={cn('far', {
+                  'fa-eye': !selectedTodo,
```
Anna-Tkachenko, line 65: you should check is current todo selected or no, not just `selectedTodo` variable

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,35 @@
...
+}
+
+export const TodoModal: React.FC<Props> = React.memo(({
+  selectedTodo,
+  openModal,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  const fetchData = async () => {
+    const data = await getUser(selectedTodo.userId);
```
Anna-Tkachenko, line 19: if there will be any errors on server - your app will be broken
let's fix it

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+    fetchData();
+  }, []);
+
+  const openModal = useCallback((todo: Todo | null) => {
+    setSelectedTodo(todo);
+  }, []);
+
+  const filterTodos = useCallback((selection: string) => {
+    switch (selection) {
+      case 'all':
```
Anna-Tkachenko, line 36: will be better to use enum here

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+
+  const changeQuery = useCallback((value: string) => {
+    setQuery(value);
+  }, []);
+
+  const resetQuery = useCallback(() => {
+    setQuery('');
+  }, []);
+
+  const visibleTodos = preparingTodos(todos, filterBy, query);
```
Anna-Tkachenko, line 57:
```suggestion
  const visibleTodos = getPreparedTodos(todos, filterBy, query);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,50 @@ export const TodoList: React.FC = () => (
...
+          <td className="has-text-right is-vcentered">
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => openModal(todo)}
+            >
+              <span className="icon">
+                <i className={cn('far', {
+                  'fa-eye': selectedTodo?.id !== todo.id,
```
Anna-Tkachenko, line 65: you checked is current todo selected several times, let's create boolean variable for this

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  const fetchData = async () => {
+    try {
+      const data = await getUser(selectedTodo.userId);
+
+      setUser(data);
+    } catch {
+      throw new Error('There is an error with loading');
```
Anna-Tkachenko, line 24: in this way your app will be broken
you shouldn't throw an error here, will be better to manage some error message with boolean state variable

src/utils.tsx
```diff
@@ -0,0 +1,30 @@
...
+        return todo.completed === false;
+      case FilterBy.Completed:
+        return todo.completed === true;
+    }
+  });
+
+  const queryToLowerCase = query.toLocaleLowerCase();
+
+  if (query) {
+    preparedTodos = [...preparedTodos]
```
Anna-Tkachenko, line 24: do you need to make a copy here? filter method returns the new array

Anna-Tkachenko (approved):

PR #1385, opened 2023-02-18
solution

victorionov:
- [DEMO LINK](https://victorionov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,74 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
+// import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { TodoModal } from './components/TodoModal';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodoId, setSelectedTodoId] = useState<number | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [selectedStatus, setSelectedStatus] = useState('all');
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [hasErrorFromServer, setHasErrorFromServer] = useState(false);
+
+  const selectedTodo = useMemo((): Todo | undefined => {
+    return todos.find(todo => todo.id === selectedTodoId);
+  }, [todos, selectedTodoId]);
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      setHasErrorFromServer(false);
+      try {
+        const todosFromServer = await getTodos();
+
+        setTodos(todosFromServer);
+        setVisibleTodos(todosFromServer);
+        setIsLoading(false);
+      } catch {
+        setHasErrorFromServer(true);
+        setIsLoading(false);
+      }
+    };
+
+    fetchTodos();
+  }, []);
+
+  useEffect(() => {
+    const filterTodos = () => {
+      const queryToLowerCase = searchQuery.toLowerCase();
+
+      let todosToShow = todos.filter(todo => {
+        switch (selectedStatus) {
+          case 'all':
+            return true;
+          case 'active':
+            return !todo.completed;
+          case 'completed':
+            return todo.completed;
+          default:
+            return true;
+        }
+      });
+
+      if (searchQuery) {
+        todosToShow = todosToShow.filter(todo => todo.title.toLowerCase().includes(queryToLowerCase));
+      }
+
+      setVisibleTodos(todosToShow);
+    };
+
+    filterTodos();
+  }, [selectedStatus, searchQuery]);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
-import { TodoModal } from './components/TodoModal';
+// import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { TodoModal } from './components/TodoModal';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodoId, setSelectedTodoId] = useState<number | null>(null);
```
denys-cheporniuk, line 16: null can't be here, use 0 instead of null

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [selectedStatus, setSelectedStatus] = useState('all');
```
denys-cheporniuk, line 19: create enum for selectedStatus and use it here

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+        const todosFromServer = await getTodos();
+
+        setTodos(todosFromServer);
+        setVisibleTodos(todosFromServer);
+        setIsLoading(false);
+      } catch {
+        setHasErrorFromServer(true);
+        setIsLoading(false);
+      }
+    };
```
denys-cheporniuk, line 40: move fetchTodos outside the useEffect

src/App.tsx
```diff
@@ -17,18 +77,35 @@ export const App: React.FC = () => {
...
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              {hasErrorFromServer
+                ? <p>Error while fetching</p>
+                : (
+                  <TodoList
+                    todos={visibleTodos}
+                    setSelectedTodoId={setSelectedTodoId}
```
denys-cheporniuk, line 95: don't pass function for updating state in child component, create wrapper for it, fix in all places

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,69 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={selectedStatus}
+            onChange={handleSelectedStatus}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
denys-cheporniuk, line 37: here should be enum

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +23,39 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr key={todo.id} data-cy="todo" className="">
```
denys-cheporniuk, line 27: remove className if you don't have class for this element

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,50 @@
...
+  setSelectedTodoId,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    const fetchUser = async () => {
+      const userFromServer = await getUser(selectedTodoId);
+
+      setUser(userFromServer);
+    };
```
denys-cheporniuk, line 26: move fetchUser outside the useEffect

denys-cheporniuk (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,73 @@
...
+      setTodos(todosFromServer);
+      setVisibleTodos(todosFromServer);
+      setIsLoading(false);
+    } catch {
+      setHasErrorFromServer(true);
+      setIsLoading(false);
+    }
+  };
+
+  fetchTodos();
```
Anna-Tkachenko, line 42: check the place, where you need to call `fetchTodos` function
looks like with this code it can cause infinite call cycle, because this function will be called for each component rerendering

src/App.tsx
```diff
@@ -1,14 +1,73 @@
...
+
+  useEffect(() => {
+    const filterTodos = () => {
```
Anna-Tkachenko, line 45: let's move the function for getting prepared todos outside of the component

src/App.tsx
```diff
@@ -17,18 +76,35 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                searchQuery={searchQuery}
+                setSearchQuery={() => setSearchQuery}
```
Anna-Tkachenko, line 81: you can't pass functions for changing the state directly to the child component
let's create separate functions before the `return` statement, where we will accept the arguments and will pass them to the functions for changing the state

and will pass those wrappers to the child components

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,47 @@
...
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { Loader } from '../Loader';
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+import { getUser } from '../../api';
+
+interface TodoModalProps {
+  selectedTodo: Todo | undefined,
+  selectedTodoId: number | 0,
```
Anna-Tkachenko, line 10: 0 is not a number?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,47 @@
...
+}
+
+export const TodoModal: React.FC<TodoModalProps> = ({
+  selectedTodo,
+  selectedTodoId,
+  setSelectedTodoId,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const fetchUser = async () => {
+    const userFromServer = await getUser(selectedTodoId);
```
Anna-Tkachenko, line 21: if there will be any error while calling `getUser` - your app will be broken, let's fix it

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,47 @@
...
+
+    setUser(userFromServer);
+  };
+
+  fetchUser();
```
Anna-Tkachenko, line 26: same here about infinite cycle
check how it works
for now you call the function for fetching user, this function updates the state, `TodoModal` component renders one more time, and this function will be called again

Anna-Tkachenko (changes requested):

denys-cheporniuk (approved):

PR #1395, opened 2023-02-20
Develop

U-Ra-All:
[DEMO LINK](https://U-Ra-All.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,7 +13,7 @@ import { Todo } from './types/Todo';
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [isLoading, setIsLoading] = useState(true);
-  const [selectedTodoId, setSelectedTodoId] = useState(0);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
 
   useEffect(() => {
     getTodos().then((receivedTodos) => {
@@ -24,12 +24,12 @@ export const App: React.FC = () => {
     });
   }, []);
 
-  const showTodoDetails = (todoId: number) => () => {
-    setSelectedTodoId(todoId);
+  const showTodoDetails = (todo: Todo | null) => () => {
+    setSelectedTodo(todo);
   };
 
   const hideTodoDetails = () => {
-    setSelectedTodoId(0);
+    setSelectedTodo(null);
   };
 
   return (
@@ -47,7 +47,7 @@ export const App: React.FC = () => {
               {isLoading && <Loader />}
               <TodoList
                 todos={todos}
-                selectedTodoId={selectedTodoId}
+                selectedTodo={selectedTodo}
                 onShowTodo={showTodoDetails}
               />
             </div>
@@ -55,12 +55,8 @@ export const App: React.FC = () => {
         </div>
       </div>
 
-      {selectedTodoId !== 0 && (
-        <TodoModal
-          todos={todos}
-          selectedTodoId={selectedTodoId}
-          onHideTodo={hideTodoDetails}
-        />
+      {selectedTodo !== null && (
+        <TodoModal selectedTodo={selectedTodo} onHideTodo={hideTodoDetails} />
       )}
     </>
   );
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,8 +11,8 @@ import { getTodos } from './api';
 import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  const [todos, setTodos] = useState<Todo[]>([]);
   const [isLoading, setIsLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
   const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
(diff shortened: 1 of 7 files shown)


U-Ra-All:
Can't understand why 1 test doesn't pass
Screenshot 2023-02-20 at 15 26 05

src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
+
+          switch (todoSelector) {
+            case 'active':
+              return !todo.completed && todoTitleIncludesFilter;
+            case 'completed':
+              return todo.completed && todoTitleIncludesFilter;
+            default:
+              return todoTitleIncludesFilter;
+          }
+        });
```
IvanFesenko, line 36: Move this to visibleTodos variable using useMemo.
Set to state all todos from server

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,62 @@
...
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          {todoFilter !== '' && (
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={onChangeTodoFilter}
```
IvanFesenko, line 54: Better crate clear handler where you will set empty query
Refactor `handleTodoFiltering ` after changes
```
const onClearQuery = () => {
  onChangeTodoFilter('')
}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,22 +1,82 @@
...
+                  <span className="icon">
+                    <i className="far fa-eye" />
+                  </span>
+                </button>
+              )}
+            </td>
+          </tr>
+        ))}
+        {/* <tr data-cy="todo" className="">
         <td className="is-vcentered">1</td>
```
IvanFesenko, line 80: Remove comments

IvanFesenko (changes requested):
Good job. Please check my comments

U-Ra-All:
> Good job. Please check my comments

Thanks, done, please take a look

src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [todoSelector, setTodoSelector] = useState('all');
+  const [todoFilter, setTodoFilter] = useState('');
+
+  useEffect(() => {
+    getTodos().then((data) => {
+      setTodos(data);
+      setIsLoading(false);
+    });
```
OlEzhA0, line 24: nb: would be better to use `async/await`

src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
+          return todoTitleIncludesFilter;
+      }
+    });
+  };
+
+  const visibleTodos = useMemo(getVisibleTodos, [
+    todos,
+    todoSelector,
+    todoFilter,
+  ]);
```
OlEzhA0, line 48:
```suggestion
  const visibleTodos = useMemo(() => {
    return todos.filter((todo) => {
      const todoTitleIncludesFilter = todo.title
        .toLowerCase()
        .includes(todoFilter.toLowerCase());

      switch (todoSelector) {
        case 'active':
          return !todo.completed && todoTitleIncludesFilter;
        case 'completed':
          return todo.completed && todoTitleIncludesFilter;
        default:
          return todoTitleIncludesFilter;
      }
    });
  }, [
    todos,
    todoSelector,
    todoFilter,
  ]);
```

OlEzhA0 (approved):
hey! GJ

src/App.tsx
IvanFesenko, line 48: This wont work. We need useMemo to avoid unnesessary loop iteration. Also if selected all and query is empty - no need to map all todos. You should do like this.
```suggestion
  const visibleTodos = useMemo(() => {
    const needsToFilter = todoSelector === 'active' || todoSelector === 'completed';
  
    if (!needsToFilter || !todoFilter) {
      return todos;
    }
    
    return todos.filter((todo) => {
      const todoTitleIncludesFilter = todo.title
        .toLowerCase()
        .includes(todoFilter.toLowerCase());
        
        switch (todoSelector) {
        case 'active':
          return !todo.completed && todoTitleIncludesFilter;
        case 'completed':
          return todo.completed && todoTitleIncludesFilter;
        default:
          return todoTitleIncludesFilter;
      }     
    });
 
  }, [todos, todoSelector, todoFilter,]);

```

IvanFesenko (changes requested):
Fix useMemo usage

src/App.tsx
U-Ra-All: > hey! GJ

Hey! Thanks, done!

src/App.tsx
U-Ra-All: > This wont work. We need useMemo to avoid unnesessary loop iteration. Also if selected all and query is empty - no need to map all todos. You should do like this.

Requested changes are added

PR #1396, opened 2023-02-21
Initial commit

yaroslav-furhala:
- [DEMO LINK](https://yaroslav-furhala.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
 export const App: React.FC = () => {
+  const [isDataLoaded, setIsDataLoaded] = useState(false);
+  const [initialTodos, setInitialTodos] = useState<Todo[]>([]);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState<Filter>(Filter.All);
+  const [searchQuery, setSearchQuery] = useState('');
+
+  useEffect(() => {
+    getTodos().then(todosFromServer => {
```
vlad-demchuk, line 23: use `catch` to avoid any fetching errors

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+
+      case Filter.Completed:
+        setTodos(initialTodos.filter(todo => todo.completed));
+        break;
+
+      default:
+        break;
+    }
+
+    setTodos(
```
vlad-demchuk, line 48: it would be bettter to create todos copy, filter it, and only then invoke `setTodos` with correct data

vlad-demchuk (changes requested):
try to fix tests

PR #1401, opened 2023-02-23
Solution

OksanaBaloh:


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,93 @@
...
+  allTodos,
+  onSetTodos,
+}) => {
+  const [completed, setCompleted] = useState('all');
+  const [query, setQuery] = useState('');
+
+  const filterCallbackByCompleted
+  = (todoStatus: boolean, selectedStatus: string) => {
+    switch (selectedStatus) {
+      case 'active':
```
loralevitska, line 19: [Try not to hardcode these values, but use ](https://www.typescriptlang.org/docs/handbook/enums.html)[Enum](https://www.typescriptlang.org/docs/handbook/enums.html) instead

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,93 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
loralevitska, line 81: Try to fix such eslint rules and remove comments

loralevitska (changes requested):
Please deploy your solution, and before review check demo_link (to description in PR). If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/App.tsx
```diff
@@ -17,18 +29,34 @@ export const App: React.FC = () => {
...
+              <TodoFilter
+                allTodos={allTodos}
+                onSetTodos={(todos: Todo[]) => setVisibleTodos(todos)}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {allTodos.length === 0 ? (
```
etojeDenys, line 39:
```suggestion
              {!allTodos.length ? (
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,95 @@
...
+type Props = {
+  allTodos: Todo[];
+  onSetTodos: (allTodos: Todo[]) => void;
+};
+
+export const TodoFilter: React.FC<Props> = ({
+  allTodos,
+  onSetTodos,
+}) => {
+  const [completed, setCompleted] = useState('all');
```
etojeDenys, line 14:
```suggestion
  const [filterType, setFilterType] = useState(Status.all);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,82 @@
...
-          <p className="has-text-danger">delectus aut autem</p>
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
+export const TodoList: React.FC<Props> = ({
+  todos,
+  selectedTodoId,
+  selectTodo,
+}) => {
+  return (
```
etojeDenys, line 16: can omit
```suggestion
}) => (
```

etojeDenys (changes requested):

src/App.tsx
OleziO: Not resolved

OleziO: Check other similar cases.

src/components/TodoFilter/TodoFilter.tsx
OleziO: Not resolved

src/components/TodoList/TodoList.tsx
OleziO: Not resolved

OleziO (changes requested):

OksanaBaloh:
I still have problem with tests. How can I fix that?

OleziO (approved):

PR #1406, opened 2023-02-28
initial

kostasburnazaki:
[DEMO LINK](https://kostasburnazaki.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,23 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { initValues } from './helpers/constants';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>(initValues.initTodos);
+  const [selectedTodoId, setTodoId] = useState<number>(initValues.initSelectedTodoId);
+  const [filter, setFilter] = useState<string>(initValues.initFilter);
+  const [search, setSearch] = useState<string>(initValues.initSearch);
+
+  useEffect(() => {
+    getTodos()
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+      });
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,29 @@
...
+  const [todos, setTodos] = useState<Todo[]>(initValues.initTodos);
+  const [selectedTodoId, setTodoId] = useState<number>(initValues.initSelectedTodoId);
+  const [filter, setFilter] = useState<string>(initValues.initFilter);
+  const [search, setSearch] = useState<string>(initValues.initSearch);
+
+  useEffect(() => {
+    getTodos()
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+      });
```
DarkMistyRoom, line 24: what if we will get an error from the server?

src/App.tsx
```diff
@@ -17,18 +32,35 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                setFilter={(value: string) => setFilter(value)}
```
DarkMistyRoom, line 36:
```suggestion
                setFilter={setFilter}
```

DarkMistyRoom: use short syntax

DarkMistyRoom: check it everywhere

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,87 @@
...
+      userId,
+    } = foundTodo;
+
+    const color = completed ? 'success' : 'danger';
+
+    useEffect(() => {
+      getUser(userId)
+        .then(userFromServer => {
+          setUser(userFromServer);
+        });
```
DarkMistyRoom, line 35: let's handle errors

DarkMistyRoom (changes requested):
Good, just handle errors

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,111 @@
...
+        case Filter.all:
+          return todo;
+        case Filter.completed:
+          return todo.completed;
+        case Filter.active:
+          return !todo.completed;
+        default:
+          throw new Error('Error during filtering');
+      }
+    })
```
Volodymyr-Melnyk-Fron10der, line 32: do different filters inside every case. because for all filter it would be extra loop

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
-          className="delete"
+export const TodoFilter: React.FC<Props> = ({
+  setFilter,
+  filter,
+  setSearch,
+  searchQuery,
+}) => {
+  return (
+    <form className="field has-addons">
+      <p className="control">
```
SerhiiKirik, line 18: why here tag "p"?
```suggestion
      <div className="control">
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
+        <span className="select">
```
SerhiiKirik, line 19: span?
```suggestion
        <span className="select">
```

SerhiiKirik (approved):
Good job 🚀

PR #1408, opened 2023-03-03
solved

Deniso4ka23:
[DEMO LINK](https://Deniso4ka23.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
+/* eslint-disable no-console */
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+
+  const handleSelelect = (newSelectFilterType: string) => {
+    setFilterType(newSelectFilterType);
+  };
+
+  const handleQuery = (newQuery: string) => {
+    setQuery(newQuery);
+  };
+
+  const visibleTodoList = getFilteredList(filterType)
```
agniya-i, line 53: it would be great to use `useMemo` here

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,24 @@
...
   <form className="field has-addons">
     <p className="control">
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
+          data-cy="statusSelect"
+          value={filterType}
+          onChange={(e) => onFilterType(e.target.value)}
+        >
           <option value="all">All</option>
```
agniya-i, line 22: you can create enum for options

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [selectedTodo, setSlectedTodo] = useState<Todo | null>();
+  const [filterType, setFilterType] = useState('all');
+  const [query, setQuery] = useState('');
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const newTodoList = await getTodos();
+
+      setTodoList(newTodoList);
```
agniya-i, line 24: wrap request into try-catch block,
what happens if there is an error?

agniya-i (changes requested):
please add all files to PR  👀

src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+  const handleSelelect = (newSelectFilterType: string) => {
+    setFilterType(newSelectFilterType);
+  };
+
+  const handleQuery = (newQuery: string) => {
+    setQuery(newQuery);
+  };
+
+  const visibleTodoList = useMemo(() => {
+    return getFilteredList(filterType)
```
witflash, line 62: You can use implicit return here

src/App.tsx
```diff
@@ -17,18 +71,35 @@ export const App: React.FC = () => {
...
+                onQuery={handleQuery}
+                query={query}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todoList.length === 0 ? (
+                hasLoadingError ? 'Error: Data has not been loaded' : <Loader />
```
witflash, line 84: It's bad practice to use ternary operator inside another one.
For the learning it's ok but for the future it's better to think about more pure solution.

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,46 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr key={todo.id} data-cy="todo" className="">
```
witflash, line 32: `className` is redundant

witflash (approved):
Well done 👍🏻 
I left some uncritical comments for your future work

PR #22, opened 2019-06-13
todos

RMasnyuk:
https://rmasnyuk.github.io/react_dynamic-list-of-todos/

PR #1409, opened 2023-03-05
finished

barantarasNEW:
https://barantarasNEW.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


barantarasNEW:
Не розумію чому на user #22 помилка відбувається

src/App.tsx
```diff
@@ -8,7 +8,68 @@ import { TodoFilter } from './components/TodoFilter';
...
+  ACTIVE = 'active',
+  COMPLETED = 'completed',
+}
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setLoading] = useState(true);
+  const [isModal, setModal] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState<Filter | string>(Filter.ALL);
```
etojeDenys, line 25: why do you need to set here `string`
```suggestion
  const [filter, setFilter] = useState<Filter>(Filter.ALL);
```

src/App.tsx
```diff
@@ -8,7 +8,68 @@ import { TodoFilter } from './components/TodoFilter';
...
+      case Filter.ACTIVE:
+        setFilteredTodos(todos.filter(todo => !todo.completed));
+
+        break;
+      case Filter.COMPLETED:
+        setFilteredTodos(todos.filter(todo => todo.completed));
+
+        break;
+      default:
+        setFilteredTodos(todos);
```
etojeDenys, line 48: it's better to add `case Filter.ALL`, it will improve readability and help us when we want to scale our app

src/App.tsx
```diff
@@ -8,7 +8,68 @@ import { TodoFilter } from './components/TodoFilter';
...
+  const showModal = (todo: Todo) => {
+    setModal(true);
+    setSelectedTodo(todo);
+  };
+
+  const closeModal = () => {
+    setModal(false);
+  };
+
+  const changeFilter = (value: string) => setFilter(value);
```
etojeDenys, line 61:
```suggestion
  const changeFilter = (value: Filter) => setFilter(value);
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  const [isLoading, setLoading] = useState(true);
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    getUser(selectedTodo?.id || 0)
+      .then(res => {
+        setUser(res);
+      }).catch((reg) => {
+        throw new Error(reg);
+      }).finally(() => setLoading(false));
```
etojeDenys, line 24: i think we can omit this
```suggestion
      .then(res => setUser(res))
      .catch((reg) => throw new Error(reg))
      .finally(() => setLoading(false));
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,28 @@
...
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
+          data-cy="statusSelect"
+          value={filter}
+          onChange={(e) => changeFilter(e.target.value)}
+        >
           <option value="all">All</option>
           <option value="active">Active</option>
           <option value="completed">Completed</option>
```
etojeDenys, line 28: we can use here Filter enum values

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,28 @@
-export const TodoFilter = () => (
+import React from 'react';
+
+type Props = {
+  filter: string;
+  changeFilter: (value: string) => void;
```
```suggestion
  changeFilter: (value: Filter) => void;
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,19 +36,24 @@ export const TodoFilter = () => (
...
-      </span>
+      {searchQuery.length > 0 && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="clearSearchButton"
+            type="button"
+            className="delete"
+            onClick={clearSearchQuery}
+          />
```
etojeDenys, line 54: just rewrite it with `<button ...></button>` instead of `<button ... />`

etojeDenys (changes requested):
good job, just make small improvements

IvanFesenko (approved):
Well done

PR #1410, opened 2023-03-06
Dynamic list of todos

Mykola-Hadupiak:
[TODOS](https://Mykola-Hadupiak.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,27 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import { Todo } from './types/Todo';
+import { Filter } from './types/Filter';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodoId, setTodoId] = useState<number | null>(null);
+  const [filter, setFilter] = useState<string>(Filter.all);
+  const [search, setSearch] = useState<string>('');
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos);
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +30,35 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                filter={filter}
+                setFilter={setFilter}
+                searchQuery={search}
+                setSearch={setSearch}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!todos.length && <Loader />}
+
+              <TodoList
+                todos={todos}
+                selectedTodo={selectedTodoId}
+                selectTodo={setTodoId}
+                filter={filter}
+                search={search}
+              />
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {selectedTodoId && (
+        <TodoModal
+          todos={todos}
+          selectedTodo={selectedTodoId}
+          unselectTodo={() => setTodoId(0)}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,100 +1,112 @@
 import React from 'react';
+import { Todo } from '../../types/Todo';
+import { Filter } from '../../types/Filter';
 
-export const TodoList: React.FC = () => (
-  <table className="table is-narrow is-fullwidth">
-    <thead>
-      <tr>
-        <th>#</th>
-        <th>
-          <span className="icon">
-            <i className="fas fa-check" />
-          </span>
-        </th>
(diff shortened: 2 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,27 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodoId, setTodoId] = useState<number | null>(null);
+  const [filter, setFilter] = useState<string>(Filter.all);
```
Volodymyr-Melnyk-Fron10der, line 17: leave default type (Filter)
```suggestion
  const [filter, setFilter] = useState(Filter.all);
```

src/App.tsx
```diff
@@ -1,14 +1,27 @@
...
+  const [search, setSearch] = useState<string>('');
```
Volodymyr-Melnyk-Fron10der, line 18: default type is string already
```suggestion
  const [search, setSearch] = useState('');
```

src/App.tsx
```diff
@@ -17,18 +30,35 @@ export const App: React.FC = () => {
...
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {selectedTodoId && (
+        <TodoModal
+          todos={todos}
+          selectedTodo={selectedTodoId}
+          unselectTodo={() => setTodoId(0)}
```
Volodymyr-Melnyk-Fron10der, line 59: as I see you use null value by deafault 
```suggestion
          unselectTodo={() => setTodoId(null)}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,112 @@
...
+        case Filter.all:
+          return todo;
+        case Filter.completed:
+          return todo.completed;
+        case Filter.active:
+          return !todo.completed;
+        default:
+          throw new Error('Something went wrong with the filtering');
+      }
+    })
```
Volodymyr-Melnyk-Fron10der, line 32: //[HINT] if you need to filter your items by selected filter - use switch/case statement in correct way.
// Perform the necessary filtering only on a case-by-case basis.

// GOOD

const filteredCards = useMemo(() => {
  switch (filter) {
    case CardFilters.All: {
      return cards;
    }

    case CardFilters.Active: {
      return cards.filter(({ deletedDate }) => !deletedDate);
    }

    case CardFilters.Deleted: {
      return cards.filter(({ deletedDate }) => Boolean(deletedDate));
    }

    default:
      return [];
  }
}, [cards, filter]);

// Don't filter your array right away with switch/case inside your callback.
// Because in case CardFilters.All your program just iterate an array and doesn't do any useful actions. So it's a wasting of memory and CPU (very bad)

// BAD

const filteredCards = useMemo(() => {
  return cards.filter(({ deletedDate }) => {
    switch (filter) {
      case CardFilters.All: {
        return true;
      }

      case CardFilters.Active: {
        return !deletedDate;
      }

      case CardFilters.Deleted: {
        return Boolean(deletedDate);
      }

      default:
        return false;
    }
  });
}, [cards, filter]);

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,89 @@
...
-        </div>
-      )}
-    </div>
-  );
+        ) : <Loader />}
+      </div>
+    );
+  }
+
+  throw new Error('No todo found');
```
Volodymyr-Melnyk-Fron10der, line 88: don't break react pls) just return smth like

`return <>No todo found</>;`

src/types/Filter.ts
```diff
@@ -0,0 +1,5 @@
+export enum Filter {
+  all = 'all',
```
Volodymyr-Melnyk-Fron10der, line 2: keys in enum should be uppercased (best practice)
```suggestion
  All = 'all',
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,89 @@
...
+
+export const TodoModal: React.FC<Props> = ({
+  todos,
+  selectedTodo,
+  unselectTodo,
+}) => {
+  const [user, setUser] = useState<User>();
+  const foundTodo = todos.find(todo => todo.id === selectedTodo);
+
+  if (foundTodo) {
```
Volodymyr-Melnyk-Fron10der, line 21: I would handle !foundTodo first in this case. It would looks more clear

Volodymyr-Melnyk-Fron10der (changes requested):
almost there👍

src/App.tsx
Mykola-Hadupiak: if i do that i get an error

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,119 @@
...
+
+      <tbody>
+        {todosForRender.map(({ id, title, completed }: Todo) => {
+          const color = completed ? 'success' : 'danger';
+
+          return (
+            <tr
+              key={id}
+              data-cy="todo"
+              className=""
```
SanyaBratashchuk, line 67: Why do you need empty className?

SanyaBratashchuk (approved):

src/components/TodoList/TodoList.tsx
Mykola-Hadupiak: It was there 😅😅 I didn't create it

etojeDenys (approved):
good job

PR #1430, opened 2023-03-22
solution

MargaritaChorniak:
[DEMO LINK](https://MargaritaChorniak.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,60 @@
 /* eslint-disable max-len */
-import React from 'react';
+
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import React, { useCallback, useEffect, useState } from 'react';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+      .catch(error => {
+        throw new Error(error);
+      });
+  }, []);
+
+  const chooseUser = useCallback((todo: Todo | null) => {
+    setSelectedTodo(todo);
+  }, []);
+
+  const filterTodos = () => {
```
TarasHoliuk, line 39:
```suggestion
  const filterTodos = (todos: Todo[], status: TodosStatus) => {
```
I see you use `useCallback`, so a few words about optimization (just FYI):
- let the function accept 2 arguments :arrow_up: 
- move it out of component
- create in component `const visibleTodos = useMemo(() => filterTodos(todos, selectedTodosStatus), [todos, selectedTodosStatus])`

=> visible todos will be recalculated only on `todos` or  `selectedTodosStatus` change

TarasHoliuk (approved):
GJ :+1: 
I left 1 comment about optimization (check it please) - just FYI

PR #1436, opened 2023-03-24
dynamic list of todos completed

MykolaKaradzha:
[DEMO LINK](https://MykolaKaradzha.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,10 @@
 /* eslint-disable max-len */
-import React, { useEffect, useState } from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -20,15 +25,25 @@ export const App: React.FC = () => {
     getTodos().then(data => setTodos(data));
   }, []);
 
-  const showTodo = (todo: Todo) => {
+  const showTodo = useCallback((todo: Todo) => {
     setSelectedTodo(todo);
+  }, []);
+
+  const getCurrentTodo = () => {
+    if (!selectedTodo) {
+      return todos[0];
+    }
+
+    return todos.filter(todo => todo.id === selectedTodo.id)[0];
   };
 
   const closeModal = () => {
     setSelectedTodo(null);
   };
 
-  const getFilter = (filter: string) => setSelectedFilter(filter);
+  const getFilter = useCallback(
+    (filter: string) => setSelectedFilter(filter), [],
+  );
 
   const filterTodos = (option: string, searchValue?: string) => {
     let filteredTodos = [...todos];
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -17,18 +79,31 @@ export const App: React.FC = () => {
...
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal
+          closeModal={closeModal}
+          currentTodo={getCurrentTodo()}
```
VitaliyBondarenko1982, line 104: do you really need this function?
suggest pass currentTodo.
and change type of currentTodo prop in TodoModal

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+    setSelectedTodo(todo);
+  }, []);
+
+  const getCurrentTodo = () => {
+    if (!selectedTodo) {
+      return todos[0];
+    }
+
+    return todos.filter(todo => todo.id === selectedTodo.id)[0];
+  };
```
VitaliyBondarenko1982, line 38:
```suggestion
```
looks like redundant. enough to have selectedTodo state

src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+  const [selectedFilter, setSelectedFilter] = useState('all');
+  const [searchBarValue, setSearchBarValue] = useState('');
+
+  useEffect(() => {
+    getTodos().then(data => setTodos(data));
+  }, []);
+
+  const showTodo = useCallback((todo: Todo) => {
+    setSelectedTodo(todo);
+  }, []);
```
etojeDenys, line 30: do we really need to cover this function in useCallback. I think we can remove it

src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+      case 'active':
+        return filteredTodos.filter(todo => !todo.completed);
+
+      default:
+        return filteredTodos;
+    }
+  };
+
+  const filteredTodos = useMemo(() => (
+    filterTodos(selectedFilter, searchBarValue)), [todos, selectedFilter, searchBarValue]);
```
etojeDenys, line 60: you forgot to add filterTodos to the dependencies array

src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+        return filteredTodos.filter(todo => todo.completed);
+
+      case 'active':
+        return filteredTodos.filter(todo => !todo.completed);
+
+      default:
+        return filteredTodos;
+    }
+  };
+
```
etojeDenys, line 58: let's cover this big function in useCallback

etojeDenys (changes requested):
almost done

src/App.tsx
MykolaKaradzha: Well, in our case, Todolist component is rerendred all time. But hypotheticaly, we can have menu component for example, which will rerender app, but will not effect todolist. That is why i wrapped this handler, so Todolist won`t be rerendred just because its parent component is rerendered.

src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
+    getTodos().then(data => setTodos(data));
+  }, []);
+
+  const showTodo = (todo: Todo) => {
+    setSelectedTodo(todo);
+  };
+
+  const closeModal = () => {
+    setSelectedTodo(null);
+  };
```
VitaliyBondarenko1982, line 34: this function also looks like redundant
we can pass setSelectedTodo where we need
just suggest to think about it)

VitaliyBondarenko1982 (approved):

PR #1444, opened 2023-03-31
solution

vladyslav-krainyk:
[DEMO LINK](https://vladyslav-krainyk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,83 @@
...
+
+        case FilterType.Completed:
+          return completed && lowerCaseTitle.includes(lowerCaseQuery);
+
+        case FilterType.Active:
+          return !completed && lowerCaseTitle.includes(lowerCaseQuery);
+
+        default:
+          return todos;
+      }
```
IvanFesenko, line 77: No need to filter when `query` is empty and is `FilterType.All`

`Array.filter` callback shoud return `true/false` - you have case when you return array

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filterType}
+            onChange={handleFilterTypeChange}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
IvanFesenko, line 39: You can use FilterType as value here

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +30,57 @@ export const TodoList: React.FC = () => (
...
+          </td>
+          <td className="has-text-right is-vcentered">
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => {
+                onSelectTodo(id);
+                onSelectUser(userId);
+              }}
```
IvanFesenko, line 72: create separate handler

IvanFesenko (changes requested):
Check my comments

src/App.tsx
```diff
@@ -17,18 +88,45 @@ export const App: React.FC = () => {
...
+                  Failed to fetch data. Please try again later.
+                </div>
+              )}
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {selectedTodoId > 0 && (
```
sTorba24, line 121:
```suggestion
      {!!selectedTodoId && (
```

sTorba24 (approved):
Good work! Some small comments to check :)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,43 @@
...
+    const fetchUser = async () => {
+      const userData = await getUser(userId);
+
+      setSelectedUser(userData);
+    };
+
+    fetchUser();
+  }, []);
+
+  const todo = todos.filter((tod => tod.id === todoId))[0];
```
ivadimko, line 34: Use `const todo = todos.find((tod) => tod.id === todoId);` instead of `filter` to get the first matching todo.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,43 @@
...
+
+  useEffect(() => {
+    const fetchUser = async () => {
+      const userData = await getUser(userId);
+
+      setSelectedUser(userData);
+    };
+
+    fetchUser();
+  }, []);
```
ivadimko, line 32: Add `[userId]` as a dependency to the `useEffect` to refetch the user data when `userId` changes.

src/App.tsx
```diff
@@ -1,14 +1,85 @@
...
+    const fetchTodos = async () => {
+      setIsLoading(true);
+      try {
+        const data = await getTodos();
+
+        setTodos(data);
+        setIsLoading(false);
+      } catch {
+        setError(true);
+        setIsLoading(false);
```
ivadimko, line 33: It's better to display a specific error message. Instead of using a generic error, catch the error and display it.

src/App.tsx
```diff
@@ -1,14 +1,85 @@
 /* eslint-disable max-len */
```
ivadimko, line 1: Consider removing the max-len eslint rule and configure the line length limit to a reasonable value instead of disabling it.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+/* eslint-disable max-len */
```
ivadimko, line 1: Consider removing the `/* eslint-disable max-len */` comment and adjusting the code to meet the max length requirement.

PR #1450, opened 2023-04-04
solution

Nordexz:
[DEMO LINK](https://nordexz.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -8,7 +7,45 @@ import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isError, setIsError] = useState(false);
+
+  const [query, setQuery] = useState('');
+  const [selectFilter, setSelectFilter] = useState('all');
+  const [activeTodoId, setActiveTodoId] = useState(0);
+
+  const activeTodo = todos.find(todo => todo.id === activeTodoId);
+
+  const getTodosList = useCallback(async (promise: Promise<Todo[]>) => {
+    try {
+      setTodos(await promise);
+    } catch {
+      setIsError(true);
+    }
+  }, []);
+
+  let VisibleTodos = todos;
+
+  if (selectFilter === 'active') {
+    VisibleTodos = VisibleTodos.filter(todo => !todo.completed);
+  }
+
+  if (selectFilter === 'completed') {
+    VisibleTodos = VisibleTodos.filter(todo => todo.completed);
+  }
+
+  VisibleTodos = VisibleTodos
+    .filter(todo => todo.title.toLocaleLowerCase()
+      .includes(query.toLocaleLowerCase()));
+
+  useEffect(() => {
+    getTodosList(getTodos());
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -8,7 +7,45 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+  let VisibleTodos = todos;
+
+  if (selectFilter === 'active') {
+    VisibleTodos = VisibleTodos.filter(todo => !todo.completed);
+  }
+
+  if (selectFilter === 'completed') {
+    VisibleTodos = VisibleTodos.filter(todo => todo.completed);
+  }
```
vladyslav-mosor, line 39: Maybe better use switch/case for that + create enum instead of comparing to strings?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,69 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={selectFilter}
+            onChange={(event) => setSelectFilter(event.target.value)}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
vladyslav-mosor, line 36: Definitely better to create enum so then you can map this options from enum values

vladyslav-mosor (changes requested):
You don`t show any errors when there is a problem when loading data so there is infinite loader problem here.
https://vimeo.com/814929487
https://vimeo.com/814930595

src/App.tsx
```diff
@@ -8,7 +7,56 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+  switch (sortType) {
+    case SortType.Active:
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+
+    case SortType.Completed:
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+
```
vladyslav-mosor, line 45: Better add SortType.All here as well as default case

src/App.tsx
```diff
@@ -8,7 +7,56 @@ import { TodoFilter } from './components/TodoFilter';
...
+  visibleTodos = visibleTodos
+    .filter(todo => todo.title.toLocaleLowerCase()
+      .includes(query.toLocaleLowerCase()));
+
+  const errorTodoList = todos.length > 0 && !isError && !isLoading;
+
+  useEffect(() => {
+    getTodosList(getTodos());
+  }, []);
```
vladyslav-mosor, line 58: I dont really understand why you pass getTodos function as a parameter to getTodosList. You could simply wait for the result of getTodos inside getTodosList. Check my suggestion please

src/App.tsx
```diff
@@ -8,7 +7,56 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+  const getTodosList = useCallback(async (promise: Promise<Todo[]>) => {
+    try {
+      setTodos(await promise);
+    } catch {
+      setIsError(true);
+    } finally {
+      setIsLoading(false);
+    }
+  }, []);
```
vladyslav-mosor, line 33:
```suggestion
  const getTodosList = useCallback(async () => {
    try {
      setTodos(await getTodos());
    } catch {
      setIsError(true);
    } finally {
      setIsLoading(false);
    }
  }, []);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,88 @@
...
+      </p>
+
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          placeholder="Search..."
+          value={query}
+          onChange={(event) => setQuery(event.target.value)}
```
vladyslav-mosor, line 67: Better move to separate handler

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,88 @@
...
+export const TodoFilter: React.FC<Props> = ({
+  todos,
+  setTodos,
+  query,
+  setQuery,
+  setSortType,
+}) => {
+  useEffect(() => {
+    setTodos(todos);
+  }, [query]);
```
vladyslav-mosor, line 22: I dont think you should do it here. So passing todos and setTodos to this component via props is not really nevessary

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,88 @@
...
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={handleSelectChange}
+          >
+            {Object.values(SortType).map(current => (
+              <option value={current}>
+                {`${current[0].toUpperCase() + current.slice(1)}`}
+              </option>
+            ))}
```
vladyslav-mosor, line 55: Dont forget to add 'key' when mapping

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,88 @@
...
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+interface Props {
+  todos: Todo[];
+  activeTodoId: number;
+  setActiveTodoId: React.Dispatch<React.SetStateAction<number>>;
```
vladyslav-mosor, line 8: Better destructure Dispatch and SetStateAction from react to improve readability

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,88 @@
...
+              <td className="is-vcentered">
+                {completed && (
+                  <span className="icon" data-cy="iconCompleted">
+                    <i className="fas fa-check" />
+                  </span>
+                )}
+              </td>
+              <td className="is-vcentered is-expanded">
+                <p className={classNames(
+                  { 'has-text-success': completed },
```
vladyslav-mosor, line 58:
```suggestion
                  'has-text-success',
```

vladyslav-mosor (changes requested):
There are still some small issues to fix

PR #1453, opened 2023-04-05
dynamic list of todos implemented

vladyslav-mosor:
[DEMO LINK](vladyslav-mosor.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,10 @@
 /* eslint-disable max-len */
-import React from 'react';
+import {
+  FC,
+  useCallback,
+  useEffect,
+  useState,
+} from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
+    try {
+      const loadedTodos = await getTodos();
+
+      setTodos(loadedTodos);
+      setLoading(false);
+    } catch (error) {
+      setHasError(true);
+      setLoading(false);
+    }
+  }, []);
```
valeriiayarzhemska, line 39: u can delete setLoading(false); from try and catch and just add it after this construction

src/helpers.ts
```diff
@@ -0,0 +1,34 @@
+import { Todo } from './types/Todo';
```
valeriiayarzhemska, line 1: i think it would be good if u create a folder for helpers

src/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+    case TodoStatus.ALL:
+    default:
+      break;
+  }
+
+  if (query) {
+    const normalizedQuery = query
+      .split(' ')
+      .map(word => word.trim().toLowerCase())
+      .join(' ');
```
valeriiayarzhemska, line 27: im not sure but do u need all of these checks? u cant just add query.trim().toLowerCase() in includes()?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,83 @@
...
+          <select
+            data-cy="statusSelect"
+            onChange={handleChange}
+          >
+            {Object.values(TodoStatus).map((enumValue) => (
+              <option
+                key={enumValue}
+                value={enumValue}
+              >
+                {enumValue}
```
valeriiayarzhemska, line 49: in example value of filters have first uppercased letter, idk if it really matters, but let it be:)

valeriiayarzhemska (changes requested):
gj! but a few issues to fix:)

src/App.tsx
vladyslav-mosor: Good point. Done!)

src/helpers.ts
vladyslav-mosor: Hmmm yes, that`s a bit overkilled

src/components/TodoFilter/TodoFilter.tsx
vladyslav-mosor: Tests are expected values to be in lowercase. I could possibly add a name using slice method but i think it`s not so critical in this case

valeriiayarzhemska (approved):

tiserett (approved):
Good job, but make sure to fix problems which I pointed 🔥

PR #1455, opened 2023-04-05
finished task

valeriiayarzhemska:
[DEMO LINK](https://valeriiayarzhemska.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
 /* eslint-disable max-len */
-import React from 'react';
+import {
+  FC,
+  useEffect,
+  useState,
+  useMemo,
+  useCallback,
+} from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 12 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,78 @@
...
+  onFilterChange: (filterType: FilterType) => void,
+};
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+export const TodoFilter: FC<Props> = (props) => {
```
sergey-lipikhin, line 11: Also can be wrapped into `React.memo` to re-render only if we change `query` or `filterType`, cause now it always re-renders with `App`. E.g. when select `todo` we don't want to re-render filter component. You can ensure of this behavior with `console.log('Render ${componentName}')` like I did :)

sergey-lipikhin (approved):

src/components/TodoFilter/TodoFilter.tsx
valeriiayarzhemska: > Also can be wrapped into `React.memo` to re-render only if we change `query` or `filterType`, cause now it always re-renders with `App`. E.g. when select `todo` we don't want to re-render filter component. You can ensure of this behavior with `console.log('Render ${componentName}')` like I did :)

cool, thanks:)

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+
+      try {
+        const todosFromServer = await getTodos();
+
+        setTodos(todosFromServer);
+      } catch {
+        setHasError(true);
+      }
+
+      setIsLoading(false);
```
tiserett, line 47: better put setIsLoading(false) inside finally block

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,58 @@
...
+
+      try {
+        const userFromServer = await getUser(userId);
+
+        setUser(userFromServer);
+      } catch {
+        setHasUserError(true);
+      }
+
+      setIsUserLoading(false);
```
tiserett, line 45: same as in App.tsx

src/helpers/helpers.ts
```diff
@@ -0,0 +1,27 @@
...
+export const getFilteredTodos = (
+  todos: Todo[],
+  filterType: FilterType,
+  query: string,
+) => {
+  const copyVisibleTodos = [...todos];
+
+  const searchedTodos = copyVisibleTodos.filter(({ title }) => {
+    return title.toLowerCase().includes(query.toLowerCase());
+  });
```
tiserett, line 13: not sure about creating copyVisibleTodos  array, since Filter method is immutable

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,78 @@
...
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filterType}
+            onChange={(event) => {
+              onFilterChange(event.target.value as FilterType);
+            }}
```
tiserett, line 28: you should be careful with using 'as', since it's not always safe

tiserett (approved):
Good work, but there are ways to improve🚀🚀

PR #1456, opened 2023-04-05
DynamicListOfTodos task done

MarkBabich:
[DEMO LINK](https://MarkBabich.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,52 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+const getVisibleTodos = (
+  todos: Todo[],
+  sortType: string,
+  query: string,
+): Todo[] => {
+  let filtered = todos;
+
+  if (sortType === 'active') {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,52 @@
...
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+const getVisibleTodos = (
+  todos: Todo[],
+  sortType: string,
+  query: string,
+): Todo[] => {
+  let filtered = todos;
```
tiserett, line 18: make a copy of goods array, to prevent possible problems

src/App.tsx
```diff
@@ -17,18 +55,39 @@ export const App: React.FC = () => {
...
+                  selectedTodo={selectedTodo}
+                />
+              )}
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {selectedTodo !== null && (
```
tiserett, line 85: this should work just fine as well if you remove !== null

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,19 +28,24 @@ export const TodoFilter = () => (
...
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
-        />
-      </span>
+      {query && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
tiserett, line 40: you can just add aria label to get rid of this problem ;)

tiserett (changes requested):
good work overall, but there are ways to make it even better 🔥

PR #1462, opened 2023-04-06
solution

kateryna-babina:
[DEMO LINK](https://kateryna-babina.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,62 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { Error } from './components/Error';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo>();
+  const [query, setQuery] = useState('');
+  const [selectedStatus, setSelectedStatus] = useState('all');
+  const [hasError, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    async function loadTodos() {
+      try {
+        const todosData = await getTodos();
+
+        setTodos(todosData);
+      } catch (error) {
+        setError('An error occurred while loading the goods.');
+      }
+    }
+
+    loadTodos();
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    let searchedTodos = [...todos];
+
+    if (query) {
+      searchedTodos = searchedTodos.filter(todo => (
+        todo.title.toLowerCase().includes(query.toLowerCase().trim())));
+    }
+
+    switch (selectedStatus) {
+      case ('active'): {
+        searchedTodos = searchedTodos.filter(todo => !todo.completed);
+        break;
+      }
+
+      case ('completed'): {
+        searchedTodos = searchedTodos.filter(todo => todo.completed);
+        break;
+      }
+
+      default: break;
+    }
+
+    return searchedTodos;
+  }, [query, selectedStatus, todos]);
+
+  const handleSelectTodo = (id: number) => {
+    const todo = todos.find(todoItem => todoItem.id === id);
+
+    setSelectedTodo(todo);
+  };
+
   return (
     <>
       <div className="section">
@@ -17,18 +71,32 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                query={query}
+                onChangeQuery={setQuery}
+                selectedStatus={selectedStatus}
+                onSelectStatus={setSelectedStatus}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              <TodoList
+                todos={visibleTodos}
+                onSelectTodo={handleSelectTodo}
+                selectedId={selectedTodo?.id}
+              />
+              {!todos.length && <Loader />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {hasError
+        && (
+          <Error hasError={hasError} onModalClose={() => setSelectedTodo(undefined)} />
+        )}
+
+      {selectedTodo && <TodoModal todo={selectedTodo} onClose={() => setSelectedTodo(undefined)} />}
     </>
   );
 };
--- a/src/components/Error/Error.tsx
+++ b/src/components/Error/Error.tsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import './Error.scss';
+
+type ErrorProps = {
+  hasError: string;
+  onModalClose: () => void;
+};
+
+export const Error: React.FC<ErrorProps> = ({ hasError, onModalClose }) => (
+  <div className="Error notification is-danger is-light">
+    <div className="Error-message">
+      {`Error: ${hasError}`}
+    </div>
+
+    {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+    <button
+      type="button"
+      className="button is-danger"
+      data-cy="modal-close"
+      onClick={() => onModalClose()}
+    >
+      Ok
+    </button>
+  </div>
+);
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,55 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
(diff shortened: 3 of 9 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,55 @@
...
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={(event) => onSelectStatus(event.target.value)}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
illiarerun, line 23: and use enum values here

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,55 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {query && (
+            // eslint-disable-next-line jsx-a11y/control-has-associated-label
```
illiarerun, line 43: don't disable linter - fix this problem

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,35 +57,42 @@ export const TodoModal: React.FC = () => {
...
-
+              {
+                completed
+                  ? (
+                    <strong className="has-text-success">Done</strong>
+                  )
+                  : (
+                    <strong className="has-text-danger">Planned</strong>
+                  )
+              }
```
illiarerun, line 86: use classnames and ternary for text, don't create 2 strong elements

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,25 @@
...
+  onModalClose: () => void;
+};
+
+export const Error: React.FC<ErrorProps> = ({ hasError, onModalClose }) => (
+  <div className="Error notification is-danger is-light">
+    <div className="Error-message">
+      {`Error: ${hasError}`}
+    </div>
+
+    {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
illiarerun, line 15: don't disbale linter, fix the problem

src/App.tsx
```diff
@@ -17,18 +71,32 @@ export const App: React.FC = () => {
...
         </div>
       </div>
 
-      <TodoModal />
+      {hasError
+        && (
+          <Error hasError={hasError} onModalClose={() => setSelectedTodo(undefined)} />
+        )}
+
+      {selectedTodo && <TodoModal todo={selectedTodo} onClose={() => setSelectedTodo(undefined)} />}
```
illiarerun, line 99: let's not write everything in one line

src/App.tsx
```diff
@@ -17,18 +71,32 @@ export const App: React.FC = () => {
...
+              {!todos.length && <Loader />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {hasError
+        && (
+          <Error hasError={hasError} onModalClose={() => setSelectedTodo(undefined)} />
```
illiarerun, line 96: here too

illiarerun (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
+      } catch (error) {
+        setError('An error occurred while loading the goods.');
+      }
+    }
+
+    loadTodos();
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    let searchedTodos = [...todos];
```
Oleksandr-kopaevich, line 37: Method `filter` returns a new array every time, so you don’t need to copy it with the spread operator.

src/App.tsx
```diff
@@ -17,18 +73,43 @@ export const App: React.FC = () => {
...
+                onChangeQuery={setQuery}
+                selectedStatus={selectedStatus}
+                onSelectStatus={setSelectedStatus}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length && (
```
Oleksandr-kopaevich, line 85: `todos.length` returns a number and it can be zero. In this case react will render `0` on the page, but you don't want to render anything. You need to transform a number to boolean, because for false react will not render anything at all. One of the ways to do it is like this.
```suggestion
              {Boolean(todos.length) && (
```

src/App.tsx
```diff
@@ -17,18 +73,43 @@ export const App: React.FC = () => {
...
+                <TodoList
+                  todos={visibleTodos}
+                  onSelectTodo={handleSelectTodo}
+                  selectedId={selectedTodo?.id}
+                />
+              )}
+              {!todos.length && <Loader />}
```
Oleksandr-kopaevich, line 92: What if there is no todos? You'll show an infinite loader which is not correct. You need another variable in state called for example `isLoading` or `isTodosLoading`. You set it to `true` before calling `getTodos` and set it to `false` after try/catch block in the function `loadTodos`

src/App.tsx
```diff
@@ -17,18 +73,43 @@ export const App: React.FC = () => {
...
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {hasError
+        && (
+          <Error
+            hasError={hasError}
+            onModalClose={() => setSelectedTodo(undefined)}
```
Oleksandr-kopaevich, line 102: Undefined mostly used when there is no value at all. But when we want to reset or clear something, we use null. In practice it'll work the same, but that's how most developers write, so let's keep the style.

Oleksandr-kopaevich, line 102: I'd call the props `onErrorClose` or just `onClose`

src/components/Error/Error.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <div className="Error-message">
+      {`Error: ${hasError}`}
+    </div>
+
+    <button
+      aria-label="modal-close"
+      type="button"
+      className="button is-danger"
+      data-cy="modal-close"
+      onClick={() => onModalClose()}
```
Oleksandr-kopaevich, line 20: You can use a shorter form by just passing reference to the function. You can use it every time when you get the same params that needed to this nested function.
```suggestion
      onClick={onModalClose}
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,61 @@
...
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={(event) => onSelectStatus(event.target.value)}
+          >
+            {Object.entries(SortTypes).map(([key, value]) => (
+              <option
```
Oleksandr-kopaevich, line 24: When you map childrens in react the highest level component must have a key attribute with unique value
```suggestion
              <option
                  key={value}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,55 @@
...
+    title,
+    completed,
+  } = todo;
+
+  const [isLoading, setIsLoading] = useState(false);
+  const [user, setUser] = useState<User>();
+  const [hasError, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    setIsLoading(true);
```
Oleksandr-kopaevich, line 27: When you have code blocks it's better to have empty line to separate them
```suggestion
    setIsLoading(true);
    
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,35 +58,44 @@ export const TodoModal: React.FC = () => {
               className="modal-card-title has-text-weight-medium"
               data-cy="modal-header"
             >
-              Todo #2
+              {`Todo #${id}`}
```
Oleksandr-kopaevich, line 61: This will have the same effect, but easier to read
```suggestion
              Todo #{id}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,35 +58,44 @@ export const TodoModal: React.FC = () => {
...
             </div>
 
-            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
             <button
+              aria-label="modal-close"
               type="button"
               className="delete"
               data-cy="modal-close"
+              onClick={() => onClose()}
```
Oleksandr-kopaevich, line 69:
```suggestion
              onClick={onClose}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,35 +58,44 @@ export const TodoModal: React.FC = () => {
...
+                )}
+              >
+                {completed ? 'Done' : 'Planned'}
+              </strong>
               {' by '}
 
               <a href="mailto:Sincere@april.biz">
-                Leanne Graham
+                {user?.name}
               </a>
```
Oleksandr-kopaevich, line 94: when there is no name, then there is no sense to render an empty link. Try to avoid redundant html generation. And you missed to update the href attribute value, we don't want one user to get all emails.
```
 {user?.name && 
     <> 
         by 
         <a href={`mailto:${user.email}`}>
            {user.name}
         </a>
     </>
 }
```

Oleksandr-kopaevich (approved):
Hey @kateryna-babina your demo is working fine, all points from the task checklist are reached, so approved. But check my comments, there are some important things for you to remember to avoid in the future and fix.

PR #1468, opened 2023-04-06
React dynamic list of todos - Solution

Yevhenii0536:
[DEMO LINK](https://yevhenii0536.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,44 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
+import { getTodos } from './api';
+import { filterTodos } from './components/Helpers/FilterTodos';
+import { getErrorMessage } from './components/Helpers/GetErrorMessage';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [sortBy, setSortBy] = useState('all');
+  const [searchValue, setSearchValue] = useState('');
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  const visibleTodos = filterTodos(todosList, sortBy, searchValue);
+
+  const fetchTodos = async () => {
+    try {
+      const todos = await getTodos();
+
+      setTodosList(todos);
+    } catch (error) {
+      setErrorMessage(getErrorMessage(error));
+    } finally {
+      setIsLoaded(false);
+    }
+  };
+
+  useEffect(() => {
+    setIsLoaded(true);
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,14 +1,44 @@
...
 import { TodoModal } from './components/TodoModal';
+import { getTodos } from './api';
+import { filterTodos } from './components/Helpers/FilterTodos';
+import { getErrorMessage } from './components/Helpers/GetErrorMessage';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [sortBy, setFilterBy] = useState('all');
```
tiserett, line 18: we are filtering, not sorting, filterType is a better name
also, create enum for storing possible values, don't forget to use it everywhere, not only here

src/App.tsx
```diff
@@ -1,14 +1,44 @@
...
+      setTodosList(todos);
+    } catch (error) {
+      setErrorMessage(getErrorMessage(error));
+    } finally {
+      setIsLoaded(false);
+    }
+  };
+
+  useEffect(() => {
+    setIsLoaded(true);
```
tiserett, line 38: better to set loading start inside of fetchTodos function

src/App.tsx
```diff
@@ -17,18 +47,37 @@ export const App: React.FC = () => {
...
+                <>
+                  {visibleTodos.length ? (
+                    <TodoList
+                      todos={visibleTodos}
+                      selectedTodo={selectedTodo}
+                      setSelectedTodo={setSelectedTodo}
+                    />
+                  ) : (
+                    <p>{errorMessage}</p>
+                  )}
```
tiserett, line 70: you should show loader when loading
show error if has error
show todos if there are no error and you're not loading
also, try to avoid nesting ternary operators

src/components/Helpers/FilterTodos.tsx
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  if (query) {
+    return filtered
+      .filter((todo) => todo
+        .title.toLowerCase()
+        .includes(searchQuery));
+  }
+
+  return filtered;
```
tiserett, line 31: you can get rid of if block, if there is nothing in query, and you make sure there are no whitespaces, this function will work fine without it

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,70 @@
...
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select data-cy="statusSelect" onChange={handleChangeSelect}>
+            <option value="all">All</option>
+
+            <option value="active">Active</option>
+
+            <option value="completed">Completed</option>
```
tiserett, line 37: can be simplified, when you use ENUM ;)

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,100 @@
...
+      <tbody>
+        {todos.map((todo) => {
+          const { title, id, completed } = todo;
+
+          return (
+            <tr
+              data-cy="todo"
+              className={classNames(
+                {
+                  'has-background-info-light': selectedTodo?.id === todo.id,
```
tiserett, line 50: selectedTodo?.id === todo.id you use this code twice, better create variable for it

tiserett (changes requested):
well done, but there are ways to make it even better

PR #1471, opened 2023-04-07
Solution

katya-suvorova:
https://katya-suvorova.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+      switch (statusSelect) {
+        case Status.Active:
+          return !todo.completed;
+        case Status.Completed:
+          return todo.completed;
+        default:
+          return true;
+      }
+    });
+  }
```
anstsot, line 41: instead of modifying variable `useMemo` can be used
```suggestion
  const visibleTodos = useMemo(() => [...todos]
    .filter(todo => (
      todo.title.toLowerCase().includes(query.toLowerCase().trim())
    )).filter((todo) => {
      switch (statusSelect) {
        case Status.Active:
          return !todo.completed;
        case Status.Completed:
          return todo.completed;
        default:
          return true;
      }
    }), [todos, statusSelect, query]);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +29,53 @@ export const TodoList: React.FC = () => (
...
+              {todo.title}
+            </p>
+          </td>
+          <td className="has-text-right is-vcentered">
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => {
+                return setSelectedTodo(todo);
```
anstsot, line 65: there is no need to return the value
```suggestion
                setSelectedTodo(todo);
```

anstsot (approved):
Good job, approving. There are some improvements (check comments)

PR #1472, opened 2023-04-07
done

AntonChetverykMA:
[DEMO LINK](https://AntonChetverykMA.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,49 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [activeTodo, setActiveTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isOpenModal, setIsOpenModal] = useState(false);
+  const [sortBy, setSortBy] = useState('all');
+  const [query, setQuery] = useState('');
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos().then((data) => {
+      setTodos(data);
+      setIsLoading(false);
+    });
+  }, []);
+
+  const sort = () => {
+    const todosCopy = [...todos].filter(todo => todo.title.includes(query.trim()));
+
+    if (sortBy !== 'all') {
+      return todosCopy.filter(todo => {
+        if (sortBy === 'completed') {
+          return todo.completed === true;
+        }
+
+        return todo.completed === false;
+      });
+    }
+
+    return todosCopy;
+  };
+
+  const sortedList = sort();
+
   return (
     <>
       <div className="section">
@@ -17,18 +52,37 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                setSortBy={setSortBy}
+                sortBy={sortBy}
+                query={query}
+                setQuery={setQuery}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading
+                ? <Loader />
+                : (
+                  <TodoList
+                    todos={sortedList}
+                    setIsOpenModal={setIsOpenModal}
+                    setActiveTodo={setActiveTodo}
+                    activeTodo={activeTodo}
+                  />
+                )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      { isOpenModal && (
+        <TodoModal
+          setIsOpenModal={setIsOpenModal}
+          activeTodo={activeTodo}
+          setActiveTodo={setActiveTodo}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,69 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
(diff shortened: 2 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
+        }
+
+        return todo.completed === false;
+      });
+    }
+
+    return todosCopy;
+  };
+
+  const sortedList = sort();
```
anstsot, line 45: you can use `useMemo` hook

anstsot (changes requested):
Good work but there is one issue to fix. The search shouldn't be case sensitive so it should filter even with large chars
image

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,111 @@
...
+
+    if (sortBy !== 'all') {
+      return todosCopy.filter(todo => {
+        if (sortBy === 'completed') {
+          return todo.completed === true;
+        }
+
+        return todo.completed === false;
+      });
+    }
```
sTorba24, line 42: use switch case

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,111 @@
...
+    const convertedQuery = query.toLowerCase().trim();
+
+    const todosCopy = [...todos].filter(todo => {
+      return todo.title.toLowerCase().includes(convertedQuery);
+    });
+
+    if (sortBy !== 'all') {
+      return todosCopy.filter(todo => {
+        if (sortBy === 'completed') {
+          return todo.completed === true;
```
sTorba24, line 37:
```suggestion
          return todo.completed;
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,111 @@
...
+        }
+
+        return todo.completed === false;
```
sTorba24, line 40:
```suggestion
        return !todo.completed;
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,111 @@
...
+              <td className="is-vcentered">{todo.id}</td>
+              <td className="is-vcentered">
+                {todo.completed && (
+                  <span className="icon" data-cy="iconCompleted">
+                    <i className="fas fa-check" />
+                  </span>
+                )}
+              </td>
+              <td className="is-vcentered is-expanded">
+                <p className={`has-text-${!todo.completed ? 'danger' : 'success'}`}>{todo.title}</p>
```
sTorba24, line 79: use classNames utility
https://www.npmjs.com/package/classnames

sTorba24 (changes requested):
Good work! lets improve it :)

src/components/TodoList/TodoList.tsx
AntonChetverykMA: i think it's not convenient use ternary operator with classnames

src/App.tsx
```diff
@@ -1,14 +1,31 @@
...
+  const [sortBy, setSortBy] = useState('all');
+  const [query, setQuery] = useState('');
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos().then((data) => {
+      setTodos(data);
+      setIsLoading(false);
+    });
+  }, []);
```
TarasHoliuk, line 27: Consider wrapping your `useEffect` cleanup function with a try-catch block to handle any errors that may occur during the API call or state update. This will provide a better user experience, especially for error handling. 🛠️

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,111 @@
...
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye-slash" />
-            </span>
-          </button>
-        </td>
-      </tr>
+export const TodoList: React.FC<TodoListProps> = (props: TodoListProps) => {
```
TarasHoliuk, line 13: No need to type annotate `props: TodoListProps` since we are already providing `TodoListProps` as a generic to `React.FC`.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,47 +1,83 @@
...
+  activeTodo: Todo | null,
+  setActiveTodo: React.Dispatch<React.SetStateAction<Todo | null>>
+}
+
+export const TodoModal: React.FC<TodoModalProps> = (props: TodoModalProps) => {
+  const { setIsOpenModal, activeTodo, setActiveTodo } = props;
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  const closeModal = () => {
```
TarasHoliuk, line 18: It's better to use useCallback for closeModal to avoid unnecessary re-renders.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,47 +1,83 @@
...
+    setActiveTodo(null);
+  };
+
+  useEffect(() => {
+    if (!activeTodo) {
+      return;
+    }
+
+    getUser(activeTodo.userId).then((data) => {
+      setUser(data);
```
TarasHoliuk, line 29: Consider handling errors in the 'getUser' promise using a catch block to prevent unhandled rejections.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,69 @@
...
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => setQuery('')}
```
TarasHoliuk, line 62: avoid using the anonymous function
it's better to create a separate handler outside JSX and use it here (e.g `resetQuery`)

src/components/TodoList/TodoList.tsx
TarasHoliuk: not fixed

src/components/TodoList/TodoList.tsx
TarasHoliuk: not fixed

src/components/TodoList/TodoList.tsx
TarasHoliuk: not fixed

TarasHoliuk (changes requested):
Please, check the previous comments and fix them
Also, I've added a few additional comments

If you have any questions - feel free to ask in the chat

PR #1485, opened 2023-04-09
task completed

ArthurKrepko:
[DEMO LINK](https://ArthurKrepko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+      case FilterStatus.Completed:
+        isStatusCorrect = todo.completed;
+        break;
+
+      default:
+        break;
+    }
+
+    return isStatusCorrect && lowerCasedTitle.includes(lowerCasedQuery);
+  });
```
tiserett, line 57: you can make this code better, by getting rid of this switch block, since it's not really necessary in your case, think on how to ;)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,73 @@
...
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          placeholder="Search..."
+          value={filterQuery}
+          onChange={(event) => (
+            setFilterQuery(event.target.value)
+          )}
```
tiserett, line 53: create separate handler, also, prevent user from entering many whitespaces

tiserett (changes requested):
good work, but few small changes required

PR #1488, opened 2023-04-10
solution

RuslanLiapin:
[DEMO LINK](https://ruslanliapin.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+export const App: React.FC = React.memo(() => {
+  const [todos, setTodos] = useState< {
+    todos: Todo[]
+    originalTodos: Todo[];
+    searchQueryTodo: Todo[];
+  }>({
+    todos: [],
+    originalTodos: [],
+    searchQueryTodo: [],
+  });
```
anstsot, line 23: it's better not to use one state for these 3 variables. you can save only all original todos (coming from API). then you can create a filter and search query states and use `useMemo` hook for filtering original todos. it will be better solution.

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+  const [selectedTodoId, setSelectedTodoId] = useState<number | null>(null);
+  const [user, setUser] = useState<{
+    user: User;
+    buttonClicked: boolean;
```
anstsot, line 27: why do you need `buttonClicked` ? could you work only with selectedTodoId?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,98 @@
...
+        break;
+      default:
+    }
+
+    setTodos(prev => ({
+      ...prev,
+      todos: filteredTodos,
+      searchQueryTodo: filteredTodos,
+    }));
+  };
```
anstsot, line 53: move this code to App.tsx. in this component would be better just to show and change filters

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,46 @@
...
+
+  const selectedTodo = todos.find(todo => todo.id === selectedTodoId) || null;
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {user.buttonClicked === true && user.user.id === 0 ? (
```
anstsot, line 43:
```suggestion
      {!selectedTodo ? (
```

anstsot (changes requested):
Good job, but there are few fixes to improve and simplify your code.
Please also fix next issue - The loader is shown when there are no filtered todos.
image

sTorba24 (approved):
Good work! Approved :)

PR #1508, opened 2023-04-22
React dynamic list of todos

ruslan-yarosh:
[DEMO LINK](https://ruslan-yarosh.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
 /* eslint-disable max-len */
-import React from 'react';
+// eslint-disable-next-line object-curly-newline
+import React, { ChangeEvent, useEffect, useMemo, useState } from 'react';
+
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,87 @@
...
-          <span className="icon" data-cy="iconCompleted">
-            <i className="fas fa-check" />
-          </span>
-        </td>
-        <td className="is-vcentered is-expanded">
-          <p className="has-text-success">quo adipisci enim quam ut ab</p>
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
+    icon.className = 'far fa-eye';
```
ivbutsykin, line 22: Bad idea. If you want to change `className` for element, you should do this directly in `jsx`.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,87 @@
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye-slash" />
-            </span>
-          </button>
-        </td>
-      </tr>
+export const TodoList: React.FC<Props> = ({ todos, onBtnClick }) => {
+  const getIcon = (target: HTMLButtonElement) => target.children[0].children[0];
```
ivbutsykin, line 11: Try to realise flow, without such function. It's violates principles of React

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { SortType } from './types/SortType';
+
+const getVisibleTodos = (filterCase: string, todos: Todo[], searchQuery: string) => {
+  const searchedTodo = todos.filter(todo => todo.title.toLowerCase().includes(searchQuery.trim().toLowerCase()));
```
ivbutsykin, line 17: I can suggest to add condition like this:
```
if (!!searchQuery.trim()) {
    ...
}
```
Because now you make expensive calculations even then they are no need

ivbutsykin (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,84 @@
...
+    });
+    setIsLoading(false);
+  }, []);
+
+  const handleChangeModalInfo = (todo: Todo | null) => {
+    setModalInfo(todo);
+  };
+
+  const filterChange = (event: ChangeEvent<HTMLSelectElement>) => {
+    if (event.target.value === 'all') {
```
vlad-demchuk, line 59: better to use switch 

and use enum values for comparison instead of strings `all, active, completed`

src/App.tsx
```diff
@@ -1,14 +1,84 @@
...
+
+  if (searchQuery.trim()) {
+    searchedTodo = todos.filter(todo => todo.title.toLowerCase().includes(searchQuery.trim().toLowerCase()));
+  }
+
+  switch (filterCase) {
+    case SortType.ACTIVE:
+      return searchedTodo === undefined
+        ? todos.filter(todo => !todo.completed)
+        : searchedTodo.filter(todo => !todo.completed);
```
vlad-demchuk, line 27: let's filter by status first and then filter by query to avoid this ternary

vlad-demchuk (changes requested):

PR #1510, opened 2023-04-24
done

andrudenko:
[DEMO LINK](https://andrudenko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,96 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import React from 'react';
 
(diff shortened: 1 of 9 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,96 @@
...
+    const nameFirstLetterUpperCased = filterByOptionSplit[0].toUpperCase();
+    const nameRemainingPart = filterByOption.slice(1);
+    const name = nameFirstLetterUpperCased + nameRemainingPart;
+
+    return {
+      ...names,
+      [filterByOption]: name,
+    };
+  },
+  {} as FilterByOptionsNames,
```
DarkMistyRoom, line 20: Consider uppercase the first letter of each option instead of this logic, it may be simpler

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,61 @@ export const TodoList: React.FC = () => (
...
-            </span>
-          </button>
-        </td>
-      </tr>
+        return (
+          <tr
+            key={id}
+            data-cy="todo"
+            className=""
+          >
```
DarkMistyRoom, line 44:
```suggestion
          <tr
            key={id}
            data-cy="todo"
          >
```

PR #1511, opened 2023-04-24
dynamic-list-of-todos was done

pushkarskiyrodion:
[DEMO LINK](https://pushkarskiyrodion.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -8,7 +7,52 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+              default:
+                return todoItem;
+            }
+          }))
+        .then(result => setTodos(result));
+    };
+
+    fetchData();
+  }, [filterValue]);
```
anstsot, line 46: why do you need to load todos from server each time the `filterValue` is changed? please load them once from API and just use `useMemo` hook to filter by filter change

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,75 @@
...
+  onSelectFilter,
+  onApplyQuery,
+}) => {
+  const handleSelectChange = (e: React.BaseSyntheticEvent) => {
+    onSelectFilter(e.target.value);
+  };
+
+  const handleInputChange = (e: React.BaseSyntheticEvent) => {
+    onSetQuery(e.target.value);
+    onApplyQuery(e.target.value);
```
anstsot, line 24: do you really need to use these 2 functions here? it would be better to use `debounce` in `onSetQuery` only. please change the logic

anstsot (changes requested):
It’s almost done! Let’s improve a little

src/components/TodoFilter/TodoFilter.tsx
pushkarskiyrodion: I did it by example from the video. Misha explains it in this video - https://mate.academy/learn/react/advanced-hooks?section=video&videoId=770

PR #1516, opened 2023-04-27
solution

MSA1991:
[DEMO LINK](https://msa1991.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -8,7 +8,66 @@ import { TodoFilter } from './components/TodoFilter';
...
+  };
+
+  const handleSelectedTodo = (todoId: number) => {
+    setSelectedTodoId(todoId);
+  };
+
+  const getFilteredTodos = () => {
+    let newTodos = [...todos];
+
+    if (filterTodos === 'active') {
```
tetiana-ivanchenko, line 49: create enum for filter types

src/components/TodoList/TodoList.tsx
```diff
@@ -14,87 +25,48 @@ export const TodoList: React.FC = () => (
...
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+        return (
+          <tr
+            data-cy="todo"
+            className={isSelected ? 'has-background-info-light' : ''}
```
tetiana-ivanchenko, line 36: use classnames library

src/components/TodoList/TodoList.tsx
```diff
@@ -14,87 +25,48 @@ export const TodoList: React.FC = () => (
...
+            </td>
+            <td className="is-vcentered">
+              {completed && (
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              )}
+            </td>
+            <td className="is-vcentered is-expanded">
+              <p className={completed ? 'has-text-success' : 'has-text-danger'}>
```
tetiana-ivanchenko, line 50: the same

tetiana-ivanchenko (changes requested):

src/components/TodoList/TodoList.tsx
MSA1991: В classnames такой ужасный синтаксис, без него как по мне все читабельней(

src/App.tsx
```diff
@@ -8,7 +8,67 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+    if (filterTodos === Filter.Active) {
+      newTodos = newTodos.filter(todo => !todo.completed);
+    }
+
+    if (filterTodos === Filter.Completed) {
+      newTodos = newTodos.filter(todo => todo.completed);
+    }
+
+    if (query.trim() !== '') {
```
sTorba24, line 58:
```suggestion
    if (query.trim()) {
```

sTorba24 (approved):
Good work! 
About classnames - if you have only one condition it can be a little simplier without it, but when you have 2 or more classes dependent from condition, it looks much better with using of classNames utility , you have this case on task. 
And it's cool to write classes dependent from condition in the same style, so it's fine to use it even for 1 condition to be consistent.

It's like function that depends from n-conditions. More you have - smarter to use classNames is. It's cool to practice this approach (using classnames) from start :)

src/App.tsx
MSA1991: Спасибо большое за разъяснение

PR #1521, opened 2023-05-01
All todos have been loaded

bojkovladislav:


src/App.tsx
```diff
@@ -8,7 +7,42 @@ import { TodoFilter } from './components/TodoFilter';
...
 import { Loader } from './components/Loader';
 
+import { Todo } from './types/Todo';
+
+import { getTodos } from './api';
+
 export const App: React.FC = () => {
+  const [visibleTodos, setVisibleTodos] = useState<Todo[]>([]);
+  const [visibleTodo, setVisibleTodo] = useState<Todo | null>(null);
+  const [value, setValue] = useState('');
```
anstsot, line 17: rename the variable to more meaningful name. and it would be great to introduce `enum Filter` type and use it everywhere
```suggestion
  const [filter, setFilter] = useState(Filter.All);
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

mykhalenych (approved):

PR #1530, opened 2023-05-05
Develop

nazar-medushevskyi:
- [DEMO LINK](https://nazar-medushevskyi.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+      filteredTodos = filteredTodos.filter(todo => {
+        return (
+          filter === 'completed'
+            ? todo.completed
+            : !todo.completed
+        );
+      });
+    }
+
+    if (serchQuery.trim() !== '') {
```
loralevitska, line 33:
```suggestion
    if (!serchQuery.trim()) {
```

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [listOfTodos, setTodoList] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState('all');
+  const [serchQuery, setSearchQuery] = useState('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  const filterTodos = () => {
```
loralevitska, line 20: better make it variable not a function. And add [useMemo](https://dmitripavlutin.com/react-usememo-hook/) for it and `filterTodos` become variable

src/App.tsx
```diff
@@ -17,18 +66,33 @@ export const App: React.FC = () => {
...
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading ? (
+                <Loader />
+              ) : (
+                <TodoList
+                  listOfTodos={filterTodos()}
```
loralevitska, line 80:
```suggestion
                  listOfTodos={filterTodos}
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,72 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {isClearButton && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
loralevitska, line 59: Try to fix eslint rule and remove comment

loralevitska (changes requested):
Good job! There are things that should be fixed:

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+
+  enum Filter {
+    All = 'all',
+    Completed = 'completed',
+    Incomplete = 'incomplete',
+  }
+
+  const filterTodos = () => {
+    let filteredTodos: Todo[] = [...listOfTodos];
```
SanyaBratashchuk, line 27: why do you need to make a copy of array?

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+
+    if (filter !== Filter.All) {
+      filteredTodos = filteredTodos.filter(todo => {
+        return (
+          filter === Filter.Completed
+            ? todo.completed
+            : !todo.completed
+        );
+      });
+    }
```
SanyaBratashchuk, line 37: In my opinion switch(filter) will be more flexible

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  enum Filter {
+    All = 'all',
+    Completed = 'completed',
+    Incomplete = 'incomplete',
+  }
+
+  const filterTodos = () => {
```
SanyaBratashchuk, line 26: use useMemo

SanyaBratashchuk (changes requested):

src/App.tsx
nazar-medushevskyi: Making a copy of the array ensures that the original array is not mutated when the filtered array is created

src/App.tsx
nazar-medushevskyi: using switch will require more code to achieve the same result as the current code

SanyaBratashchuk: But it will be more flexible in case you will have more Filter options

src/App.tsx
SanyaBratashchuk: Does filter mutate an array?

SanyaBratashchuk (approved):

PR #1590, opened 2023-05-17
Solution

LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,35 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>();
+  const [searchValue, setSearchValue] = useState('');
+  const [selectValue, setSelectValue] = useState('all');
+  const [currentTodo, setCurrentTodo] = useState<Todo | null>(null);
+
+  const todoFilter = (): Todo[] | void => {
+    const result = selectValue === 'all'
+      ? todos
+      : todos?.filter(todo => {
+        return selectValue === 'active' ? !todo.completed : todo.completed;
+      });
+
+    return result ? result.filter(todo => todo.title.replace(' ', '').toLowerCase()
+      .includes(searchValue.toLowerCase().trim()))
+      : undefined;
+  };
+
+  useEffect(() => {
+    getTodos().then((items: Todo[]) => setTodos(items));
+  }, []);
+
+  const todoFilterList = todoFilter();
+  const onChangeSelect = (value: string) => setSelectValue(value);
+  const onChangeSearchValue = (value: string) => setSearchValue(value);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


polosanya (changes requested):
Good job! Let's improve your code a bit)

LBondarchuk:
Hello. I changed everything, but the select tests give errors, this is because the enum has a different case. It seems to me that we used to write them in upper case

src/types/SELECT.ts
```diff
@@ -0,0 +1,5 @@
+export enum SELECT {
+  ALL = 'All',
+  ACTIVE = 'Active',
+  COMPLETED = 'Completed',
```
agniya-i, line 4:
```suggestion
  ALL = 'all',
  ACTIVE = 'active',
  COMPLETED = 'completed',
```

agniya-i (changes requested):

src/types/SELECT.ts
```diff
@@ -0,0 +1,5 @@
+export enum SELECT {
```
polosanya, line 1: Consider renaming the enum to 'Select' to follow the naming convention of PascalCase for enums.

src/App.tsx
```diff
@@ -1,14 +1,57 @@
...
+    return result ? result.filter(todo => todo.title.replaceAll(' ', '').toLowerCase()
+      .includes(searchValue.toLowerCase().trim()))
+      : [];
+  };
+
+  useEffect(() => {
+    getTodos().then((items: Todo[]) => setTodos(items));
+  }, []);
+
+  const todoFilterList = useMemo(() => todoFilter(), [searchValue, selectValue, todos.length]);
```
polosanya, line 47: Instead of using 'todos.length' as a dependency, you should use the entire 'todos' array. Change the dependency from 'todos.length' to 'todos'.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,28 @@
...
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
+        <select
+          data-cy="statusSelect"
+          onChange={(e) => {
+            onChangeSelect(SELECT[e.target.value as keyof typeof SELECT]);
+          }}
+        >
+          {Object.entries(SELECT).map(item => {
+            return <option value={item[0]}>{item[1]}</option>;
```
polosanya, line 24: Add a 'key' prop to the <option> element to comply with React's list rendering requirements.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,28 @@
...
     <p className="control">
       <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
+        <select
+          data-cy="statusSelect"
+          onChange={(e) => {
+            onChangeSelect(SELECT[e.target.value as keyof typeof SELECT]);
```
polosanya, line 20: Consider using a separate function for the 'onChange' event handler instead of using an inline function

src/types/SELECT.ts
polosanya: Not fixed 
image

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,68 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { Todo } from '../../types/Todo';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
+
+type Props = {
+  currentTodo: Todo | null
```
polosanya, line 8: What's the purpose of rendering TodoModal if currentTodo is null?

src/components/TodoFilter/TodoFilter.tsx
polosanya, line 20:
```suggestion
            onChangeSelect(e.target.value as Select);
```

polosanya (changes requested):
Almost done! Try to pass the tests

src/types/SELECT.ts
LBondarchuk: Hello. I can't seem to change it so that the variables are in upper case and the values are in lower case (when iterating dynamically). I also did not find an answer to this question in the chat. allowed to send in this version. I made it to fit the tests. but variables are lowercase. If it's important to you, I can make all these selects non-dynamic. Or you can tell me how to do it correctly

src/components/TodoFilter/TodoFilter.tsx
LBondarchuk: it doesn't work like that

src/components/TodoModal/TodoModal.tsx
LBondarchuk: when there is no null rendering. Render only when not null. This is to reset currentTodo

src/components/TodoFilter/TodoFilter.tsx
LBondarchuk: it doesn't work like that

OleziO (approved):

PR #1608, opened 2023-05-27
develop

nata-prog123:
[DEMO LINK](https://nata-prog123.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,7 +12,7 @@ import { Todo } from './types/Todo';
 import { getTodos } from './api';
 import { FilterBy } from './types/Filter';
 
-const filteredTodos = (todos: Todo[], query: string, filterBy: FilterBy) => {
+const getFilteredTodos = (todos: Todo[], query: string, filterBy: FilterBy) => {
   let newTodos = [...todos];
 
   if (query) {
@@ -42,7 +42,7 @@ export const App: React.FC = () => {
   const [query, setQuery] = useState('');
   const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
 
-  const [load, setLoad] = useState(true);
+  const [isLoading, setisLoading] = useState(true);
   const [error, setError] = useState(false);
 
   const getTodosServer = async () => {
@@ -50,7 +50,7 @@ export const App: React.FC = () => {
       const arrayTodos = await getTodos();
 
       setTodos(arrayTodos);
-      setLoad(false);
+      setisLoading(false);
       setError(false);
     } catch {
       setError(true);
@@ -61,15 +61,11 @@ export const App: React.FC = () => {
     getTodosServer();
   }, []);
 
-  const visibleTodos = filteredTodos(todos, query, filterBy);
-
-  const showTodo = useCallback((todo: Todo) => {
-    setSelectedTodo(todo);
-  }, [selectedTodo]);
+  const visibleTodos = getFilteredTodos(todos, query, filterBy);
 
   const closeTodo = useCallback(() => {
     setSelectedTodo(null);
-  }, [selectedTodo]);
+  }, []);
 
   return (
     <>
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -8,7 +8,69 @@ import { TodoFilter } from './components/TodoFilter';
...
+  return newTodos;
+};
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ALL);
+
+  const [load, setLoad] = useState(true);
```
TarasHoliuk, line 45:
```suggestion
  const [isLoading, setIsLoading] = useState(true);
```
about boolean variables naming: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/App.tsx
```diff
@@ -8,7 +8,69 @@ import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { FilterBy } from './types/Filter';
+
+const filteredTodos = (todos: Todo[], query: string, filterBy: FilterBy) => {
```
TarasHoliuk, line 15:
```suggestion
const getFilteredTodos = (todos: Todo[], query: string, filterBy: FilterBy) => {
```
or `filterTodos` would be the correct name for the function
`filteredTodos` sounds like In this variable stored an array of filtered todos, not the function

src/App.tsx
```diff
@@ -8,7 +8,69 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+  useEffect(() => {
+    getTodosServer();
+  }, []);
+
+  const visibleTodos = filteredTodos(todos, query, filterBy);
+
+  const showTodo = useCallback((todo: Todo) => {
+    setSelectedTodo(todo);
+  }, [selectedTodo]);
```
TarasHoliuk, line 68: wrong value in the dependency array
you shouldn't recreate this function every time when selectedTodo changes. Empty dependency array or `setSelectedTodo` there would be correct choice

But actually, you don't need this wrapper at all

src/App.tsx
```diff
@@ -17,18 +79,36 @@ export const App: React.FC = () => {
...
             <div className="block">
-              <Loader />
-              <TodoList />
+              {load && <Loader />}
+              {error
+                ? <p>Error, server is unavailable</p>
+                : (
+                  <TodoList
+                    todos={visibleTodos}
+                    showTodo={showTodo}
```
TarasHoliuk, line 97:
```suggestion
                    showTodo={setSelectedTodo}
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,19 +37,28 @@ export const TodoFilter = () => (
         type="text"
         className="input"
         placeholder="Search..."
+        value={query}
+        onChange={(event) => {
```
TarasHoliuk, line 41: in such cases better to create a separate function instead of passing an anonymous function
for example: `handleInputChange`. the same for select and clear button

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,83 @@
...
+
+  const fetchData = async () => {
+    try {
+      const currentUser = await getUser(selectedTodo?.userId || 0);
+
+      setUser(currentUser);
+    } catch (error) {
+      if (error instanceof Error) {
+        // eslint-disable-next-line no-console
+        console.warn(error.message);
```
TarasHoliuk, line 23: you may handle the error in a similar way as in App.tsx - create `isError` state and show some error message in case it's `true`

TarasHoliuk (changes requested):
GJ :+1: 
I've left a few comments (most of them about naming)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,83 @@
...
-              className="modal-card-title has-text-weight-medium"
-              data-cy="modal-header"
-            >
-              Todo #2
-            </div>
+      {!user
+        ? (
+          <Loader />
+        )
+        : (
```
TarasHoliuk, line 39:
```suggestion
      {!user ? (
        <Loader />
      ) : (
```
looks better if format it in this way (just FYI)

TarasHoliuk (approved):
Well done :+1: - Approved :heavy_check_mark:

demetrxx (approved):

PR #1614, opened 2023-05-30
Solution

NadieinOleh:
https://NadieinOleh.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -17,18 +60,39 @@ export const App: React.FC = () => {
...
+                selectOption={selectOption}
+                setSelectOption={setSelectOption}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0
+                ? <Loader />
```
demetrxx, line 73: It's better practice to create a separate state to track wheter the data isLoading and show spinner only if that state is true. Because the request may return with empty data, and that case should be handled differently

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,91 @@
...
+            <p className="has-text-success">quo adipisci enim quam ut ab</p>
+          </td>
+          <td className="has-text-right is-vcentered">
+            <button data-cy="selectButton" className="button" type="button">
+              <span className="icon">
+                <i className="far fa-eye" />
+              </span>
+            </button>
+          </td>
+        </tr> */}
```
demetrxx, line 87: remove comments

src/App.tsx
```diff
@@ -8,7 +8,50 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+  const filterTodo = () => {
+    const notCompleted = todos.filter(todo => !todo.completed);
+    const completed = todos.filter(todo => todo.completed);
+
+    switch (selectOption) {
+      case 'All':
+        return !inputValue
+          ? todos
+          : todos?.filter((todo: Todo) => todo.title.includes(inputValue.toLocaleLowerCase().trim()));
```
demetrxx, line 30: Try no to repeat yourself and create helper fuctions to reuse logic

src/App.tsx
```diff
@@ -8,7 +8,50 @@ import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [currentTodo, setCurrentTodo] = useState<number>(0);
+  const [hideTodoModal, setHideTodoModal] = useState(false);
```
demetrxx, line 17: replace "hide" with "show" to convey sense properly

demetrxx (approved):
Well done 👍 Go ahead!
Take note of the comments though.

PR #1630, opened 2023-06-04
Develop

Vaskovics:
<a href="https://vaskovics.github.io/react_dynamic-list-of-todos/">Demo</a>

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,21 +12,18 @@ import { getTodos } from './api';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [selectedTodo, setSelectedTodo] = useState<Todo | undefined>();
-  const [isSelected, setIsSelected] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>();
 
   useEffect(() => {
     getTodos().then(data => setTodos(data));
-  }, [todos]);
+  }, []);
 
   const handleSelectBtn = (todo: Todo) => {
     setSelectedTodo(todo);
-    setIsSelected(true);
   };
 
   const handleCross = () => {
     setSelectedTodo(undefined);
-    setIsSelected(false);
   };
 
   return (
@@ -54,13 +51,12 @@ export const App: React.FC = () => {
         </div>
       </div>
 
-      {isSelected && (
+      {selectedTodo && (
         <TodoModal
           list={todos}
           onCross={handleCross}
         />
       )}
-
     </>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,28 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | undefined>();
+  const [isSelected, setIsSelected] = useState(false);
+
+  useEffect(() => {
+    getTodos().then(data => setTodos(data));
+  }, [todos]);
+
+  const handleSelectBtn = (todo: Todo) => {
+    setSelectedTodo(todo);
+    setIsSelected(true);
+  };
+
+  const handleCross = () => {
+    setSelectedTodo(undefined);
+    setIsSelected(false);
+  };
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
+
+  useEffect(() => {
+    getTodos().then(data => setTodos(data));
+  }, []);
+
+  const handleSelectBtn = (todo: Todo) => {
+    setSelectedTodo(todo);
+  };
+
+  const handleCross = () => {
```
TarasHoliuk, line 28:
```suggestion
  const handleClickCross = () => {
```

src/App.tsx
```diff
@@ -17,18 +49,38 @@ export const App: React.FC = () => {
...
+                onInput={handleInputFilter}
+                query={query}
+                onCross={handleClearInputBtn}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0 ? (
```
TarasHoliuk, line 61: the wrong expression was used here
what if the server todos array is empty? then you will always show `Loader`. you should show the loader only during loading todos
to implement it - create a new state `isLoading` and change it accordingly

src/App.tsx
```diff
@@ -17,18 +49,38 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                onSelect={handleSelectFilter}
+                onInput={handleInputFilter}
+                query={query}
+                onCross={handleClearInputBtn}
```
TarasHoliuk, line 56:
```suggestion
                onClickClearButton={handleClickClearButton}
```
`onCross` doesn't explainable enough

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,56 @@
...
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="clearSearchButton"
+            type="button"
+            className="delete"
+            onClick={() => onCross(query)}
```
TarasHoliuk, line 50:
```suggestion
            onClick={onCross}
```
the `onCross` (should be renamed) doesn't accept any argument, why do you pass query there

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,56 @@
...
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="clearSearchButton"
```
TarasHoliuk, line 47: according to the task requirements, this button should be displayed only if smth was typed in the input

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,56 @@
...
+      </p>
+
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          placeholder="Search..."
+          value={query}
+          onChange={(e) => onInput(e.target.value)}
```
TarasHoliuk, line 38: do not pass an anonymous functuon
create `handleChangeInput` function above instead

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,56 @@
...
+  onInput,
+  query,
+  onCross,
+}) => {
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            onChange={(e) => onSelect(e.target.value as unknown as SortBy)}
```
TarasHoliuk, line 21: create a separate function above

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,96 @@
...
-              <i className="far fa-eye-slash" />
-            </span>
-          </button>
-        </td>
-      </tr>
+export const TodoList: React.FC<ListOfTodos> = ({
+  listOfTodos,
+  onSelect,
+  sortBy,
+  query,
```
TarasHoliuk, line 17: do not pass such values here
`TodoList` component shouldn't care about `query` and `sortBy` values (single responsibility principle). Pass the prepared array of todos (which should be visible here), i.e. filter the todos array in the App.tsx

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -75,10 +96,10 @@ export const App: React.FC = () => {
 
       {selectedTodo && (
         <TodoModal
-          list={todos}
-          onCross={handleCross}
+          onCross={handleClickCross}
```
demetrxx, line 99: Rename "onCross" prop to "onClickClearButton", as was mentioned by the Taras

demetrxx (changes requested):
Good job! But let's improve your code a lil' bit.

loralevitska (changes requested):
Almost perfect) Let's improve

- add loading icon instead of information and after it downloaded change to text, also pay attention for modal component, make it fix size for avoid jumping content
https://github.com/mate-academy/react_dynamic-list-of-todos/assets/70009205/bb5a2399-89e3-4e9b-8853-db5fd7e00da0
- fix search with `completed todo` filter 
image

loralevitska (approved):
Well done)

PR #1631, opened 2023-06-04
Solution

SuaBy514:
[DEMO LINK](https://suaby514.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, {
+  useCallback, useEffect, useMemo, useState,
+} from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+  const [query, setQuery] = useState('');
+  const [filterType, setFilterType] = useState('all');
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos);
+  }, []);
+
+  const filteredTodos = useMemo(() => {
+    let currentTodos = todos;
```
d-wasilewski, line 29: you don't need to reassign `todos` to a new variable. You only use `filter` on this that don't modify this array. Operate on `todos` when you filter, and create new variable to assign the result to it

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+    if (filterType !== 'all') {
+      currentTodos = currentTodos.filter((todo) => {
+        const { completed } = todo;
+
+        if ((filterType === 'completed' && completed)
+          || (filterType === 'active' && !completed)) {
+          return true;
+        }
+
+        return false;
```
d-wasilewski, line 41: I believe it should do the same
```suggestion
        return ((filterType === 'completed' && completed)
          || (filterType === 'active' && !completed)) 
```

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+
+  const handleSelectTodo = (todo: Todo) => {
+    setCurrentTodo(todo);
+    getUser(todo.userId)
+      .then(setCurrentUser);
+  };
+
+  const handleCloseModal = useCallback(() => {
+    setCurrentTodo(undefined);
+    setCurrentUser(undefined);
```
d-wasilewski, line 60: You should always set it to `null` instead of `undefined` if you set it manually

src/App.tsx
```diff
@@ -17,18 +68,33 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                filterType={filterType}
+                query={query}
+                setFilterType={setFilterType}
+                setQuery={setQuery}
```
d-wasilewski, line 75: Read the article from checklist about naming props and functions and apply the changes in all places

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos, getUser } from './api';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [currentTodo, setCurrentTodo] = useState<Todo | undefined>();
+  const [currentUser, setCurrentUser] = useState<User | undefined>();
```
d-wasilewski, line 19: You should fetch this user in `TodoModal`

d-wasilewski (changes requested):

src/App.tsx
SuaBy514: I do not think I understand what you are saying; you say not to reassign to a new variable, then you say to create a new variable to assign the result to it... when it's already exactly what I'm doing. The variable exists because there are 2 separate filters below, filter by completion status and by query, as it would be difficult to filter it using a single filter expression.

src/App.tsx
SuaBy514: I used undefined to take care of errors saying they cannot assign "Todo | null | undefined" to "Todo | null", it was the simpler solution; I will attempt to mitigate

SuaBy514: image
I mean this error right here

src/App.tsx
SuaBy514: I really want to do this. However, with this approach, I get nothing but errors.
image

Not to mention the tests that did pass saw a hefty increase in execution time, I am not entirely sure why.

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #1632, opened 2023-06-05
Dynamic todos list (react app) - solution

ctgdprzyb:
[DEMO LINK](https://ctgdprzyb.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -34,23 +34,15 @@ export const TodoList: React.FC<Props>
       switch (filteringMode) {
         case FilteringMode.Active:
           filteredTodos = todosArg?.filter(todo => !todo.completed);
-          console.log('filtering undone!');
           break;
         case FilteringMode.Completed:
           filteredTodos = todosArg?.filter(todo => todo.completed);
-          console.log('filtering done!');
           break;
         case FilteringMode.All:
           filteredTodos = todosArg;
-          console.log('not filtering!');
           break;
         default:
       }
-
-      console.log(filteringMode);
-      console.log(todos);
-      console.log(filteredTodos);
-      console.log(typeof filteringMode);
 
       return filteredTodos.filter(todo => todo.title.toLocaleLowerCase()
         .includes(searchQuery.toLocaleLowerCase()));
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -16,7 +16,7 @@ export const TodoFilter: React.FC<Props>
             <select
               data-cy="statusSelect"
               onChange={change => setFilteringMode(
-                Number(change.target.value) as unknown as FilteringMode,
+                change.target.value as FilteringMode,
               )}
             >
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -1,14 +1,19 @@
...
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState<boolean>(true);
```
d-wasilewski, line 12: Types from simple types like strings and booleans can be infered
```suggestion
  const [isLoading, setIsLoading] = useState(true);
```

src/App.tsx
```diff
@@ -17,18 +22,34 @@ export const App: React.FC = () => {
...
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              <TodoList
+                setIsLoading={setIsLoading}
+                setInspectedTodo={setInspectedTodo}
+                filteringMode={filteringMode}
+                searchQuery={searchQuery}
+                inspectedTodo={inspectedTodo}
+              />
```
d-wasilewski, line 40: You can use ternary operator and display either loader or todo list

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,110 @@
...
-    </tbody>
-  </table>
-);
+      if (todosArg === null) {
+        return null;
+      }
+
+      switch (filteringMode) {
+        case 'active':
+          filteredTodos = todosArg?.filter(todo => !todo.completed);
```
d-wasilewski, line 35:
```suggestion
          filteredTodos = todosArg.filter(todo => !todo.completed);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,110 @@
...
-      </tr>
-    </tbody>
-  </table>
-);
+      if (todosArg === null) {
+        return null;
+      }
+
+      switch (filteringMode) {
+        case 'active':
```
d-wasilewski, line 34: You can add an enum for it so that you won't have to use strings in the code
```suggestion
        case FilterMode.ACTIVE:
```

d-wasilewski (changes requested):

src/App.tsx
ctgdprzyb: this breaks the data fetching, as TodoList never gets mounted

I've instead hidden the table header in TodoList when _todos_ is empty, giving a similar visual effect while still mounting the component necessary for fetching data

src/components/TodoList/TodoList.tsx
ctgdprzyb: I specifically added a check for null to get rid of the ?s, but forgot to remove them afterwards 😕
fixed

d-wasilewski (approved):

PR #1633, opened 2023-06-07
initial commit

ihorivna007:
[DEMO LINK](https://ihorivna007.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -17,18 +56,40 @@ export const App: React.FC = () => {
...
+                onQuery={onQuery}
+                query={query}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0
+                ? <Loader />
```
demetrxx, line 68: It's better practice to create a separate state to track whether the data isLoading and show spinner only if that state is true. Because the request may return with empty data, and that case should be handled differently

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +41,32 @@ export const TodoModal: React.FC = () => {
...
-              <strong className="has-text-danger">Planned</strong>
+              {completed
+                ? <strong className="has-text-success">Done</strong>
+                : <strong className="has-text-danger">Planned</strong>}
 
               {' by '}
 
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
+              <a href={user.email}>
```
demetrxx, line 68: you forgot mailto: part

demetrxx (approved):
Well done 👍 Go ahead!
Take note of the comments though.

PR #1635, opened 2023-06-12
add dynamic-list-of-todos

DidenkoDenis306:
[DEMO LINK](https://DidenkoDenis306.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,5 +1,7 @@
 import React from 'react';
 
+import { capitalize } from '../../helpers/capitalize';
+import { TodoStatus } from '../../types/TodoStatus';
 import { FilterBy } from '../../types/FilterBy';
 
 interface Props {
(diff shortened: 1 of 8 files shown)


And678 (changes requested):
Good job, but when pressing enter on your input, the page needlessly reloads. Try using `event.preventDefault()` on form submit

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,74 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
dffUqp, line 62: let's fix this eslint issue:smiley:

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,91 @@
...
+                    data-cy="selectButton"
+                    className="button"
+                    type="button"
+                    onClick={() => openTodo(todo)}
+                  >
+                    <span className="icon">
+                      <i className="far fa-eye" />
+                    </span>
+                  </button>
+                )}
```
dffUqp, line 84: I think here we change only icon, so let's replace this huge part of code with [classnames](https://www.npmjs.com/package/classnames) library

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

33Orange (approved):
Well done

PR #1636, opened 2023-06-12
Add React Dynamic List of Todos

ruslanpashkov:
[DEMO LINK](https://ruslanpashkov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -33,7 +33,7 @@ export const App: React.FC = () => {
 
   const selectTodo = useCallback((todo: Todo) => {
     setSelectedTodo(todo);
-  }, []);
+  }, [selectedTodo]);
 
   const clearSelectedTodo = useCallback(() => {
     setSelectedTodo(null);
@@ -41,11 +41,11 @@ export const App: React.FC = () => {
 
   const filterQuery = useCallback((value: string) => {
     setQuery(value);
-  }, []);
+  }, [query]);
 
   const selectStatus = useCallback((value: string) => {
     setStatus(value);
-  }, []);
+  }, [status]);
 
   const visibleTodos = useMemo(() => {
     return todos.filter((todo) => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-/* eslint-disable max-len */
 import React, {
-  useCallback, useEffect, useMemo, useState,
+  useState,
+  useEffect,
+  useCallback,
+  useMemo,
 } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,4 @@
-import React, {
-  useState,
-  useEffect,
-  useCallback,
-  useMemo,
-} from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -13,6 +8,7 @@ import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 import { Todo } from './types/Todo';
+import { TodoStatus } from './types/TodoStatus';
 
 import { normalize } from './helpers/normalize';
 
@@ -22,7 +18,7 @@ export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
   const [query, setQuery] = useState('');
-  const [status, setStatus] = useState('all');
+  const [status, setStatus] = useState<TodoStatus>(TodoStatus.All);
   const [isLoaded, setIsLoaded] = useState(false);
 
   useEffect(() => {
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,89 @@
...
-        </td>
-      </tr>
+export const TodoList: React.FC<Props> = React.memo(({
+  todos,
+  selectedTodo,
+  onSelect,
+}) => {
+  const isTodoSelected = useCallback((todoId: number) => (
+    todoId === selectedTodo?.id
+  ), [selectedTodo, todos]);
```
dffUqp, line 18: i think this useCallback unnecessary🤔

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,73 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={status}
+            onChange={handleStatusSelectOnChange}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
dffUqp, line 41: Can we create a helper const or enum with our statuses and then just map it here? In the future, if we decide to add one more status, it will be easier.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,73 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {query && (
+            /* eslint-disable-next-line jsx-a11y/control-has-associated-label */
```
dffUqp, line 61: let's fix this eslint issue😃

dffUqp (changes requested):
you wrote a excellent code but a lot of `useCallback`🤯

`useCallback` and `useMemo` are also functions with some logic, so we need to think twice before we start using them. Try to refactor your code and leave only important `useCallback`

src/App.tsx
```diff
@@ -8,27 +12,105 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+import { normalize } from './helpers/normalize';
+
+import { getTodos } from './api';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [status, setStatus] = useState('all');
```
dffUqp, line 25: let's use `TodoStatus` enum 


```suggestion
  const [status, setStatus] = useState(TodoStatus.all);
```

src/App.tsx
```diff
@@ -8,27 +12,105 @@ import { TodoFilter } from './components/TodoFilter';
...
+    return todos.filter((todo) => {
+      const normalizedQuery = normalize(query);
+      const normalizedTodo = normalize(todo.title);
+
+      const isTodoIncluded = normalizedTodo.includes(normalizedQuery);
+
+      let isStatusMatch: boolean;
+
+      switch (status) {
+        case 'active':
```
dffUqp, line 62: here we also can use `TodoStatus` enum

src/components/TodoFilter/TodoFilter.tsx
dffUqp: try to give `aria-label` to button, i think it can help

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,84 @@
...
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={handleClearButtonOnClick}
+            />
```
dffUqp, line 78:
```suggestion
            <button
              data-cy="clearSearchButton"
              type="button"
              className="delete"
              aria-label="clear"
              onClick={handleClearButtonOnClick}
            />
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,87 @@
...
+          <tr
+            key={todo.id}
+            data-cy="todo"
+            className={cn({
+              'has-background-info-light': isTodoSelected(todo.id),
+            })}
+          >
+            <td className="is-vcentered">
+              {todo.id}
+            </td>
```
dffUqp, line 44: Can we create separate component for tobo?🤔

In this case, we'll clean our code and remove multiple `isTodoSelected` calls .

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,87 @@
...
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              )}
+            </td>
+
+            <td className="is-vcentered is-expanded">
+              <p className={todo.completed
+                ? 'has-text-success'
+                : 'has-text-danger'}
```
dffUqp, line 57: [classnames](https://www.npmjs.com/package/classnames )

src/App.tsx
```diff
@@ -8,27 +12,105 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+  useEffect(() => {
+    getTodos()
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+        setIsLoaded(true);
+      });
+  }, []);
+
+  const selectTodo = useCallback((todo: Todo) => {
```
dffUqp, line 36: probably I think we can remove all `useCallback` and `useMemo` in this application.

look if we change input or filter we re-render `TodoFilter` and `TodoList`. In case we open `TodoModal` we won't render TodoFilter but this component doesn't have hard logic so don't worry.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,84 @@
...
+  onFilter: (value: string) => void;
+  onSelectStatus: (value: string) => void;
+}
+
+export const TodoFilter: React.FC<Props> = React.memo(({
+  query,
+  status,
+  onFilter,
+  onSelectStatus,
+}) => {
```
dffUqp, line 17: in which case this memo can help us?🤔

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

dffUqp (approved):
:fire:

maksym-bobryk (commented):
Search works with whitespaces.
image

PR #1653, opened 2023-06-28
Solution

Shevtsooov:
[DEMO LINK](https://Shevtsooov.github.io/react_dynamic-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,58 @@ export const TodoList: React.FC = () => (
...
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          data-cy="todo"
+          className=""
```
cephy77, line 34: When todo is selected it has className="has-background-info-light"

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,74 @@
...
+
+  const loadUser = useCallback(async () => {
+    const user = await getUser(selectedTodo.userId);
+
+    setTodoOwner(user);
+  }, []);
+
+  useEffect(() => {
+    loadUser();
+  }, []);
```
cephy77, line 26: In my opinion, you can write fetching logic inside useEffect, so you can get rid of useCallback:
```suggestion
  useEffect(() => {
    getUser(selectedTodo.userId)
      .then (user => setTodoOwner(user))
  }, []);
```

cephy77 (changes requested):
Looks great! Here are the small suggestions

Shevtsooov:
@cephy77 Thanks for the review! The changes have been applied.

src/App.tsx
```diff
@@ -8,7 +8,69 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+  const handleCloseModal = useCallback(() => {
+    setSelectedTodo(null);
+  }, []);
+
+  const handleSelected = (todo: Todo | null) => {
+    setSelectedTodo(todo);
+  };
+
+  const handleQuery = useCallback((search: string) => {
```
dffUqp, line 47: `useCallback` and `useMemo` are also functions with some logic, so we need to think twice before we start using them. Try to refactor your code and leave only important `useCallback`




`useCallback` we use to prevent rendering for memoized component because we don't create a new link on this function

 `useMemo` we use to prevent rendering for memoized component because we don't create a new link on array or object, and the second reason is to prevent some hard operation of array or object


memoized component - a component wrapped with `React.memo`

dffUqp (changes requested):
LGTM:fire: but needs a few fixes:upside_down_face:

Shevtsooov:
@dffUqp Thank you for the detailed review. The issues have been fixed.
Hope I handled them correctly.

src/App.tsx
```diff
@@ -17,18 +80,34 @@ export const App: React.FC = () => {
...
+                handleQuery={handleQuery}
+                handleStatus={handleStatus}
+                handleEraseInput={handleEraseInput}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {filteredTodos.length
```
33Orange, line 92: Let's add some new boolean `useState` like `isLoading` and use it here.
Initial this `isLoading` will be true and after you get response from the server you will change it to false

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,60 @@ export const TodoList: React.FC = () => (
...
+                  onClick={() => handleSelected(todo)}
+                >
+                  <span className="icon">
+                    <i className="far fa-eye" />
+                  </span>
+                </button>
+              )}
+          </td>
+        </tr>
+      ))}
```
33Orange, line 84: Let's create a component for this items
```suggestion
      {todos.map(todo => (
        <TodoItem 
          ...some props
        />
      ))}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,69 @@
...
+export const TodoModal: React.FC<Props> = ({
+  selectedTodo,
+  handleCloseModal,
+}) => {
+  const [todoOwner, setTodoOwner] = useState<User | null>(null);
+
+  useEffect(() => {
+    getUser(selectedTodo.userId)
+      .then(user => setTodoOwner(user));
+  }, []);
```
33Orange, line 21: Create a boolean flag for `isLoading` same as in the app

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,69 @@
...
-      ) : (
-        <div className="modal-card">
-          <header className="modal-card-head">
-            <div
-              className="modal-card-title has-text-weight-medium"
-              data-cy="modal-header"
-            >
-              Todo #2
-            </div>
+      {!todoOwner
```
33Orange, line 27:
```suggestion
      {isLoading
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,69 @@
...
+        ? (
+          <Loader />
+        )
+        : (
```
33Orange, line 31:
```suggestion
        ? <Loader />
        : (
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,69 @@
...
+          <div className="modal-card">
+            <header className="modal-card-head">
+              <div
+                className="modal-card-title has-text-weight-medium"
+                data-cy="modal-header"
+              >
+                {`Todo #${selectedTodo.id}`}
+              </div>
+
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
33Orange, line 41: Add an aria-label to the button and remove this rule disabler

33Orange (changes requested):
Almost done

A few moments to fix and improve

Shevtsooov:
@33Orange Hello, I really appreciate the detailed review with suggestions. All the suggestion were applied.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,76 @@
...
+    setQuery(search);
+  };
+
+  const handleEraseInput = () => {
+    setQuery('');
+  };
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+  };
```
dffUqp, line 29:
```suggestion
  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
    event.preventDefault();
  };
```

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,69 @@
...
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => handleSelected(todo)}
+            >
+              <span className="icon">
+                <i className="far fa-eye" />
+              </span>
+            </button>
+          )}
```
dffUqp, line 65: I think here we change only icon, so we can replace this huge part of code with [classnames](https://www.npmjs.com/package/classnames) library

src/components/TodoInfo/TodoInfo.tsx
```diff
@@ -0,0 +1,69 @@
...
+export const TodoInfo: React.FC<Props> = ({
+  todo,
+  selectedTodo,
+  handleSelected,
+}) => {
+  return (
+    <tr
+      data-cy="todo"
+      className={cn({
+        'has-background-info-light': selectedTodo?.id === todo.id,
```
dffUqp, line 19: can we create a variable for `selectedTodo?.id === todo.id`?

src/helpers/FilterTodos.ts
```diff
@@ -0,0 +1,24 @@
+import { FilterTodos } from '../types/FilterTodos';
+import { Todo } from '../types/Todo';
+
+export const filter = (
```
dffUqp, line 4: use more descriptive name

dffUqp (approved):
GJ🚀

PR #1657, opened 2023-06-29
Develop

okosohor:
[DEMO LINK](https://okosohor.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,49 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import './app.scss';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [todoStatus, setTodoStatus] = useState('all');
+  const [todoModalId, setTodoModalId] = useState<null | number>(null);
+  const [selectedTodo, setSelectedTodo] = useState<null | Todo>(null);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => (setIsLoading(false)));
+  }, []);
+
+  let visibleTodos = todos;
+
+  switch (todoStatus) {
+    case 'active':
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+
+    case 'completed':
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+
+    default:
+      break;
+  }
+
+  if (searchQuery) {
+    visibleTodos = visibleTodos.filter(todo => todo.title.toLowerCase().includes(searchQuery.toLowerCase()));
+  }
+
   return (
     <>
       <div className="section">
@@ -16,19 +51,39 @@ export const App: React.FC = () => {
           <div className="box">
             <h1 className="title">Todos:</h1>
 
-            <div className="block">
-              <TodoFilter />
+            <div
+              className="block"
+            >
+              <TodoFilter
+                searchQuery={searchQuery}
+                setSearchQuery={setSearchQuery}
+                todoStatus={todoStatus}
+                setTodoStatus={setTodoStatus}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              {!isLoading && (
+                <TodoList
+                  todos={visibleTodos}
+                  setTodoModalId={setTodoModalId}
+                  setSelectedTodo={setSelectedTodo}
+                  todoModalId={todoModalId}
+                />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {todoModalId && (
+        <TodoModal
+          userId={todoModalId}
+          selectedTodo={selectedTodo}
+          setTodoModalId={setTodoModalId}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,8 +1,24 @@
-export const TodoFilter = () => (
+interface Props {
+  searchQuery: string,
+  setSearchQuery: React.Dispatch<React.SetStateAction<string>>
+  todoStatus: string,
+  setTodoStatus: React.Dispatch<React.SetStateAction<string>>,
+}
+
+export const TodoFilter: React.FC<Props> = ({
+  searchQuery,
+  setSearchQuery,
+  todoStatus,
+  setTodoStatus,
+}) => (
   <form className="field has-addons">
     <p className="control">
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
+          data-cy="statusSelect"
+          value={todoStatus}
+          onChange={event => setTodoStatus(event.target.value)}
+        >
           <option value="all">All</option>
           <option value="active">Active</option>
           <option value="completed">Completed</option>
@@ -14,21 +30,29 @@ export const TodoFilter = () => (
       <input
         data-cy="searchInput"
         type="text"
+        value={searchQuery}
         className="input"
         placeholder="Search..."
+        onChange={(event) => setSearchQuery(event.target.value)}
       />
       <span className="icon is-left">
         <i className="fas fa-magnifying-glass" />
       </span>
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
-        />
-      </span>
+      {searchQuery && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="clearSearchButton"
+            type="button"
+            className="delete"
+            onClick={() => {
+              setTodoStatus('all');
+              setSearchQuery('');
+            }}
+          />
+        </span>
+      )}
     </p>
   </form>
 );
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import cn from 'classnames';
 import { Todo } from '../../types/Todo';
-// import { User } from '../../types/User';
 
 interface Props {
   todos: Todo[]
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,6 +1,21 @@
 import React from 'react';
+import cn from 'classnames';
+import { Todo } from '../../types/Todo';
+// import { User } from '../../types/User';
 
-export const TodoList: React.FC = () => (
+interface Props {
+  todos: Todo[]
+  setTodoModalId: (id: number) => void;
+  setSelectedTodo: (todo: Todo) => void;
+  todoModalId: null | number;
+}
+
+export const TodoList: React.FC<Props> = ({
+  todos,
+  setTodoModalId,
+  setSelectedTodo,
+  todoModalId,
+}) => (
   <table className="table is-narrow is-fullwidth">
     <thead>
       <tr>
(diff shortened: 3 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,46 @@ export const TodoList: React.FC = () => (
...
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => {
+                setSelectedTodo(todo);
+              }}
+            >
+              <span className="icon">
+                <i className={`far ${todo.id === selectedTodo?.id ? 'fa-eye-slash' : 'fa-eye'}`} />
```
julshvets, line 65: I think it is better to use classNames here.

src/helpers.tsx
```diff
@@ -0,0 +1,28 @@
...
+      visibleTodos = visibleTodos.filter(todo => !todo.completed);
+      break;
+
+    case 'completed':
+      visibleTodos = visibleTodos.filter(todo => todo.completed);
+      break;
+
+    default:
+      break;
+  }
```
julshvets, line 19: I think if condition should be here (if (status) {switch...} and case for the status 'all'.

julshvets (changes requested):
Good job! 
But I have some suggestions.

In the app, when you select a todo, the show button (with the eye) of only that todo should have a slash, but in your app,  show buttons of all todos where user is loaded user have slash.

And delete button in search form should only clear the search form.

src/helpers.tsx
okosohor: If status === 'all' we don`t need to do any filtering, and i think if we add case for `all`, it`ll  be redundant

src/components/TodoList/TodoList.tsx
okosohor: image
 i think case without classnames looks better

src/App.tsx
```diff
@@ -1,34 +1,77 @@
...
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              {!isLoading && (
+                <TodoList
+                  todos={visibleTodos}
+                  setSelectedTodo={setSelectedTodo}
+                  selectedTodo={selectedTodo}
+                />
+              )}
```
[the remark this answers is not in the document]
okosohor: agree with you

src/helpers.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  if (searchQuery) {
+    visibleTodos = visibleTodos
+      .filter(todo => todo.title
+        .toLowerCase().includes(searchQuery.toLowerCase()));
+  }
+
+  return visibleTodos;
+};
```
dffUqp, line 28: let's use one filter instead

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -14,21 +30,28 @@ export const TodoFilter = () => (
...
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
-        />
-      </span>
+      {searchQuery && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
dffUqp, line 44: let's fix this eslint issue:smiley:
try to give `aria-label` to button, i think it can help

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +35,35 @@ export const TodoModal: React.FC = () => {
               className="modal-card-title has-text-weight-medium"
               data-cy="modal-header"
             >
-              Todo #2
+              {`Todo #${selectedTodo?.id}`}
             </div>
 
             {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
dffUqp, line 41: same about eslint

src/App.tsx
```diff
@@ -1,34 +1,77 @@
...
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { getFilteretTodos } from './helpers';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [todoStatus, setTodoStatus] = useState('all');
```
dffUqp, line 19: can we create a const or enum with our statuses and use it in all places in our application when we need status?

src/App.tsx
```diff
@@ -1,34 +1,77 @@
...
+  const [selectedTodo, setSelectedTodo] = useState<null | Todo>(null);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => (setIsLoading(false)));
```
dffUqp, line 25: let's handle error for this request

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,32 @@
...
+export const TodoModal: React.FC<Props> = ({
+  userId,
+  selectedTodo,
+  clearSelectedTodo,
+}) => {
+  const [user, setUser] = useState<null | User >(null);
+
+  useEffect(() => {
+    getUser(userId)
+      .then(setUser);
```
dffUqp, line 22: let's handle error for this request

src/App.tsx
```diff
@@ -1,34 +1,77 @@
...
-              <Loader />
-              <TodoList />
+              {isLoading ? (<Loader />
+              ) : (
+                <TodoList
+                  todos={visibleTodos}
+                  setSelectedTodo={setSelectedTodo}
+                  selectedTodo={selectedTodo}
+                />
+              )}
```
dffUqp, line 62:
```suggestion
        {isLoading ? (
          <Loader />
        ) : (
          <TodoList
            todos={visibleTodos}
            setSelectedTodo={setSelectedTodo}
            selectedTodo={selectedTodo}
          />
        )}
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,24 @@
...
+  setTodoStatus,
+}) => (
   <form className="field has-addons">
     <p className="control">
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
+          data-cy="statusSelect"
+          value={todoStatus}
+          onChange={event => setTodoStatus(event.target.value)}
```
dffUqp, line 20: create separate handler for this

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -14,21 +30,28 @@ export const TodoFilter = () => (
       <input
         data-cy="searchInput"
         type="text"
+        value={searchQuery}
         className="input"
         placeholder="Search..."
+        onChange={(event) => setSearchQuery(event.target.value)}
```
dffUqp, line 36: create separate handler for this

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -14,21 +30,28 @@ export const TodoFilter = () => (
...
+      {searchQuery && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="clearSearchButton"
+            type="button"
+            className="delete"
+            onClick={() => {
+              setSearchQuery('');
+            }}
```
dffUqp, line 51: create separate handler for this

dffUqp (changes requested):
LGTM:fire: but needs a few fixes:upside_down_face:

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,78 @@
...
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+interface Props {
+  searchQuery: string,
+  setSearchQuery: React.Dispatch<React.SetStateAction<string>>
+  todoStatus: string,
```
roman-mirzoian, line 6: In the file `src/types/TodoStatus.ts`, you have already defined how the type should be in `todoStatus`, so you should use this type for more accurate typing, and not specify the type of the `string`.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  selectedTodo,
+  clearSelectedTodo,
+}) => {
+  const [user, setUser] = useState<null | User >(null);
+
+  useEffect(() => {
+    getUser(userId)
+      .then(setUser)
+      .catch(error => (new Error(error.message)));
+  }, [user]);
```
roman-mirzoian, line 24: Since you set the value of the `user` in the middle of the `useEffect` hook via `setUser`, the reference to the `user` object as a dependency for `useEffect` is not correct, and in this case it would be more correct to refer to the `userId` or not to transfer anything to the dependency array at all, because we need to get the data for a component only once.

src/App.tsx
```diff
@@ -1,34 +1,80 @@
...
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { getFilteretTodos } from './helpers';
+import { TodoStatus } from './types/TodoStatus';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [todoStatus, setTodoStatus] = useState(`${TodoStatus.all}`);
```
roman-mirzoian, line 20: The initial state for `todoStatus` is set to a string template literal. You can just use `TodoStatus.all` without the template literal.

src/App.tsx
```diff
@@ -1,34 +1,80 @@
...
 import '@fortawesome/fontawesome-free/css/all.css';
+import './app.scss';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { getFilteretTodos } from './helpers';
```
roman-mirzoian, line 13: Be careful when naming functions (getFiltere`t`Todos is now typo, should be getFiltere`d`Todos).

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,50 @@ export const TodoList: React.FC = () => (
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          key={todo.id}
+          data-cy="todo"
+          className=""
```
roman-mirzoian, line 35: Do we really need a `className` definition with an empty string here?

roman-mirzoian (changes requested):
You have done a good job on the previous issues and it remains to polish your work a little 👍

src/App.tsx
okosohor: if i set TodoStatus.all i have some trouble, because i need to use setTodoStatus status with event.target.value and its string. notTodoStatus
image

src/helpers.tsx
```diff
@@ -0,0 +1,31 @@
+import { Todo } from './types/Todo';
+import { TodoStatus } from './types/TodoStatus';
+
+export const getFilteredTodos = (todos: Todo[],
+  searchQuery: string,
+  status: string) => {
```
Moroz-Dmytro, line 6:
```suggestion
export const getFilteredTodos = (todos: Todo[],
  searchQuery: string,
  status: string
) => {
```

Moroz-Dmytro (changes requested):
Have you run `npm deploy`?
Now when I click on "Eye icon" I see that a lot of requests are sending
image

roman-mirzoian (changes requested):
The task is well done and needs only minor corrections.
I also noticed now that when searching for non-existent ones, the layout jumps there, so it's better to fix it:
https://gyazo.com/f61bcdbf44a03f8e1d5768897add3264

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,91 @@
...
+      case 'active':
+        setTodoStatus(TodoStatus.ACTIVE);
+        break;
+
+      default:
+        break;
+    }
+  };
+
+  const handleInputChange = (
```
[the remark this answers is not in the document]
okosohor: in handleClick we have mouse event (MouseEvent<HTMLButtonElement, MouseEvent>) in input we have React.ChangeEvent<HTMLInputElement> and when we create 1 handler for input and button, this will call errors with ts

src/App.tsx
```diff
@@ -1,34 +1,82 @@
...
-              <TodoList />
+              {isLoading ? (
+                <Loader />
+              ) : (
+                <TodoList
+                  todos={visibleTodos}
+                  setSelectedTodo={setSelectedTodo}
+                  selectedTodo={selectedTodo}
+                />
+              )}
```
naziamoff, line 67:
```suggestion
              {isLoading 
                ? <Loader />
                : (
                  <TodoList
                    todos={visibleTodos}
                    setSelectedTodo={setSelectedTodo}
                    selectedTodo={selectedTodo}
                  />
              )}
```

naziamoff (approved):
👍

PR #1663, opened 2023-06-29
add solution

growingananas:
DEMO LINK](https://growingananas.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,69 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-
+import { getTodos } from './api';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { Todo } from './types/Todo';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [selectedFilter, setSelectedFilter] = useState('all');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [loader, setLoader] = useState(true);
+  const [modalOpened, setModalOpened] = useState(false);
+
+  const getfilteredTodos = useCallback(() => {
+    const normQuery = query.toLowerCase().trim();
+    let filteredTodos = [...todos];
+
+    if (normQuery) {
+      filteredTodos = filteredTodos
+        .filter(todo => todo.title.toLowerCase().includes(normQuery));
+    }
+
+    switch (selectedFilter) {
+      case 'active':
+        return filteredTodos.filter(todo => !todo.completed);
+
+      case 'completed':
+        return filteredTodos.filter(todo => todo.completed);
+
+      default:
+        return filteredTodos;
+    }
+  }, [query, selectedFilter, todos]);
+
+  const visibleTodos = getfilteredTodos();
+
+  const getTodo = useCallback(async () => {
+    try {
+      const loadedTodo = await getTodos();
+
+      setTodos(loadedTodo);
+      setLoader(false);
+    } catch {
+      setLoader(true);
+    }
+  }, []);
+
+  useEffect(() => {
+    getTodo();
+  }, [getTodo]);
+
+  const clearDataModal = useCallback(() => {
+    setSelectedTodo(null);
+  }, []);
+
+  const openDataModal = useCallback((todo: Todo | null) => {
+    setSelectedTodo(todo);
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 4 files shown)


ivankhrsh:
Selected item should have different background color, compare with mate version
image

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +45,41 @@ export const TodoModal: React.FC = () => {
...
-              <strong className="has-text-danger">Planned</strong>
+              {selectedTodo?.completed
+                ? <strong className="has-text-success">Done</strong>
+                : <strong className="has-text-danger">Planned</strong>}
 
               {' by '}
 
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
+              <a href={user.email}>
```
ivankhrsh, line 81: You need to add 'mailto:' to email, so it's will be the link for sending email. Now on demo you have 404 error while clicking it.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,91 @@
...
-    </thead>
+type Props = {
+  selectedTodo: Todo | null
+  todos: Todo[];
+  setModalOpened: (modalOpened: boolean) => void;
+  handleDataModal: (Todo: Todo) => void;
+};
 
-    <tbody>
-      <tr data-cy="todo" className="">
```
ivankhrsh, line 19: On this line  you should have className 'has-background-info-light' while todo is selected

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+
+      case 'completed':
+        return filteredTodos.filter(todo => todo.completed);
+
+      default:
+        return filteredTodos;
+    }
+  }, [query, selectedFilter, todos]);
+
+  const visibleTodos = getfilteredTodos();
```
ivankhrsh, line 42: It's better to use useMemo here, because it's complex calculation and we want to prevent recalculation

ivankhrsh (changes requested):

ivankhrsh (approved):
GJ 🔥

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
+
+    switch (selectedFilter) {
+      case 'active':
+        return filteredTodos.filter(todo => !todo.completed);
+
+      case 'completed':
+        return filteredTodos.filter(todo => todo.completed);
+
+      default:
+        return filteredTodos;
```
dffUqp, line 42: you can use only one filter call for this logic

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,68 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
dffUqp, line 55: let's fix this eslint issue:smiley:

try to give `aria-label` to button, i think it can help

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { Todo } from './types/Todo';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [selectedFilter, setSelectedFilter] = useState('all');
```
dffUqp, line 21: can we create a const or enum with our statuses and use it in all places in our application when we need status?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,68 @@
...
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+  const handleSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const { value } = event.target;
+
+    setSelectedFilter(value);
+  };
```
dffUqp, line 22: try to write the code in the same style, I mean that you didn't use destructuring in the `handleInput` function, why are you doing it here?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,94 @@
...
+                    onClick={() => {
+                      handleDataModal(todo);
+                      setModalOpened(true);
+                    }}
+                  >
+                    <span className="icon">
+                      <i className="far fa-eye" />
+                    </span>
+                  </button>
+                )}
```
dffUqp, line 86: I think here we change only icon, so let's replace this huge part of code with [classnames](https://www.npmjs.com/package/classnames) library

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,94 @@
...
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      <tbody>
+        {todos.map(todo => (
+          <tr
```
dffUqp, line 35: we can create a separate component for todo it helps us to create a variable for `todo.id === selectedTodo?.id` and improve readability

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,94 @@
...
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+            <td className="is-vcentered is-expanded">
+              <p className={
+                todo.completed ? 'has-text-success' : 'has-text-danger'
+              }
```
dffUqp, line 54: [classnames](https://www.npmjs.com/package/classnames)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +45,41 @@ export const TodoModal: React.FC = () => {
...
           </header>
 
           <div className="modal-card-body">
             <p className="block" data-cy="modal-title">
-              quis ut nam facilis et officia qui
+              {selectedTodo?.title}
             </p>
 
+            {error && (
+              <h2 style={{ color: 'red' }}>
```
dffUqp, line 69: try to avoid using `inline-styles`

dffUqp (approved):
Looks good:smiley: but always can be better, try not to make these mistakes in the following tasks.

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [loader, setLoader] = useState(true);
+  const [modalOpened, setModalOpened] = useState(false);
```
dffUqp, line 24: We can have one state for modal. If there is selected todo, then modal should be open - if no modal should be closed
And when you close modal set selected todo to null

PR #1678, opened 2023-07-02
Develop

SofiaLiubchenko:
[DEMO LINK](https://SofiaLiubchenko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,66 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import './App.scss';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todosList, setTodosList] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filteredTodos, setFilteredTodos] = useState<string>('');
+  const [searchQuery, setSearchQuery] = useState<string>('');
+
+  useEffect(() => {
+    getTodos().then(setTodosList);
+  }, []);
+
+  let visibleTodos = [...todosList];
+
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
+        visibleTodos = todosList;
+        break;
+      case 'active':
+        visibleTodos = visibleTodos.filter(todo => todo.completed === false);
+        break;
+      case 'completed':
+        visibleTodos = visibleTodos.filter(todo => todo.completed === true);
+        break;
+      default:
+        break;
```
naziamoff, line 38: let's use enum instead. Strings are difficult to maintain

src/App.tsx
```diff
@@ -17,18 +69,20 @@ export const App: React.FC = () => {
...
+              <TodoFilter filter={handleSelect} setSearchQuery={setSearchQuery} searchQuery={searchQuery} clearSearch={clearSearch} />
             </div>
 
-            <div className="block">
-              <Loader />
-              <TodoList />
-            </div>
+            {visibleTodos.length !== 0 ? (
+              <div className="block">
+                <TodoList todos={visibleTodos} onTodoSelect={onTodoSelect} selectedTodo={selectedTodo} />
```
naziamoff, line 77: you sure it should fit into one line?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,61 @@
...
+      </p>
+
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          placeholder="Search..."
+          value={searchQuery}
+          onChange={event => setSearchQuery(event.target.value)}
```
naziamoff, line 39: let's create a handler for that

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +26,58 @@ export const TodoList: React.FC = () => (
...
+                onClick={() => onTodoSelect(todo)}
+              >
+                <span className="icon">
+                  <i className={classNames(
+                    'far',
+                    {
+                      'fa-eye': todo !== selectedTodo,
+                      'fa-eye-slash': todo === selectedTodo,
+                    },
+                  )}
```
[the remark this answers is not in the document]
naziamoff: also, one of these classnames can be set as default and just overwritten if matches other condition

naziamoff (changes requested):
not bad, but let's polish

PR #1686, opened 2023-07-04
added solution

vladyslav10shevchenko:
[DEMO LINK](https://sheva10barca.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,76 @@
 /* eslint-disable max-len */
-import React from 'react';
-import 'bulma/css/bulma.css';
-import '@fortawesome/fontawesome-free/css/all.css';
+import React, { useEffect, useState } from 'react';
+
+import { getTodos } from './api';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { TodoModal } from './components/TodoModal';
+
+import { Todo } from './types/Todo';
+import { FilterStatus } from './types/FilterStatus';
+
+import 'bulma/css/bulma.css';
+import '@fortawesome/fontawesome-free/css/all.css';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+  const [query, setQuery] = useState('');
+  const [filterStatus, setFilterStatus] = useState(FilterStatus.ALL);
+
+  useEffect(() => {
+    getTodos()
+      .then((todosFromServer) => {
+        setTodos(todosFromServer);
+        setIsLoading(false);
+      })
+      .catch((error) => {
+        throw new Error('Loading todos error: ', error.message);
+      });
+  }, []);
+
+  const visibleTodos = todos.filter((todo) => {
+    const normalizedQuery = query.toLowerCase().trim();
+
+    const normalizedTodo = todo.title
+      .toLowerCase()
+      .trim()
+      .includes(normalizedQuery);
+
+    switch (filterStatus) {
+      case FilterStatus.ALL:
+        return normalizedTodo;
+
+      case FilterStatus.ACTIVE:
+        return normalizedTodo && !todo.completed;
+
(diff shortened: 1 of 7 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,36 @@
...
+  useEffect(() => {
+    getUser(todo.userId)
+      .then((u) => {
+        setUser(u);
+        setIsLoading(false);
+      })
+      .catch((error) => {
+        throw new Error('Error fetching user: ', error.message);
+      });
+  }, []);
```
Viktor-Kost, line 28: Missing dependency in useEffect, add todo.userId to dependencies array to avoid unexpected behavior.

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+  const [filterStatus, setFilterStatus] = useState(FilterStatus.ALL);
+
+  useEffect(() => {
+    getTodos()
+      .then((todosFromServer) => {
+        setTodos(todosFromServer);
+        setIsLoading(false);
+      })
+      .catch((error) => {
+        throw new Error('Loading todos error: ', error.message);
```
Viktor-Kost, line 31: In the catch block, the error object is being thrown incorrectly. The Error constructor accepts only one argument. You cannot pass the error message directly after the custom message.

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+      });
+  }, []);
+
+  const visibleTodos = todos.filter((todo) => {
+    const normalizedQuery = query.toLowerCase().trim();
```
Viktor-Kost, line 36: You are filtering 'todos' on every render. Consider using useMemo to optimize this operation and avoid unnecessary recalculations.

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+  const [query, setQuery] = useState('');
+  const [filterStatus, setFilterStatus] = useState(FilterStatus.ALL);
+
+  useEffect(() => {
+    getTodos()
+      .then((todosFromServer) => {
+        setTodos(todosFromServer);
```
Viktor-Kost, line 27: It's a good practice to set 'isLoading' to false in the 'finally' block to ensure it's executed whether the promise is fulfilled or rejected.

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +32,51 @@ export const TodoList: React.FC = () => (
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map((todo) => (
+        <tr
+          data-cy="todo"
+          key={todo.id}
+          className={cn('', {
```
Viktor-Kost, line 39: The 'className' attribute receives an empty string ''. If this is not necessary, consider removing it for cleaner code.

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +32,51 @@ export const TodoList: React.FC = () => (
...
+            >
+              {todo.title}
+            </p>
+          </td>
+          <td className="has-text-right is-vcentered">
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={(event) => onSelectTodo(event, todo)}
```
Viktor-Kost, line 66: In onClick method, you are directly calling onSelectTodo method. It would be better to define a separate method inside your component where you can handle the event and call onSelectTodo with the required arguments.

Viktor-Kost (changes requested):

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,63 @@
...
+  };
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select data-cy="statusSelect" onChange={handleFilterStatus}>
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
etojeDenys, line 31: pass enum values as value

etojeDenys (changes requested):
nit:

VitaliyBondarenko1982 (approved):

PR #1687, opened 2023-07-05
add dynamic list of todos solution

vchervinko:
[DEMO LINK](https://vchervinko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [isCompleted, setIsCompleted] = useState('all');
```
naziamoff, line 18: the name tells me we're dealing with boolean value here

src/App.tsx
```diff
@@ -1,14 +1,56 @@
 /* eslint-disable max-len */
```
naziamoff, line 1: you should remove it from your file. You should fix linting issues instead of disabling them

src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
+    todosList.filter(todo => todo.completed === todoStatus)
+  );
+
+  if (isCompleted === 'completed') {
+    visibleTodos = filterByStatus(visibleTodos, true);
+  }
+
+  if (isCompleted === 'active') {
+    visibleTodos = filterByStatus(visibleTodos, false);
+  }
```
naziamoff, line 44: do we use `if/else` or `switch/case` with pre-defined number of limited conditions?

src/App.tsx
```diff
@@ -17,18 +59,35 @@ export const App: React.FC = () => {
...
-              <TodoList />
+              {isLoading
+                ? (
+                  <TodoList
+                    todos={visibleTodos}
+                    selectedTodo={selectedTodo}
+                    onSelect={selectTodo}
+                  />
+                )
+                : <Loader />}
```
naziamoff, line 79: if is loading - we render todolist, and if not - we render loader? Does this make sense?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,86 @@
...
-      </span>
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {searchQuery !== ''
```
naziamoff, line 71: looks like a crutch. Let's convert to boolean like that `Boolean(value)`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,19 @@
 import React from 'react';
+// import cn from 'classnames';
```
naziamoff, line 2: you should not add comments to your commits unless it is really necessary

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,75 @@
...
-      ) : (
-        <div className="modal-card">
-          <header className="modal-card-head">
-            <div
-              className="modal-card-title has-text-weight-medium"
-              data-cy="modal-header"
-            >
-              Todo #2
-            </div>
+      {!user
```
naziamoff, line 32: we usually start with positive condition implementation - and only then negative. Conditions like that are easier to read for humans

naziamoff (changes requested):

PR #1690, opened 2023-07-06
done

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,25 +9,31 @@ import { Loader } from './components/Loader';
 import { getTodos } from './api';
 import { Todo } from './types/Todo';
 import { TodoUser } from './types/User';
+import { SortTypes } from './types/SortType';
 
 export const App: React.FC = () => {
   const [loaded, setLoaded] = useState(false);
   const [sourceData, setSourceData] = useState<Todo[]>([]);
-  const [sortBy, setSortBy] = useState('all');
-  const [userModal, setUserModal] = useState<TodoUser | null>(null);
+  const [sortTypeForTodos, setSortTypeForTodos] = useState<SortTypes | string>(
+    SortTypes.ALL,
+  );
+  const [userInfoForModal, setUserInfoForModal] = useState<TodoUser | null>(
+    null,
+  );
   const [showModal, setShowModal] = useState(false);
-  const [eyeMark, setEyeMark] = useState<number>(-1);
+  const [selectButtonMark, setSelectButtonMark] = useState<number>(-1);
   const sortTodos = () => {
     if (loaded) {
-      switch (sortBy) {
-        case 'active':
+      switch (sortTypeForTodos) {
+        case SortTypes.ACTIVE:
           return sourceData.filter((todo) => !todo.completed);
-        case 'completed':
+        case SortTypes.COMPLETED:
           return sourceData.filter((todo) => todo.completed);
-        case 'all':
+        case SortTypes.ALL:
           return sourceData;
         default:
-          return sourceData.filter((todo) => todo.title.includes(sortBy));
+          return sourceData.filter((todo) => todo.title
+            .includes(sortTypeForTodos));
       }
     }
 
@@ -51,17 +57,20 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter sortType={setSortBy} sortValue={sortBy} />
+              <TodoFilter
+                sortType={setSortTypeForTodos}
+                sortValue={sortTypeForTodos}
+              />
             </div>
 
             <div className="block">
               {loaded ? (
                 <TodoList
-                  todoList={todos !== null ? todos : []}
-                  setUserModal={setUserModal}
+                  todoList={todos}
+                  setUserInfoForModal={setUserInfoForModal}
                   setShowModal={setShowModal}
-                  eyeMark={eyeMark}
-                  setEyeMark={setEyeMark}
+                  selectButtonMark={selectButtonMark}
+                  setSelectButtonMark={setSelectButtonMark}
                 />
               ) : (
                 <Loader />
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+import { TodoUser } from './types/User';
 
 export const App: React.FC = () => {
+  const [sourceData, setSourceData] = useState<Todo[] | null>(null);
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [sortBy, setSortBy] = useState('all');
+  const [userModal, setUserModal] = useState<TodoUser | null>(null);
+  const [eyeMark, setEyeMark] = useState<number>(-1);
+
+  const preRender = () => {
```
TarasHoliuk, line 20: consider renaming this function to smth like `getVisibleTodos` etc., to be more descriptive

src/App.tsx
```diff
@@ -17,18 +58,36 @@ export const App: React.FC = () => {
...
+                )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {check(userModal)
+        && (
+          <TodoModal
```
TarasHoliuk, line 85: modal should open immediately, and show loading during user fetch

src/api.ts
```diff
@@ -22,6 +22,10 @@ function get<T>(url: string): Promise<T> {
     .then(res => res.json());
 }
 
+export const check = (arg: Todo | TodoUser | null): boolean => {
```
TarasHoliuk, line 25: shouldn't be here

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+import { check, getTodos } from './api';
+import { Todo } from './types/Todo';
+import { TodoUser } from './types/User';
 
 export const App: React.FC = () => {
+  const [sourceData, setSourceData] = useState<Todo[] | null>(null);
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [sortBy, setSortBy] = useState('all');
+  const [userModal, setUserModal] = useState<TodoUser | null>(null);
+  const [eyeMark, setEyeMark] = useState<number>(-1);
```
TarasHoliuk, line 18: consider rename it to be more descriptive, e.g. `selectedTodoId`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,22 +1,73 @@
...
+  todoList, setUserModal, eyeMark, setEyeMark,
+}) => {
+  const eyeButtonHandler = (todo: Todo) => {
+    setEyeMark(todo.id);
+
+    return getUser(todo.userId)
+      .then(user => {
+        return { ...user, todo };
+      })
+      .then(mutateUser => setUserModal(mutateUser));
```
TarasHoliuk, line 23: you may save user here, in local state

src/components/TodoList/TodoList.tsx
```diff
@@ -1,22 +1,73 @@
...
+        </tr>
+      </thead>
+
+      <tbody>
+        {todoList.map(todo => (
+          <tr data-cy="todo" className="" key={todo.id}>
+            <td className="is-vcentered">{todo.id}</td>
+            <td className="is-vcentered" />
+            <td className="is-vcentered is-expanded">
+              <p className={`${todo.completed
```
TarasHoliuk, line 47: it's better for readability to use classnames library to do such things

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,64 @@
...
-      ) : (
+      { check(user.todo) ? (
         <div className="modal-card">
           <header className="modal-card-head">
             <div
               className="modal-card-title has-text-weight-medium"
               data-cy="modal-header"
             >
-              Todo #2
+              {`Todo #${user.todo.id}`}
```
TarasHoliuk, line 33: use destructuring assignment

TarasHoliuk (changes requested):

src/App.tsx
victor-buldenko: we don`t know, what we have to show modal todo or modal error, it depends on getUser, then I show them the result of the request, the user doesn't need to wait, he will see the mark when it is done.

src/components/TodoList/TodoList.tsx
victor-buldenko: I need to pass it parent, then to the modal. Save here? what for?

src/components/TodoModal/TodoModal.tsx
victor-buldenko: It's impossible because the return consists of the conditional operator. Destruct in arguments is impossible to do because need to assign Props & at the moment when the user object has come to the component null (wait for promise done)

src/components/TodoList/TodoList.tsx
victor-buldenko: for 1 class? no tnx.

src/App.tsx
OleziO: Not resolved

src/App.tsx
```diff
@@ -1,14 +1,48 @@
...
+  const [loaded, setLoaded] = useState(false);
+  const [sourceData, setSourceData] = useState<Todo[]>([]);
+  const [sortBy, setSortBy] = useState('all');
+  const [userModal, setUserModal] = useState<TodoUser | null>(null);
+  const [showModal, setShowModal] = useState(false);
+  const [eyeMark, setEyeMark] = useState<number>(-1);
+  const sortTodos = () => {
+    if (loaded) {
+      switch (sortBy) {
+        case 'active':
```
OleziO, line 23: Move all cases to enum

src/App.tsx
```diff
@@ -17,18 +51,35 @@ export const App: React.FC = () => {
...
+              ) : (
+                <Loader />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {/* {check(userModal) && ( */}
```
OleziO, line 74: Remove all comments.

OleziO (changes requested):
Fix all comments

src/App.tsx
victor-buldenko: its no make sense, human has no access to select values and never use them in code

SanyaBratashchuk (approved):

PR #1702, opened 2023-07-15
add task solution

VolodymyrPst:
https://volodymyrpst.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,52 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selected, setSelected] = useState<Todo | null>();
+  const [filterSelect, setFilterSelect] = useState('');
+  const [search, setSearch] = useState('');
+
+  const getTodosFromServer = async () => {
+    let todosArr;
+
+    try {
+      todosArr = await getTodos();
+    } catch (error) {
+      throw new Error('Error todos loading');
+    }
+
+    switch (filterSelect) {
+      case 'all':
+        break;
+      case 'active':
+        todosArr = todosArr.filter(item => item.completed);
+        break;
+      case 'completed':
+        todosArr = todosArr.filter(item => !item.completed);
+        break;
+      default:
+        break;
+    }
+
+    const str = search.toLowerCase();
+
+    todosArr = todosArr.filter((item) => {
+      const title = item.title.toLowerCase();
+
+      return title.includes(str);
+    });
+
+    setTodos(todosArr);
+  };
+
+  useEffect(() => {
+    getTodosFromServer();
+  }, [filterSelect, selected, search]);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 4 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,80 @@
...
+                  )}
+                  >
+                    {todo.title}
+                  </p>
+                </td>
+                <td className="has-text-right is-vcentered">
+                  <button
+                    data-cy="selectButton"
+                    className="button"
+                    type="button"
```
anastasiiavorobiova, line 61: The type of the button is not necessary as the type is button by default.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,80 @@
...
-  </table>
-);
+      <tbody>
+        {
+          todos.map((todo) => {
+            return (
+              <tr
+                data-cy="todo"
+                className=""
+                key={todo.id}
```
anastasiiavorobiova, line 38: The TODO items are not having any unique class for the selected items. It would be better to add a class for selected items for better UX or you can remove redundant className

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,44 +1,72 @@
...
+  selectedTodo,
+  selectMethod,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  const getUserFromServer = async () => {
+    try {
+      setUser(await getUser(selectedTodo.userId));
+    } catch (error) {
+      throw new Error('Error user loading');
```
anastasiiavorobiova, line 22: In the catch block, you are throwing a new error but there is no error handling mechanism in the component. It would be better to have a state to hold the error and display it to the user.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,44 +1,72 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+import { getUser } from '../../api';
 import { Loader } from '../Loader';
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+
+type Pros = {
+  selectedTodo: Todo,
```
anastasiiavorobiova, line 8: Typo in type name 'Pros', it should be 'Props'.

src/App.tsx
```diff
@@ -17,18 +61,31 @@ export const App: React.FC = () => {
...
+                searchWord={setSearch}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {
+                todos.length === 0
+                  ? <Loader />
```
anastasiiavorobiova, line 74: Currently the Loader is shown if there are no todos. It would be better to have a separate state to track whether loading is in progress or not.

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+  const [filterSelect, setFilterSelect] = useState('');
+  const [search, setSearch] = useState('');
+
+  const getTodosFromServer = async () => {
+    let todosArr;
+
+    try {
+      todosArr = await getTodos();
+    } catch (error) {
+      throw new Error('Error todos loading');
```
anastasiiavorobiova, line 25: In case of exception, the error is being thrown, but it is not being caught anywhere. It would be better to handle that error in a way that does not crash the entire application.

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+    }
+
+    switch (filterSelect) {
+      case 'all':
+        break;
+      case 'active':
+        todosArr = todosArr.filter(item => item.completed);
```
anastasiiavorobiova, line 32: The filter for 'active' todos is filtering for completed todos. It should be filtering for todos that are not completed (`!item.completed`).

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+        break;
+      case 'completed':
+        todosArr = todosArr.filter(item => !item.completed);
```
anastasiiavorobiova, line 35: The filter for 'completed' todos is filtering for active todos. It should be filtering for todos that are completed (`item.completed`).

anastasiiavorobiova (changes requested):

src/components/TodoList/TodoList.tsx
vlad-demchuk: not fixed

vlad-demchuk: not fixed

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,44 +1,68 @@
...
+};
+
+export const TodoModal: React.FC<Props> = ({
+  selectedTodo,
+  selectMethod,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  const getUserFromServer = async () => {
+    setUser(await getUser(selectedTodo.userId));
```
vlad-demchuk, line 19: why did you remove try/catch statement? it was just suggested to add some value to state. And then you could use this value for showing some error(title) on your page.

vlad-demchuk (changes requested):

volodymyr-soltys97 (approved):

PR #1704, opened 2023-07-17
dynamic todos

Ir-ra:
[DEMO LINK](https://Ir-ra.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,72 @@
...
-        <td className="is-vcentered">1</td>
-        <td className="is-vcentered" />
-        <td className="is-vcentered is-expanded">
-          <p className="has-text-danger">delectus aut autem</p>
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
+export const TodoList: React.FC<Props> = ({
+  todos,
+  onSelect = () => { },
```
lerastarynets, line 12: Empty function used as a default prop for `onSelect`. This could be potentially dangerous, as it might be expected to perform an action. If `onSelect` is not provided, it might be more appropriate to make it required or handle its absence correctly.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,72 @@
...
+              >
+                {todo.title}
+              </p>
+            </td>
+            <td className="has-text-right is-vcentered">
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                key={todo.id}
```
lerastarynets, line 58: The `key` prop is not needed on the `button` element and it may cause problems if multiple buttons have the same key. Keys should be on the outermost child being created in a loop, which is the `tr` element in this case.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,72 @@
...
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+          <tr
+            data-cy="todo"
+            className=""
```
lerastarynets, line 33: The `className` prop is an empty string. If no classes are added, you can safely remove this prop.

src/App.tsx
```diff
@@ -17,18 +49,32 @@ export const App: React.FC = () => {
...
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {loading && (
+                <Loader />
+              )}
+
+              {!loading && todos.length > 0 && (
```
lerastarynets, line 64: pls [see falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)
```suggestion
              {!loading && !!todos.length && (
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,31 +33,36 @@ export const TodoModal: React.FC = () => {
...
+                ? (<strong className="has-text-success">Done</strong>)
+                : (<strong className="has-text-danger">Planned</strong>)}
 
               {' by '}
 
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
-              </a>
+              {users && (
+                <a href="mailto:Sincere@april.biz">
```
lerastarynets, line 61: The email address seems to be hardcoded. Instead, it should be fetched from the user data.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

lerastarynets (approved):
You've got this, keep pushing forward!

PR #1706, opened 2023-07-19
sol

GGsmou:
[DEMO LINK](https://GGsmou.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+        default:
+          return true;
+      }
+    });
+
+    newTodos = newTodos.filter((todo) => todo.title.trim().toLowerCase()
+      .includes(filterToApply[1].trim().toLocaleLowerCase()));
+
+    return newTodos;
+  };
```
adv555, line 47: A bit hard for understanding `filterToApply[1] or filterToApply[0]`... might be better to use an `object` with filters instead of `array` of filterToApplywhat. it could be difficult in future if you need to add more filtering options
for instance ..
```
const filterTodos = (
  todos: Todo[],

  {
    filterSelect: status,
    filterTitle: query,
  }: { filterSelect: FilterSelect; filterTitle: string },

) => {
  let newTodos = [...todos];

  newTodos = newTodos.filter((todo) => {
    switch (status) {
      case FilterSelect.Active:
        return !todo.completed;

      case FilterSelect.Completed:
        return todo.completed;

      default:
        return true;
    }
  });

  if (query) {
    const normalizedQuery = query.trim().toLowerCase();

    newTodos = newTodos.filter((todo) => {
      const normalizedTodoTitle = todo.title.trim().toLowerCase();

      return normalizedTodoTitle.includes(normalizedQuery);
    });
  }

  return newTodos;
};
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,19 +37,24 @@ export const TodoFilter = () => (
...
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
-        />
-      </span>
+      {filterTitle !== '' && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
adv555, line 49: you may solve this lint issue just add arial-lable attribute `aria-label="clear search"`

The rule is triggered by the role="button" attribute.  it needs a text label for accessibility reasons (so screen readers know what to read out, for example). To comply with the rule, you can either give the "button" textual content or add an `aria-label attribute`

adv555 (approved):
Nice work!😊

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+        setError('');
+        setTodos(data);
+      })
+      .catch((err) => {
+        setError(err.message);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
```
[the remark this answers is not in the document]
GGsmou: It is not really suitable for my solution because I represent error as string not an object

src/components/TodoFilter/TodoFilter.tsx
GGsmou: It was here from beginning, so I decided not to fix it, but thanks for advice)

And678 (approved):
Good work, approved :heavy_check_mark: 
But check out @adv555's comment

And678 (approved):
2

PR #1708, opened 2023-07-19
React__dynamic-list-of-todos

Soi4An:
DEMO - https://Soi4An.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -17,6 +17,7 @@ export const App: React.FC = () => {
   const [filterByCompleted, setFilterByCompleted] = useState(FilterValues.all);
   const [filterByTitle, setFilterByTitle] = useState('');
   const [selectTodo, setSelectTodo] = useState<Todo | null>(null);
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+          const generalTitle = todo.title.toLocaleLowerCase();
+
+          return generalTitle.includes(generalQuery);
+        });
+
+        setVisibleTodos(filteredTodos);
+      })
+      .catch(() => {
+        setVisibleTodos([]);
+        throw new Error('Can\'t load Todos with server...')
```
anastasiiavorobiova, line 50: it's an error handler it shouldn't throw new errors. It's great to set an empty value for loaded todos, but further we would like to use it for setting an error message in the component state + showing this error message

anastasiiavorobiova (changes requested):
Great work! Added a tiny comment to fix

volodymyr-soltys97 (approved):
Well done 🔥

PR #1721, opened 2023-07-25
dynamic_todos

hrynkevych:
[DEMO LINK](https://hrynkevych.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,75 @@
...
+        case Filter.ACTIVE:
+          return isTodoSuitable && !todo.completed;
+
+        case Filter.COMPLETED:
+          return isTodoSuitable && todo.completed;
+
+        default:
+          throw new Error('Unknown filter');
+      }
+    });
```
IvanFesenko, line 54: No need to filter in case when Filter is All and query is empty

IvanFesenko (changes requested):
Good job. One improvement required

src/App.tsx
hrynkevych: Ok, I think I got it. Can you, please, look at my fix below

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,91 @@
...
+              >
+                {todo.title}
+              </p>
+            </td>
+            <td className="has-text-right is-vcentered">
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={(event) => handleSelectTodo(event, todo)}
```
etojeDenys, line 74: we can call `event.preventDefault();` here, and pass only the todo without event

etojeDenys (changes requested):

src/components/TodoList/TodoList.tsx
hrynkevych: ok, thank you, fixed

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,92 @@
...
+            </td>
+            <td className="has-text-right is-vcentered">
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={(event) => {
+                  event.preventDefault();
+                  handleSelectTodo(todo);
+                }}
```
IvanFesenko, line 75: Recomendation is create separate handler instead of usign anonymous function

IvanFesenko (approved):
Good job

PR #1728, opened 2023-07-26
solution

mariuszgit:


src/App.tsx
```diff
@@ -1,14 +1,52 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState<string>('');
+  const [activeFilter, setActiveFilter] = useState<string>('All');
```
Radoslaw-Czerniawski, line 16: Here for these values as they are set in stone you could use a union:
image

src/App.tsx
```diff
@@ -1,14 +1,52 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
```
Radoslaw-Czerniawski, line 20: Generally some of the logic could be refactored into a separate hook, for now I present you with some simple refactor and useMemo usage
image

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,25 @@
-export const TodoFilter = () => (
+/* eslint-disable jsx-a11y/control-has-associated-label */
+interface Props {
```
Radoslaw-Czerniawski, line 2: Generally there is a different way of typing setters for state:
image

Radoslaw-Czerniawski: And below typing of the select:
image

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,57 @@
...
-        <th>
-          <span className="icon">
-            <i className="fas fa-check" />
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+interface Props {
```
Radoslaw-Czerniawski, line 5: Same here:
image

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,37 @@
-import React from 'react';
+/* eslint-disable max-len */
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { Todo } from '../../types/Todo';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
+
+interface Props {
```
Radoslaw-Czerniawski, line 8: image

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,37 @@
...
+  selectedTodo: Todo
+  onModalClose: (value: null) => void
+}
+
+export const TodoModal: React.FC<Props> = ({ selectedTodo, onModalClose }: Props) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    const loadUsers = async () => {
```
Radoslaw-Czerniawski, line 18: Generally you should assume your api calls may fail, you have no error catching so your app is prone to throwing and dying on the user.

Radoslaw-Czerniawski: Also take notice that this loading and fetching logic is repeated, you could create your own custom hook that can fetch data and return { loading, value } object

Radoslaw-Czerniawski (changes requested):
- you have some repetitions regarding data fetching, you could some custom hook that will do the fetching and loading states while it's happening, it could return { loading, value, error } object that would change depending on the state of the fetch. (after implementing something similar message me I will send you some code snippet with nice pattern)
- there should be error handling in your implementation for HTTP requests
- typing could be improved

etojeDenys (changes requested):
provide the demo link to PR description

src/App.tsx
```diff
@@ -17,18 +53,26 @@ export const App: React.FC = () => {
...
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              <TodoList
+                todos={filteredByQuery}
+                onSelectingTodo={setSelectedTodo}
+              />
```
etojeDenys, line 69: let's show only loader when isLoading == true

etojeDenys (changes requested):
nit:

nazarmatsevych (approved):
Well done!

PR #1744, opened 2023-07-26
chore/dynamic-list-todos

FlameInBrains:
[DEMO LINK](https://FlameInBrains.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,4 +29,4 @@ loaded and show them using `TodoList` (check the code in the `api.ts`);
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_dynamic-list-of-todos/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://FlameInBrains.github.io/react_dynamic-list-of-todos/) and add it to the PR description.
--- a/src/api.ts
+++ b/src/api.ts
@@ -24,4 +24,4 @@ function get<T>(url: string): Promise<T> {
 
 export const getTodos = () => get<Todo[]>('/todos');
 
-export const getUser = (userId: number) => get<User>(`/users/${userId}`);
+export const getUser = (userId: number | undefined) => get<User>(`/users/${userId}`);
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,106 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -29,4 +29,4 @@ loaded and show them using `TodoList` (check the code in the `api.ts`);
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_dynamic-list-of-todos/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://FlameInBrains.github.io/react_dynamic-list-of-todos/) and add it to the PR description.
```
NataliGru, line 32: Your demo link isn't working(

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,106 @@
...
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filterBy}
+            onChange={
+              (event: React.ChangeEvent<HTMLSelectElement>) => setFilterBy(
+                event.target.value as FilteredBy,
```
NataliGru, line 57: You could create a separate handleChangeFilter

src/api.ts
```diff
@@ -24,4 +24,4 @@ function get<T>(url: string): Promise<T> {
 
 export const getTodos = () => get<Todo[]>('/todos');
 
-export const getUser = (userId: number) => get<User>(`/users/${userId}`);
+export const getUser = (userId: number | undefined) => get<User>(`/users/${userId}`);
```
And678, line 27: Never make requests that a guaranteed to fail. If userId is undefined, just don't call get user
```suggestion
export const getUser = (userId: number) => get<User>(`/users/${userId}`);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,105 @@
...
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+import { useState, useEffect } from 'react';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todos: Todo[];
```
And678, line 5: Don't pass the entire todos array here, just to filter it and return back. This is too complicated.

A better idea would be to move filtering logic to App.tsx. `TodoFilter`'s job is to select from `FilteredBy` enum

And678 (changes requested):

src/utils/filter.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  if (filterBy === 'completed') {
+    preparedTodos = todos.filter(todo => todo.completed);
+  }
+
+  if (query) {
+    return preparedTodos.filter(todo => todo.title
+      .toLowerCase()
+      .includes(query.toLowerCase()));
+  }
```
YegorKorenkov, line 22: don't forget to trim your search value
image

PR #1750, opened 2023-07-26
Solution

OlhaSheliakina:
[DEMO LINK](https://OlhaSheliakina.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,34 +1,108 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { Filter } from './types/Filter';
+
+function getVisibleTodos(todos: Todo[], query: string, filter: string) {
+  let preparedTodos = [...todos];
+  const normalizedQuery = query.toLowerCase().trim();
```
AlyonaSarapina, line 16: in an example of a working page our array changes if you type spaces. so, I'm not sure, if we should use trim().

AlyonaSarapina (approved):
Good job👍

src/App.tsx
```diff
@@ -1,34 +1,108 @@
...
+  useEffect(() => {
+    setLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .catch(error => setErrorMessage(`Error ${error.message}. Please try again`))
+      .finally(() => setLoading(false));
+  }, []);
+
+  const visibleTodos = getVisibleTodos(todos, query, filter);
```
And678, line 53: This needs to memoized, otherwise you are filtering on every render (e.g when selected todo changed) (useMemo)

And678 (changes requested):
Almost done, one thing to improve :)

YegorKorenkov (approved):
Good job!

PR #1752, opened 2023-07-26
add task solution

NadiaKovalchuk:
[DEMO LINK] https://NadiaKovalchuk.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,94 @@
+/* eslint-disable @typescript-eslint/no-use-before-define */
+/* eslint-disable no-console */
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
+import { getTodos, getUser } from './api';
 import { Loader } from './components/Loader';
+import { TodoModal } from './components/TodoModal';
+import { User } from './types/User';
+
+type FilterArgs = {
+  searchField: string;
+  sortField: 'completed' | 'active' | '';
+};
+
+function filteredTodos(todos: Todo[], { searchField, sortField }: FilterArgs) {
+  let visibleFilteredTodos = [...todos];
+
+  if (searchField) {
+    const searchFieldNormalize = searchField.toLowerCase().trim();
+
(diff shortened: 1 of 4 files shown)


kylyvnyk (approved):
Great job! Approve))

src/App.tsx
```diff
@@ -1,14 +1,108 @@
...
+    setSearchField(str);
+  };
+
+  const updateSortField = (
+    str: (typeof SortFieldEnum)[keyof typeof SortFieldEnum],
+  ) => {
+    setSortField(str);
+  };
+
+  const visibleTodos = filteredTodos(todos, { searchField, sortField });
```
And678, line 104: `useMemo` needed here, you filtering on every render currently

src/App.tsx
```diff
@@ -1,14 +1,108 @@
...
+  return visibleFilteredTodos;
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodo, setVisibleTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [visibleUser, setVisibleUser] = useState<User | null>(null);
+  const [isLoadingUser, setIsLoadingUser] = useState(false);
```
And678, line 59: Move these states and corresponding logic to `TodoModal`, it is only used there

And678 (changes requested):
Gj, but can be improved

src/App.tsx
NadiaKovalchuk: Hi) I use this states when create getUserById function and this function is a pros in TodoList )

src/App.tsx
```diff
@@ -1,14 +1,106 @@
...
+
+  const getUserById = (userId: number) => {
+    setIsLoadingUser(true);
+    getUser(userId)
+      .then((user) => setVisibleUser(user))
+      .catch((error) => {
+        console.warn(error);
+      })
+      .finally(() => setIsLoadingUser(false));
+  };
```
YegorKorenkov, line 72: you can move all user-related logic to the TodoModal component, why do you need it directly in the App component?
You click on todo and set coressponding todo object to state -> show the modal and take the user ID from the selected todo object and make request

YegorKorenkov (changes requested):

src/App.tsx
NadiaKovalchuk: Hi) cause I create the function getUserById, this function use use this states (  const [visibleUser, setVisibleUser] = useState<User | null>(null);
  const [isLoadingUser, setIsLoadingUser] = useState(false);) and then I use this function as a props in TodoList.

YegorKorenkov (approved):
Good job!

PR #1756, opened 2023-07-27
added task solution

tetianakostiuk1:
[DEMO LINK](https://tetianakostiuk1.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/api.ts
```diff
@@ -24,4 +24,4 @@ function get<T>(url: string): Promise<T> {
 
 export const getTodos = () => get<Todo[]>('/todos');
 
-export const getUser = (userId: number) => get<User>(`/users/${userId}`);
+export const getUser = (userId: number | undefined) => get<User>(`/users/${userId}`);
```
And678, line 27: Why call `getUser` with `undefined`, just don't call it if we don't have the userId.

And678: This way, you are making a network request that is guaranteed to fail

And678 (changes requested):

src/api.ts
And678, line 27:
```suggestion
export const getUser = (userId: number) => get<User>(`/users/${userId}`);
```

And678 (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+            return visibleTodos;
+        }
+      });
+  };
+
+  useEffect(() => {
+    getTodos()
+      .then(allTodos => setTodos(getVisibleTodos(allTodos, query, filter)))
+      .catch((error) => `Error when fetching todos ${error}`);
+  }, [query, filter]);
```
YegorKorenkov, line 49: why would you send a request for every change in the value of a filter or search? Please fix this

src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+        switch (selectedFilter) {
+          case 'completed':
+            return todo.completed;
+          case 'active':
+            return !todo.completed;
+          default:
+            return visibleTodos;
+        }
+      });
+  };
```
YegorKorenkov, line 43: Take it outside the component and wrap the function call in a useMemo, where to provide a filter, query dependency

src/api.ts
YegorKorenkov, line 27: Not fixed
image

YegorKorenkov (changes requested):
Please check comments

PR #1763, opened 2023-07-27
added task solution with useContext

vlopatko:
[DEMO LINK](https://vlopatko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,100 +1,44 @@
-import React from 'react';
+/*eslint-disable*/
+import TodoInfo from '../TodoInfo/TodoInfo';
+import { useAppContext } from '../Context/AppContext';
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,44 @@
...
-            <i className="fas fa-check" />
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+export const TodoList = () => {
+  const { todos, searchQuery, filterType } = useAppContext();
+  const preparedTodos = [...todos]
```
And678, line 7: Use useMemo here, to avoid extra renders

And678 (approved):
Looks good, nice context usage, one thing improve, and this is done :100:

PR #1771, opened 2023-07-29
add for test

LiliiaDanylenko:
[DEMO LINK](https://LiliiaDanylenko.github.io/react_dynamic-list-of-todos/)

Не розумію, чому не проходять тести...

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,24 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
+import { getTodos } from './api';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleTodos, setVisibleTodos] = useState(todos);
+  const [visibleModal, setVisibleModal] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos().then(setTodos);
+    getTodos().then(setVisibleTodos);
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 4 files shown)


LiliiaDanylenko:
Переробила

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,28 @@
...
+};
+
+export const TodoModal: React.FC<Props> = ({
+  visibleModal, setVisibleModal,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    getUser(visibleModal.userId).then(setUser);
+  }, []);
```
volodymyr-soltys97, line 19: The useEffect hook's dependency array is empty, which means it will only run once after the initial render. However, you are using 'visibleModal.userId' inside the hook, so this should be included in the dependency array. This way the hook will run again whenever 'visibleModal.userId' changes.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,28 @@
...
+  visibleModal: Todo;
+  setVisibleModal: (todo: Todo | null) => void;
+};
+
+export const TodoModal: React.FC<Props> = ({
+  visibleModal, setVisibleModal,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
```
volodymyr-soltys97, line 17: You are not handling errors in your asynchronous function. This can lead to uncaught exceptions if the promise rejects. Always handle errors in promises by using a .catch() block or a try/catch statement if inside an async function.

src/App.tsx
```diff
@@ -17,18 +48,35 @@ export const App: React.FC = () => {
...
+                query={query}
+                setQuery={setQuery}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0 ? (
+                <Loader />
```
volodymyr-soltys97, line 61: The length of todos being zero does not necessarily mean data is still loading. It could also mean that there are no todos. A separate loading state is recommended to differentiate these two cases.

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+      case Status.COMPLETED:
+        return currentTodos.filter(todo => todo.completed);
+
+      default:
+        return currentTodos;
+    }
+  }, [filter, query, todos]);
+
+  useEffect(() => {
+    getTodos().then(setTodos);
```
volodymyr-soltys97, line 40: You are not handling the error that might occur during the API call. It's a good practice to handle the potential error in a try-catch block or then-catch chain for promise-based functions.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
+      default:
+        return currentTodos;
+    }
+  }, [filter, query, todos]);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(error => {
+        throw new Error(error.message);
```
anastasiiavorobiova, line 44: The catch block is considered for handling errors, not throwing them

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
+  visibleModal, setVisibleModal,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    getUser(visibleModal.userId)
+      .then(setUser)
+      .catch(error => {
+        throw new Error(error.message);
```
anastasiiavorobiova, line 22: The catch block is considered for handling errors, not throwing them

anastasiiavorobiova (changes requested):
Great work! Almost done!

PR #1776, opened 2023-08-01
Solution

HannaVasylieva:
https://HannaVasylieva.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [filterStatus, setFilterStatus] = useState(FilterType.All);
+
+  useEffect(() => {
+    setLoading(true);
```
VitaliyBodnar, line 22: in this case, you can set isLoading `true` as a default value, because it called only once

VitaliyBodnar (changes requested):
Almost done!

src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+    getTodos()
+      .then(setTodos)
+      .catch((error) => {
+        throw new Error(error.message);
+      })
+      .finally(() => setLoading(false));
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    let prepareTodos = [...todos];
```
etojeDenys, line 31: we don't need to create the array copy here

etojeDenys (approved):
lgtm

PR #1779, opened 2023-08-02
solution

myplekan:
[DEMO LINK](https://myplekan.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [status, setStatus] = useState('all');
```
polosanya, line 16: Better use Enum instead of hardcoded values

src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
+  const [loading, setLoading] = useState(false);
```
polosanya, line 17:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
image

polosanya (changes requested):
Good job! Let's improve your code a bit)

volodymyr-soltys97 (approved):

PR #1781, opened 2023-08-04
solution

ShamievDavid:
[DEMO LINK](https://ShamievDavid.github.io/react_dynamic-list-of-todos/)


anastasiia-tilikina (changes requested):
Your app doesn't implement all the requirements for this task.
Please have a look at the task carefully and make sure that you've implemented all requirements before requesting review!

1. Input should have `x` only when the query is entered.
2. when you press the button which opens the modal the icon on the button should change while the modal is open.
3. look at [working version](https://mate-academy.github.io/react_dynamic-list-of-todos/). completed todos have a checkmark in the second column.
4. search should be case-insensitive.
5. make sure that tests are passed.

ShamievDavid:
make sure that tests are passed.

if someone can talk to me about tests. And show me how to resolve issues in it it would be nice. 
because now I cannot understand it. for example:
not passed case with modal window: should closes with close button. 
if you will click on close button on my modal window it will close, so how should i resolve this problem.
thank you.

sTorba24 (approved):
It's good!

PR #1792, opened 2023-08-09
solution

Krivets:


src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
+  const [selectTodoId, setSelectTodoId] = useState<number | null>(null);
+
+  const selectTodos = getFilteredTodos(todos, query, filter);
+  const selectTodo = selectTodos.find(todo => todo.id === selectTodoId);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
```
maxim2310, line 59: need catch block

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,19 +34,24 @@ export const TodoFilter = () => (
...
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
-        />
-      </span>
+      {query && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
maxim2310, line 46: why do you need it?

maxim2310 (changes requested):
Nice work, but would be better to see demo link)

src/components/TodoFilter/TodoFilter.tsx
Krivets: I need to show this button only if the query is set
image

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const [query, setQuery] = useState<string>('');
+  const [filter, setFilter] = useState<FilterType>(FilterType.ALL);
+  const [selectTodoId, setSelectTodoId] = useState<number | null>(null);
+
+  const selectTodos = getFilteredTodos(todos, query, filter);
+  const selectTodo = selectTodos.find(todo => todo.id === selectTodoId);
```
witflash, line 53: It's better to name `selectedTodo`, `selectedTodoId`, etc

src/App.tsx
```diff
@@ -17,18 +70,35 @@ export const App: React.FC = () => {
...
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectTodo && (
+        <TodoModal
+          todo={selectTodo}
+          reset={(todoId: null) => setSelectTodoId(todoId)}
```
witflash, line 99: If you want to reset selected todo do it this way:
```
reset={() => setSelectedTodoId(null)}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,31 +43,35 @@ export const TodoModal: React.FC = () => {
...
-              Todo #2
+              {`Todo #${id}`}
             </div>
 
             {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
             <button
               type="button"
               className="delete"
               data-cy="modal-close"
+              onClick={() => reset(null)}
```
witflash, line 54: and then here you can use this way:
```
onClick={reset}
```

witflash (changes requested):
Great work!
I've just proposed a little improvement.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,31 +43,35 @@ export const TodoModal: React.FC = () => {
...
 
               {' by '}
 
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
-              </a>
+              {user && (
+                <a href={`mailto:${user.email}`}>
+                  {user.name}
+                </a>
```
Anastasiia-Svintsova, line 73: Don't forget about the destructuring

Anastasiia-Svintsova (approved):

PR #1795, opened 2023-08-10
Develop

Spec0s:
https://spec0s.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,51 @@ export const TodoList: React.FC = () => (
...
+            >
+              {todo.title}
+            </p>
+          </td>
+          <td className="has-text-right is-vcentered">
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => onTodoSelected(todo)}
```
Anastasiia-Svintsova, line 63: Avoid creating functions within a component's render method. This can lead to unnecessary re-renders and hurt performance. Consider moving this function to a separate method.

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,51 @@ export const TodoList: React.FC = () => (
...
+          })}
+        >
+          <td className="is-vcentered">{todo.id}</td>
+          <td className="is-vcentered">
+            {todo.completed ? (
+              <span className="icon" data-cy="iconCompleted">
+                <i className="fas fa-check" />
+              </span>
+            ) : (
+              ''
```
Anastasiia-Svintsova, line 46: Instead of  rendering an empty string '' when todo is not completed, consider rendering null. This is a more common practice in React when you don't want to render anything. Use `&&` operator instead

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +32,32 @@ export const TodoModal: React.FC = () => {
...
+                <strong className="has-text-success">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
               {' by '}
-
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
+              <a href={`mailto:${user.email}`}>
+                {user?.name}
```
Anastasiia-Svintsova, line 60: Optional chaining on `user?.name` is not necessary because you've already checked if user is not null before rendering this part of the component. You can simply use `user.name`.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,29 @@
...
+};
+
+export const TodoModal: React.FC<Props> = ({
+  todo,
+  close,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    getUser(todo.userId).then(setUser);
```
Anastasiia-Svintsova, line 19: You have not handled the case if 'getUser' function fails. It would be beneficial to handle any potential errors that might arise from the API call to provide a better user experience.

src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const visibleTodos = useMemo(() => {
+    const lowerQuery = query.toLowerCase();
+
+    return todos.filter(todo => {
+      const isCompleted = status === Status.Completed ? todo.completed : true;
+      const isActive = status === Status.Active ? !todo.completed : true;
+      const isMatchQuery = todo.title.toLowerCase().includes(lowerQuery);
```
Anastasiia-Svintsova, line 38: This logic looks complicated. Consider refactoring it by using switch/case construction

src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
+
+      return isCompleted && isActive && isMatchQuery;
+    });
+  }, [todos, query, status]);
+
+  let content = null;
+
+  if (isLoading) {
+    content = <Loader />;
```
Anastasiia-Svintsova, line 47: Using let for content variable is not a good practice. Use a conditional rendering instead

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,51 @@ export const TodoList: React.FC = () => (
...
+            )}
+          </td>
+          <td className="is-vcentered is-expanded">
+            <p className={cn({
+              'has-text-danger': !todo.completed,
+              'has-text-success': todo.completed,
+            })}
+            >
+              {todo.title}
```
Anastasiia-Svintsova, line 55: Would be great to use destructuring for objects for better readability

Anastasiia-Svintsova (changes requested):

maxim2310 (approved):
Good work!!!

PR #1797, opened 2023-08-13
add task solution

kbekher:
[DEMO LINK](https://kbekher.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,68 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import React from 'react';
+import { Status } from '../../types/Status';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  setFilter: (value: Status) => void;
+  query: string
+  setQuery: (value: string) => void;
+};
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,83 @@
...
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+type Props = {
+  todos: Todo[];
+  selectedTodoId?: number;
+  setSelectedTodo?: (value: Todo) => void;
```
anastasiia-tilikina, line 8: why do you make `setSelectedTodo` and `selectedTodoId`  not required props?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +32,35 @@ export const TodoModal: React.FC = () => {
...
-              <strong className="has-text-danger">Planned</strong>
+              {selectedTodo.completed ? (
+                <strong className="has-text-success">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
 
               {' by '}
 
               <a href="mailto:Sincere@april.biz">
```
anastasiia-tilikina, line 62: your email is always the same, but it should change dynamically depending on user

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,68 @@
...
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={handleFilterChange}
+          >
+            <option value={Status.ALL}>All</option>
```
anastasiia-tilikina, line 33: you can use `map` here to render options as a list.

anastasiia-tilikina (changes requested):
well done!

let's add some fixes

src/components/TodoList/TodoList.tsx
kbekher: because selected Todo is possibly null, I've changed that and added additional checks..

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={handleFilterChange}
+          >
+
+            {(Object.keys(Status) as Array<keyof typeof Status>).map(key => (
+              <option value={Status[key]}>
```
Viktor-Kost, line 35: add react `key` since you map an array

Viktor-Kost (approved):

PR #1798, opened 2023-08-15
task solution

Bonum-taurus:
https://Bonum-taurus.github.io/react_dynamic-list-of-todos

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -8,7 +8,54 @@ import { TodoFilter } from './components/TodoFilter';
...
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [isOpenedModal, setIsOpenedModal] = useState(false);
+  const [idTodo, setIdTodo] = useState(0);
+  const [titleTodo, setTitleTodo] = useState('');
+  const [user, setUser] = useState<User | null>(null);
+  const [filter, setFilter] = useState('all');
+  const [query, setQuery] = useState('');
+
+  const filtredByQuery = (items: Todo[]) => {
+    return items.filter(item => item.title.includes(query));
```
TarasHoliuk, line 26: it should be case insensitive (check tests, requirements)

src/App.tsx
```diff
@@ -8,7 +8,54 @@ import { TodoFilter } from './components/TodoFilter';
...
+  };
+
+  const filteredTodos = (items: Todo[]) => {
+    if (filter === 'active') {
```
TarasHoliuk, line 30: use `switch`

src/App.tsx
```diff
@@ -8,7 +8,54 @@ import { TodoFilter } from './components/TodoFilter';
...
+  };
+
+  useEffect(() => {
+    getTodos().then(response => setTodos(response));
+  }, []);
+
+  const onOpenModal = (
+    value: boolean, todoId: number, userId: number, title: string,
+  ) => {
+    setIsOpenedModal(value);
```
TarasHoliuk, line 52: you will always set it `true` here - no need to pass it as value

src/App.tsx
```diff
@@ -8,7 +8,54 @@ import { TodoFilter } from './components/TodoFilter';
...
+    setIdTodo(todoId);
+    setTitleTodo(title);
+
+    getUser(userId).then(person => setUser(person));
```
TarasHoliuk, line 56: you use it only inside `TodoModal` - move this state to this component (load in useEffect)

src/App.tsx
```diff
@@ -8,7 +8,54 @@ import { TodoFilter } from './components/TodoFilter';
...
+      const filtredItems = items.filter(item => item.completed);
+
+      return filtredByQuery(filtredItems);
+    }
+
+    return filtredByQuery(items);
+  };
+
+  useEffect(() => {
+    getTodos().then(response => setTodos(response));
```
TarasHoliuk, line 46:
```suggestion
    getTodos().then(todos => setTodos(todos));
```
OR
```suggestion
    getTodos().then(setTodos);
```

src/App.tsx
```diff
@@ -17,18 +64,42 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                query={query}
+                onSetFilter={(filt: string) => setFilter(filt)}
```
TarasHoliuk, line 69:
```suggestion
                onSetFilter={setFilter}
```
in such cases additional wrapper function is redundant
fix all similar cases

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -8,7 +8,55 @@ import { TodoFilter } from './components/TodoFilter';
...
+  const [filter, setFilter] = useState<string>(Filter.all);
+  const [query, setQuery] = useState('');
+
+  const filtredByQuery = (items: Todo[]) => {
+    return items.filter(item => item.title.includes(query));
+  };
+
+  const filteredTodos = (items: Todo[]) => {
+    switch (filter) {
+      case 'active': {
```
volodymyr-soltys97, line 34: Need to use enum here instead of `'active'`

src/App.tsx
```diff
@@ -8,7 +8,55 @@ import { TodoFilter } from './components/TodoFilter';
...
+        const filtredItems = items.filter(item => !item.completed);
+
+        return filtredByQuery(filtredItems);
+      }
+
+      case 'completed': {
```
volodymyr-soltys97, line 40: The same as above

volodymyr-soltys97 (changes requested):
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #1803, opened 2023-08-22
add task solution

vovaAlieksieiev:
[DEMO LINK](https://vovaAlieksieiev.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -41,7 +41,9 @@ export const App: React.FC = () => {
   useEffect(() => {
     getTodos()
       .then(setTodos)
-      .catch(error => new Error(error.message))
+      .catch(error => {
+        throw error;
+      })
       .finally(() => setLoading(false));
   }, []);
 
@@ -80,7 +82,6 @@ export const App: React.FC = () => {
         <TodoModal
           visibleModal={visibleModal}
           setVisibleModal={setVisibleModal}
-
         />
       )}
     </>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,50 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
+import { getTodos } from './api';
 import { Loader } from './components/Loader';
+import { Status } from './Enum/Status';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [visibleModal, setVisibleModal] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState<Status>(Status.ALL);
+  const [query, setQuery] = useState('');
+  const [loading, setLoading] = useState(true);
+
+  const visibleTodos = useMemo(() => {
+    const currentTodos = query === '' ? todos : todos.filter(todo => (
+      todo.title.toLowerCase().includes(query.toLowerCase())
+    ));
+
+    switch (filter) {
+      case Status.ALL:
+        return currentTodos;
+
+      case Status.ACTIVE:
+        return currentTodos.filter(todo => !todo.completed);
+
+      case Status.COMPLETED:
+        return currentTodos.filter(todo => todo.completed);
+
+      default:
+        return currentTodos;
+    }
+  }, [filter, query, todos]);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(error => new Error(error.message))
+      .finally(() => setLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,78 @@
...
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {visibleTodos.map(todo => (
+          <tr
```
volodymyr-soltys97, line 31: Use a desctructuring for `todo`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,64 @@
...
+
+        {query && (
+
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => setQuery('')}
```
volodymyr-soltys97, line 55: Create a helper function for this logic and use them

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+
+      default:
+        return currentTodos;
+    }
+  }, [filter, query, todos]);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(error => new Error(error.message))
```
anastasiiavorobiova, line 44: There's a problem in your error handling. You are creating a new Error instance, but not throwing or logging it. This will make debugging difficult if the Promise from getTodos() fails.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
+  filter: Status,
+  setFilter: (filter: Status) => void,
+  query: string,
+  setQuery: (query: string) => void,
+};
+
+export const TodoFilter: React.FC<Props> = ({
+  filter, setFilter, query, setQuery,
+}) => {
+  const renderSearchIcon = () => (
```
anastasiiavorobiova, line 14: It's better to use a semantic HTML tag for the search icon. A button would be appropriate here.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
+    <span className="icon is-left">
+      <i className="fas fa-magnifying-glass" />
+    </span>
+  );
+
+  const renderClearButton = () => (
+    <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+      <button
```
anastasiiavorobiova, line 23: Eslint disabled for jsx-a11y/control-has-associated-label rule. Even though sometimes it is necessary to disable eslint rules, in this case it is better to fix the accessibility issue by adding a meaningful aria-label to the button.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,32 @@
...
+      .finally(() => setLoading(false));
+  }, [visibleModal.userId]);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {loading && !user ? (
```
anastasiiavorobiova, line 29: This line can lead to a condition where the Loader is displayed indefinitely. If 'getUser' fails, 'user' will be null and 'loading' will be false. A better check might be just for 'loading' and handle the error state separately

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,32 @@
...
+export const TodoModal: React.FC<Props> = ({
+  visibleModal, setVisibleModal,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    getUser(visibleModal.userId)
+      .then(setUser)
+      .catch(error => new Error(error.message))
```
anastasiiavorobiova, line 21: Error is being created but not thrown or logged, this is anti-pattern. It won't provide any feedback during runtime. Consider logging the error or throwing it

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,32 @@
...
+      .finally(() => setLoading(false));
+  }, [visibleModal.userId]);
```
anastasiiavorobiova, line 23: The useEffect is dependent on 'visibleModal.userId' but not 'setVisibleModal', which is also used inside the effect. It might cause problems if 'setVisibleModal' changes and the effect doesn't re-run. Include all used state variables in dependency array or use the callback form of the state setter function inside the effect

anastasiiavorobiova (changes requested):

mykhalenych (changes requested):
https://github.com/mate-academy/react_dynamic-list-of-todos/assets/53228259/28e003bc-daf1-4673-acf8-b8550aee78c5

fix it

PR #1804, opened 2023-08-23
Dynamic list of todo

ValeraViachalo:
[DEMO LINK](https://valeraviachalo.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,85 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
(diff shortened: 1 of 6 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,85 @@
...
+            onChange={handleFilterChange}
+          >
+            {(Object.keys(Status) as Array<keyof typeof Status>).map(key => (
+              <option
+                value={Status[key]}
+                key={key}
+              >
+                {capitalizeFirstLetter(key)}
+              </option>
+            ))}
```
IvanFesenko, line 44: Instead of creating new array on each render create const with enum values and use it here

`const SELECT_VALUES = [Status.ALL, Status.ACTIVE, Status.COMPLETED];`

IvanFesenko (changes requested):
Good job. Some improvements required

maxim2310 (approved):
Good work!!!

PR #1823, opened 2023-09-08
Develop

yuran007:
[DEMO LINK](https://yuran007.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,88 @@
...
+    getTodos()
+      .then(fetchedTodos => {
+        setTodos(fetchedTodos);
+        setIsLoading(false);
+      });
+  }, []);
+
+  const handleFilterSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    switch (event.target.value) {
+      case 'active':
```
Anastasiia-Svintsova, line 30: Just use an enum for it to avoid errors connected with typos in the future

src/App.tsx
```diff
@@ -1,14 +1,88 @@
...
+    }
+  };
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const clearQuery = () => setQuery(null);
+
+  const setTodoID = (todoID: number) => setSelectedTodoID(todoID);
```
Anastasiia-Svintsova, line 50: Do you really need a wrapper for the setSelectedTodoID here? It seems that this function just calls the `setSelectedTodoID`. You can call it directly

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,31 +35,33 @@ export const TodoModal: React.FC = () => {
...
+              >
+                {todo.completed ? 'Done' : 'Planned'}
+              </strong>
 
               {' by '}
 
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
-              </a>
+              <a href={`mailto:${user?.email}`}>{user?.name}</a>
```
Anastasiia-Svintsova, line 64: Just don't render this part if there is no user. And don't forget about object destructuring

src/App.tsx
```diff
@@ -1,14 +1,88 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { ShowTodos } from './types/ShowTodos';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodos, setSelectedTodos] = useState<ShowTodos>(ShowTodos.All);
```
Anastasiia-Svintsova, line 16: Seems that the `selectedTodos` name isn't descriptive enough. Maybe selectedStatus will be better?

Anastasiia-Svintsova (changes requested):
Great start!

src/App.tsx
yuran007: Done

anastasiia-tilikina (approved):
Well done!

PR #1826, opened 2023-09-11
react_dynamic-list-of-todos solution

anpiet:
[DEMO LINK](https://anpiet.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,11 +5,10 @@ import React, {
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
-import { getTodos, getUser } from './api';
+import { getTodos } from './api';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { Todo } from './types/Todo';
-import { User } from './types/User';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,111 @@
...
+
+      if (todo) {
+        getUser(todo?.userId).then(setUser).catch(e => {
+          setError(e.message);
+        }).finally(() => setLoadingUser(false));
+      }
+    };
+
+    loadUser();
+  }, [todo]);
```
choeqq, line 57: since user is used only in TodoModal, I think you can fetch it in that component

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,26 @@
...
+  selectedinpputfilter: string;
+  onclickhandle: () => void;
+};
+
+export const TodoFilter = ({
+  selectfilter,
+  selectinputfilter,
+  selectedfilter,
+  selectedinpputfilter,
+  onclickhandle,
```
choeqq, line 14: please use camelCase

choeqq (changes requested):

PR #1830, opened 2023-09-13
done

yaroslav5481:
https://yaroslav5481.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState, useMemo } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


yaroslav5481:
> Demo link isn't working. You forgot to deploy ypur work?
Yes, you are right. Please check now.

src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+      if (searchTerm.trim()
+        && !currentTodo.title.toLowerCase()
+          .includes(searchTerm.toLowerCase())
+      ) {
+        return false;
+      }
+
+      return true;
+    });
+  };
```
IvanFesenko, line 52: My assumption is spli this on 2 cases:
Why do you need need filter by query in case query is empt or why do you do you need filter by status in case query is not empty but is `Status.all`

src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+
+  const visibleTodos = useMemo(() => {
+    return filterTodos(todos, filter, query);
```
IvanFesenko, line 55: my thinking is no need in additiional function, place logic here

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,85 @@
...
+    setQuery(event.target.value);
+  };
+
+  const handleQueryClear = () => {
+    setQuery('');
+  };
+
+  const capitalizeFirstLetter = (str: string) => {
+    return str.charAt(0).toUpperCase() + str.slice(1).toLowerCase();
+  };
```
IvanFesenko, line 28: Can omit 
```suggestion
  const capitalizeFirstLetter = (str: string) => (
    str.charAt(0).toUpperCase() + str.slice(1).toLowerCase()
  )
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,86 @@
...
-        </td>
-      </tr>
+export const TodoList: React.FC<Props> = ({
+  todo, selectedTodo, setSelectedTodo,
+}) => {
+  const handleClick = (currentTodo: Todo) => {
+    if (setSelectedTodo) {
+      setSelectedTodo(currentTodo);
+    }
+  };
```
IvanFesenko, line 18: Let's modify a bit
```suggestion
  const handleClick = (currentTodo: Todo) => () => {
    if (setSelectedTodo) {
      setSelectedTodo(currentTodo);
    }
  };
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,86 @@
...
+              >
+                {currentTodo.title}
+              </p>
+            </td>
+            <td className="has-text-right is-vcentered">
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => handleClick(currentTodo)}
```
IvanFesenko, line 69:
```suggestion
                onClick={handleClick(currentTodo)}
```

IvanFesenko (changes requested):
Good joib. Check my comments

IvanFesenko (approved):

PR #1835, opened 2023-09-14
Solved task

Wesses:
[DEMO LINK](https://Wesses.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useContext } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -9,10 +9,11 @@ import { getTodos } from './api';
 import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { TodoContext } from './components/TodoContext';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [showedTodo, setShowedTodo] = useState<Todo | null>(null);
+  const { showedTodo } = useContext(TodoContext);
 
   useEffect(() => {
     getTodos().then(setTodos);
@@ -31,7 +32,7 @@ export const App: React.FC = () => {
 
             <div className="block">
               {todos.length ? (
-                <TodoList todos={todos} onShowedTodo={setShowedTodo} />
+                <TodoList todos={todos} />
               ) : (
                 <Loader />
               )}
@@ -40,7 +41,7 @@ export const App: React.FC = () => {
         </div>
       </div>
 
-      {showedTodo && <TodoModal showedTodo={showedTodo} />}
+      {showedTodo && <TodoModal />}
     </>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,9 @@
-/* eslint-disable max-len */
-import React, { useEffect, useState, useContext } from 'react';
+import React, {
+  useEffect,
+  useState,
+  useContext,
+  useMemo,
+} from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -10,10 +14,48 @@ import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 import { TodoContext } from './components/TodoContext';
+import { Filter, FilterEnum } from './types/Filter';
+
+const DEFAULT_FILTER: Filter = {
+  select: FilterEnum.all,
+  input: '',
+};
+
+const getFiltredTodos = (filter: Filter, todos: Todo[]): Todo[] => {
+  const { input, select } = filter;
+
+  if (!input && select === FilterEnum.all) {
+    return todos;
+  }
+
+  let newTodos = [...todos];
+  const validatedInput = input.toLowerCase().trim();
+
+  newTodos = newTodos
+    .filter(({ title }) => title.toLowerCase().includes(validatedInput));
+
+  switch (select) {
+    case FilterEnum.active:
+      newTodos = newTodos.filter(({ completed }) => !completed);
+      break;
+    case FilterEnum.completed:
+      newTodos = newTodos.filter(({ completed }) => completed);
+      break;
+    default:
+  }
+
+  return newTodos;
+};
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<Filter>(DEFAULT_FILTER);
   const { showedTodo } = useContext(TodoContext);
+
+  const filtredTodos = useMemo(
+    () => getFiltredTodos(filter, todos),
+    [filter, todos],
+  );
 
   useEffect(() => {
     getTodos().then(setTodos);
@@ -27,12 +69,12 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter filter={filter} onFilter={setFilter} />
             </div>
 
             <div className="block">
               {todos.length ? (
-                <TodoList todos={todos} />
+                <TodoList todos={filtredTodos} />
               ) : (
                 <Loader />
               )}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,23 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [showedTodo, setShowedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos().then(setTodos);
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -21,14 +30,17 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length ? (
+                <TodoList todos={todos} onShowedTodo={setShowedTodo} />
+              ) : (
+                <Loader />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {showedTodo && <TodoModal showedTodo={showedTodo} />}
     </>
   );
 };
--- a/src/components/TodoContext/TodoContext.tsx
+++ b/src/components/TodoContext/TodoContext.tsx
@@ -1,0 +1,26 @@
+import React, { createContext, useState } from 'react';
+import { Todo } from '../../types/Todo';
+
+type ContextObject = {
+  showedTodo: Todo | null;
+  setShowedTodo: (u: Todo | null) => void;
+};
+
+export const TodoContext = createContext<ContextObject>({
+  showedTodo: null,
+  setShowedTodo: () => {},
+});
+
+type Props = {
+  children: React.ReactNode;
+};
+
+export const TodoContextProvider:React.FC<Props> = ({ children }) => {
+  const [showedTodo, setShowedTodo] = useState<Todo | null>(null);
+
+  return (
+    <TodoContext.Provider value={{ showedTodo, setShowedTodo }}>
+      {children}
+    </TodoContext.Provider>
+  );
+};
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,75 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import React from 'react';
+import { Filter, FilterEnum } from '../../types/Filter';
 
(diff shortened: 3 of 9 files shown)


yuliiaant (approved):
Great job!

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,93 @@
...
+            >
+              <td className="is-vcentered">{id}</td>
+              {completed ? (
+                <td className="is-vcentered">
+                  <span className="icon" data-cy="iconCompleted">
+                    <i className="fas fa-check" />
+                  </span>
+                </td>
+              ) : (
+                <td className="is-vcentered" />
```
tiserett, line 60: create todo component

tiserett (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
+const getFiltredTodos = (filter: Filter, todos: Todo[]): Todo[] => {
+  const { input, select } = filter;
+
+  if (!input && select === FilterEnum.all) {
+    return todos;
+  }
+
+  const validatedInput = input.toLowerCase().trim();
+  let newTodos = [...todos]
+    .filter(({ title }) => title.toLowerCase().includes(validatedInput));
```
Moroz-Dmytro, line 33: Why you need to make copy here?

src/App.tsx
```diff
@@ -17,18 +67,21 @@ export const App: React.FC = () => {
...
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length ? (
+                <TodoList todos={filtredTodos} />
+              ) : (
+                <Loader />
+              )}
```
Moroz-Dmytro, line 78: Would be better to create state for `isLoading` and change it before and after you make API requests
What if you will get an empty response from API? You will see the loader all the time

src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
+  const [filter, setFilter] = useState<Filter>(DEFAULT_FILTER);
+  const { showedTodo } = useContext(TodoContext);
+
+  const filtredTodos = useMemo(
+    () => getFiltredTodos(filter, todos),
+    [filter, todos],
+  );
+
+  useEffect(() => {
+    getTodos().then(setTodos);
```
Moroz-Dmytro, line 59: Add `catch` block

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,77 @@
...
+  };
+
+  const handleChangeInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    onFilter({
+      ...filter,
+      input: event.target.value,
+    });
+  };
+
+  const handleClickCross = () => {
```
Moroz-Dmytro, line 24:
```suggestion
  const handleClearInput = () => {
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,77 @@
...
+            <button
+              aria-label="deleteButton"
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={handleClickCross}
+            />
+          </span>
+        )}
+
```
Moroz-Dmytro, line 73:
```suggestion
```

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,71 @@
...
+  };
+
+  const handleMouseLeave = (event: React.MouseEvent<HTMLTableRowElement>) => {
+    event.currentTarget.classList.remove('has-background-info-light');
+  };
+
+  return (
+    <tr
+      data-cy="todo"
+      className=""
```
Moroz-Dmytro, line 30: Delete empty className

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,30 @@
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map((todo) => <TodoItem todo={todo} key={todo.id} />)}
```
Moroz-Dmytro, line 26:
```suggestion
        {todos.map((todo) => (
          <TodoItem todo={todo} key={todo.id} />
        ))}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,24 @@
...
+    }
+  }, []);
+
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!showedUser ? (
         <Loader />
```
Moroz-Dmytro, line 22: The same comment about loader as above

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,24 @@
...
+import { getUser } from '../../api';
+import { TodoContext } from '../TodoContext';
 
 export const TodoModal: React.FC = () => {
+  const [showedUser, setShowedUser] = useState<User | null>(null);
+  const { showedTodo, setShowedTodo } = useContext(TodoContext);
+
+  useEffect(() => {
+    if (showedTodo?.userId) {
+      getUser(showedTodo.userId).then(setShowedUser);
```
Moroz-Dmytro, line 13: Add `catch`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +27,36 @@ export const TodoModal: React.FC = () => {
...
-              quis ut nam facilis et officia qui
+              {showedTodo?.title}
             </p>
 
             <p className="block" data-cy="modal-user">
               {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              <strong
+                className={showedTodo?.completed
+                  ? 'has-text-success' : 'has-text-danger'}
```
Moroz-Dmytro, line 51: Use `classnames`

src/types/Filter.ts
```diff
@@ -0,0 +1,10 @@
+export enum FilterEnum {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
```
Moroz-Dmytro, line 4:
```suggestion
  All = 'all',
  Active = 'active',
  Completed = 'completed',
```

Moroz-Dmytro (changes requested):

src/types/Filter.ts
```diff
@@ -0,0 +1,10 @@
+export enum FilterEnum {
```
roman-mirzoian, line 1:
```suggestion
export enum TodoStates {
```

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+  const filtredTodos = useMemo(
+    () => getFiltredTodos(filter, todos),
+    [filter, todos],
+  );
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(setTodos)
+      .catch(() => setTodos([]))
```
roman-mirzoian, line 63: In the `catch()` block, it is necessary to catch an error in order to understand what went wrong.
Fix in all code.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,76 @@
...
+          <select
+            data-cy="statusSelect"
+            value={filter.select}
+            onChange={handleChangeSelect}
+          >
+            {Object.values(FilterEnum).map(value => (
+              <option value={value} key={value}>
+                {value.at(0)?.toUpperCase() + value.slice(1)}
+              </option>
+            ))}
```
roman-mirzoian, line 44: Here it is now impossible to understand what is happening, transfer the calculations to separate variables and helpers and use ready-made results with good naming.

roman-mirzoian (changes requested):
Now, when opening information about the place, the modal window flashes, try to fix it, if you can't - ask for help:
https://gyazo.com/c1bd1c7720463aab7cf9392f54de0636

src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
+
+  switch (select) {
+    case TodoStates.Active:
+      newTodos = newTodos.filter(({ completed }) => !completed);
+      break;
+    case TodoStates.Completed:
+      newTodos = newTodos.filter(({ completed }) => completed);
+      break;
+    default:
+  }
```
BudnikOleksii, line 43: Nothing happens in the `default` case.
Also, You can filter just one time instead of filtering by query and completed separately

src/App.tsx
```diff
@@ -17,18 +75,24 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading ? (
+                <Loader />
+              ) : (
+                <>
+                  <TodoList todos={filtredTodos} />
+                  {!filtredTodos.length && 'No todos'}
```
BudnikOleksii, line 87:
```suggestion
                 {filtredTodos.length > 0 ? (
                    <TodoList todos={filtredTodos} />
                  ) : (
                    'No todos'
                  )}
```
Looks like if you have no todos you don't need to render `TodoList` at all

src/components/TodoContext/TodoContext.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { createContext, useState } from 'react';
+import { Todo } from '../../types/Todo';
+
+type ContextObject = {
+  showedTodo: Todo | null;
+  setShowedTodo: (u: Todo | null) => void;
+};
```
BudnikOleksii, line 7: `u: Todo | null` what does it mean `u`? Avoid one character variables

Also, `ContextObject` isn't an informative type name, it looks like `TodoContext`, you can't use this type for other contexts

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,82 @@
...
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+import React from 'react';
+import { Filter, TodoStates } from '../../types/Filter';
+
+function capitalizer(str: string): string {
+  return str.at(0)?.toUpperCase() + str.slice(1);
+}
```
BudnikOleksii, line 6: As you use it only for UI, to display in options you can just use CSS property - `text-transform: capitalize;`
https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,82 @@
...
+
+const todoStatesArray = Object.values(TodoStates);
+
+type Props = {
+  filter: Filter;
+  onFilter: (f: Filter) => void;
```
BudnikOleksii, line 12: Avoid one-character variable names

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,82 @@
...
+};
+
+export const TodoFilter: React.FC<Props> = ({ filter, onFilter }) => {
+  const handleChangeSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    onFilter({
+      ...filter,
+      select: event.target.value as TodoStates,
+    });
```
BudnikOleksii, line 20: As I see `onFilter` is a `setFilter` function, so you can use functional setState to get previous values.
Fix it in all places

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,82 @@
...
         />
-      </span>
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {!!filter.input && (
```
BudnikOleksii, line 68:
```suggestion
        {filter.input && (
```
You don't need to convert it to a boolean, not empty string is already truthy value
https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,82 @@
...
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
BudnikOleksii, line 69: Try to don't use inline styles

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const { showedTodo, setShowedTodo } = useContext(TodoContext);
+
+  const handleMouseEnter = (event: React.MouseEvent<HTMLTableRowElement>) => {
+    event.currentTarget.classList.add('has-background-info-light');
+  };
+
+  const handleMouseLeave = (event: React.MouseEvent<HTMLTableRowElement>) => {
+    event.currentTarget.classList.remove('has-background-info-light');
+  };
```
BudnikOleksii, line 25: Don't manipulate DOM elements directly, create a state and add the class conditionally to the element(use classnames library)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,32 @@
...
+        .finally(() => setIsLoading(false));
+    }
+  }, []);
+
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {isLoading && !showedUser ? (
```
BudnikOleksii, line 29:
```suggestion
      {isLoading ? (
```
It's better to check if it's loading, so the user won't be confused about why the user stays the same while loading

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +35,37 @@ export const TodoModal: React.FC = () => {
               className="modal-card-title has-text-weight-medium"
               data-cy="modal-header"
             >
-              Todo #2
+              {`Todo #${showedTodo?.id}`}
```
BudnikOleksii, line 38: If you don't have todo just don't render UI, as in your case you render it but with empty elements

BudnikOleksii (changes requested):
Great work 👍 
But there are some things that have to be fixed, check the comments

src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
+  if (!input && select === TodoStates.All) {
+    return todos;
+  }
+
+  const validatedInput = input.toLowerCase().trim();
+
+  switch (select) {
+    case TodoStates.Active:
+      return todos.filter(({ completed, title }) => (
+        title.toLowerCase().includes(validatedInput) && !completed));
```
BudnikOleksii, line 36: You can create a helper for `title.toLowerCase().includes(validatedInput)` as yo do it 3 times or `filter` just once and do these checks inside it

PR #1838, opened 2023-09-14
Develop

Koliras:
[DEMO LINK](https://Koliras.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,71 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+function getFilteredTodos(key: string, query: string, todosToFilter: Todo[]) {
+  let filteredTodos = [];
+
+  switch (key) {
+    case 'all': {
+      filteredTodos = todosToFilter;
+      break;
+    }
+
+    case 'active': {
+      filteredTodos = todosToFilter.filter(todo => !todo.completed);
+      break;
+    }
+
+    case 'completed': {
+      filteredTodos = todosToFilter.filter(todo => todo.completed);
+      break;
+    }
+
+    default:
+      filteredTodos = todosToFilter;
+  }
+
+  filteredTodos = filteredTodos.filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()));
+
+  return filteredTodos;
+}
 
 export const App: React.FC = () => {
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [loading, setLoading] = useState(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterKey, setFilterKey] = useState('all');
+  const [query, setQuery] = useState('');
+
+  const filteredTodos = getFilteredTodos(filterKey, query, todos);
+
+  const handleTodoSelection = (todo: Todo | null) => {
+    setSelectedTodo(todo);
+  };
+
+  const handleFilterKeySelection = (key: string) => {
+    setFilterKey(key);
+  };
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const handleQueryDelete = () => {
+    setQuery('');
+  };
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+      })
+      .finally(() => setLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,77 @@
...
+    case 'completed': {
+      filteredTodos = todosToFilter.filter(todo => todo.completed);
+      break;
+    }
+
+    default:
+      filteredTodos = todosToFilter;
+  }
+
+  filteredTodos = filteredTodos.filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()));
```
rostyslav48, line 36: it is better to trim query `(query.toLowerCase().trim())` in this way we will remove extra spaces and improve our search system

src/App.tsx
```diff
@@ -1,14 +1,77 @@
...
+
+  return filteredTodos;
+}
 
 export const App: React.FC = () => {
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [loading, setLoading] = useState(false);
```
rostyslav48, line 43: You are using a boolean value for this var, so consider naming it something like `isLoading`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,22 @@
-export const TodoFilter = () => (
+import React from 'react';
+
+type Props = {
+  onFilterKeyChange: (key: string) => void,
+  onQueryChange: (event: React.ChangeEvent<HTMLInputElement>) => void,
+  onDelete: () => void
+  query: string,
```
rostyslav48, line 7: It is really cool that you added some optimization features. Although, you pass some props which are not wrapped in `useMemo()`. The question is whether this optimization features needed.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,35 +35,40 @@ export const TodoModal: React.FC = () => {
...
           <div className="modal-card-body">
             <p className="block" data-cy="modal-title">
-              quis ut nam facilis et officia qui
+              {selectedTodo.title}
             </p>
 
             <p className="block" data-cy="modal-user">
               {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              <strong className="has-text-danger">
```
rostyslav48, line 57: You haven't changed color for the done tasks
Знімок екрана 2023-09-14 о 22 14 19

rostyslav48 (changes requested):
Well Done👍
However there is a tiny problem with your app please check the last comment and fix it. All other comments are just a suggestions, so you may take them into consideration)

src/App.tsx
```diff
@@ -1,14 +1,79 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+function getFilteredTodos(key: string, query: string, todosToFilter: Todo[]) {
+  let filteredTodos = [];
+
+  switch (key) {
+    case 'all': {
```
Moroz-Dmytro, line 17: Create `enum` for these values

src/App.tsx
```diff
@@ -17,18 +82,34 @@ export const App: React.FC = () => {
...
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              {!isLoading && (
+                <TodoList
+                  todos={filteredTodos}
+                  onClick={handleTodoSelection}
+                  selectedTodo={selectedTodo}
+                />
```
Moroz-Dmytro, line 100: Use ternary here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,84 @@
...
+              {todo.completed
+                ? (
+                  <td className="is-vcentered">
+                    <span className="icon" data-cy="iconCompleted">
+                      <i className="fas fa-check" />
+                    </span>
+                  </td>
+                ) : (
+                  <td className="is-vcentered" />
+                )}
```
Moroz-Dmytro, line 51: DRY. `<td className="is-vcentered" />` is the same for both variants

Moroz-Dmytro (changes requested):
I guess, you forgot to do deploy again, because I don't see changes after your fixes

src/App.tsx
roman-mirzoian: Not fixed 😢

src/App.tsx
```diff
@@ -1,14 +1,85 @@
...
+    setQuery('');
+  }, []);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(todosFromServer => {
+        setTodos(todosFromServer);
+      })
+      .catch(() => setTodos([]))
```
roman-mirzoian, line 79: In the `catch()` block, it is necessary to catch an error in order to understand what went wrong.
Fix in all code.

src/components/TodoList/TodoList.tsx
roman-mirzoian: Not fixed 😢

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,35 +36,44 @@ export const TodoModal: React.FC = () => {
...
-              Todo #2
+              {`Todo #${selectedTodo.id}`}
             </div>
 
             {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
             <button
               type="button"
               className="delete"
               data-cy="modal-close"
+              onClick={() => onClick(null)}
```
roman-mirzoian, line 47: From the name `onClick(null)` it is not very clear what will happen after the click.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,19 +33,24 @@ export const TodoFilter = () => (
...
-        />
-      </span>
+      {!!query && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="clearSearchButton"
+            type="button"
+            className="delete"
+            onClick={onDelete}
```
roman-mirzoian, line 50: From the name onDelete it is not very clear what will delete after the click.

roman-mirzoian (changes requested):
Do not ignore comments from the mentor: if you do not agree with the comment - write it, if you did not understand the comment - ask the mentor in the personal channel what he meant.
Also, now when opening information about the place, the modal window flashes, try to fix it, if you can't - ask for help:
https://gyazo.com/c1bd1c7720463aab7cf9392f54de0636

alexander-ignatow (approved):
LGTM!

PR #1841, opened 2023-09-14
get data from the server, make functional for filter module, for modal and list

rostyslav48:
[DEMO LINK](https://rostyslav48.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -54,7 +54,8 @@ export const App: React.FC = () => {
       .then((todoList) => {
         setTodosFromServer(todoList);
       })
-      .catch()
+      // eslint-disable-next-line
+      .catch((error) => console.warn(error))
       .finally(() => setIsLoading(false));
   }, []);
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,59 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { CompletionFilter } from './types/CompletionFilter';
+
+function filterByQuery(todos: Todo[], searchQuery: string) {
+  const normalisedQuery = searchQuery.toLowerCase().trim();
+
+  return todos.filter(({ title }) => title.toLowerCase().includes(normalisedQuery));
+}
+
+function filterByCompletion(todos: Todo[], completionParam: CompletionFilter) {
+  if (completionParam === CompletionFilter.All) {
+    return todos;
+  }
+
+  const completionStatus = completionParam === CompletionFilter.Completed;
+
+  return todos.filter(({ completed }) => completed === completionStatus);
+}
+
+function postsFilter(dataToFilter: Todo[], completionParam: CompletionFilter, searchQuery: string) {
+  let filteredTodos = dataToFilter;
+
+  filteredTodos = filterByQuery(filteredTodos, searchQuery);
+  filteredTodos = filterByCompletion(filteredTodos, completionParam);
+
+  return filteredTodos;
+}
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+
+  const [isLoaderVisible, setIsLoaderVisible] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  const [completionFilter, setCompletionFilter] = useState<CompletionFilter>(CompletionFilter.All);
+  const [searchQuery, setSearchQuery] = useState('');
+
+  useEffect(() => {
+    setIsLoaderVisible(true);
+
+    getTodos()
+      .then((todoList) => {
+        setTodosFromServer(todoList);
+      })
+      .finally(() => setIsLoaderVisible(false));
+  }, []);
+
+  useEffect(() => {
+    setFilteredTodos(postsFilter(todosFromServer, completionFilter, searchQuery));
+  }, [todosFromServer, completionFilter, searchQuery]);
+
   return (
     <>
       <div className="section">
@@ -17,18 +68,34 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                searchQuery={searchQuery}
+                changeSearchQuery={setSearchQuery}
+                completionStatus={completionFilter}
+                changeCompletionStatus={setCompletionFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoaderVisible
+                ? <Loader />
+                : (
+                  <TodoList
+                    todos={filteredTodos}
+                    selectedTodo={selectedTodo}
+                    changeSelectedTodo={setSelectedTodo}
+                  />
+                )}
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal
+          selectedTodo={selectedTodo}
+          changeSelectedTodo={setSelectedTodo}
+        />
+      )}
     </>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -27,7 +27,7 @@ function filterByCompletion(todos: Todo[], completionParam: CompletionFilter) {
   return todos.filter(({ completed }) => completed === completionStatus);
 }
 
-function postsFilter(dataToFilter: Todo[], completionParam: CompletionFilter, searchQuery: string) {
+function getFilteredTodos(dataToFilter: Todo[], completionParam: CompletionFilter, searchQuery: string) {
   let filteredTodos = dataToFilter;
 
   filteredTodos = filterByQuery(filteredTodos, searchQuery);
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+  filteredTodos = filterByCompletion(filteredTodos, completionParam);
+
+  return filteredTodos;
+}
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+
+  const [isLoaderVisible, setIsLoaderVisible] = useState(false);
```
Moroz-Dmytro, line 43:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+  const [searchQuery, setSearchQuery] = useState('');
+
+  useEffect(() => {
+    setIsLoaderVisible(true);
+
+    getTodos()
+      .then((todoList) => {
+        setTodosFromServer(todoList);
+      })
+      .finally(() => setIsLoaderVisible(false));
```
Moroz-Dmytro, line 56: Add `catch` for handling errors

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+  if (completionParam === CompletionFilter.All) {
+    return todos;
+  }
+
+  const completionStatus = completionParam === CompletionFilter.Completed;
+
+  return todos.filter(({ completed }) => completed === completionStatus);
+}
+
+function postsFilter(dataToFilter: Todo[], completionParam: CompletionFilter, searchQuery: string) {
```
Moroz-Dmytro, line 30:
```suggestion
function getFilteredTodos(dataToFilter: Todo[], completionParam: CompletionFilter, searchQuery: string) {
```

src/App.tsx
```diff
@@ -1,14 +1,65 @@
...
+
+  filteredTodos = filterByQuery(filteredTodos, searchQuery);
+  filteredTodos = filterByCompletion(filteredTodos, completionParam);
+
+  return filteredTodos;
+}
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
Moroz-Dmytro, line 41: You don't really need to make this as a state, you can use just constant for this

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,76 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            value={completionStatus}
+            data-cy="statusSelect"
+            onChange={handleCompletionStatusChange}
+          >
+            <option value={CompletionFilter.All}>All</option>
+            <option value={CompletionFilter.Active}>Active</option>
+            <option value={CompletionFilter.Completed}>Completed</option>
```
Moroz-Dmytro, line 44: Use `map` here

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,76 @@
...
+        changeCompletionStatus(CompletionFilter.Completed);
+        break;
+
+      case CompletionFilter.Active:
+        changeCompletionStatus(CompletionFilter.Active);
+        break;
+
+      default:
+        changeCompletionStatus(CompletionFilter.All);
+    }
```
Moroz-Dmytro, line 30: Why you need `switch/case` here?

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,63 @@
...
+            <i className="fas fa-check" />
+          </span>
+        )}
+      </td>
+
+      <td className="is-vcentered is-expanded">
+        <p
+          className={completed
+            ? 'has-text-success'
+            : 'has-text-danger'}
```
Moroz-Dmytro, line 37: Use `classnames`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  changeSelectedTodo,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  const {
+    id, title, userId, completed,
+  } = selectedTodo;
+
+  useEffect(() => {
+    getUser(userId).then((foundUser) => setUser(foundUser));
```
Moroz-Dmytro, line 23: Add `catch`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  }, [selectedTodo]);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!user ? (
```
Moroz-Dmytro, line 30: Better to create `isLoading` state for handling the loader state. What if you get an empty response for the user? You will see the loader all the time

Moroz-Dmytro (changes requested):

src/components/TodoFilter/TodoFilter.tsx
rostyslav48: event.target.value is a string value, and typescript is complaining that we cannot assign a string value to `CompletionFilter` type. Honestly, I don't really like this switch case too, it looks messy. So If you have any suggestions how to fix this problem, I would like to hear them.

src/components/TodoItem/TodoItem.tsx
rostyslav48: Is it a good idea to use a classnames in this case? From my perspective, in this situation when we are choosing 1 class of 2 it is better to use this operator to avoid duplications.

src/components/TodoModal/TodoModal.tsx
rostyslav48: But in this case, we cannot be sure that we will get our user. I suppose it would be a good idea, if we had some error messages in this task

Moroz-Dmytro: You can add some error message(design is up to you), or just print a message to the console(for now, to not waste a lot of time) and close the modal window, it will be better than just show the loader all the time

src/components/TodoItem/TodoItem.tsx
Moroz-Dmytro: I guess that yes. It will be also easy to read. But also in the future, if you need to add 1 more class, you will refactor your code anyway using `classnames`, so why not use it immediately

src/components/TodoFilter/TodoFilter.tsx
Moroz-Dmytro: I guess, you can use  `as CompletionFilter` in this case

rostyslav48: I didn't manage to fix this problem. I have tried different approaches, but it didn't work(
Знімок екрана 2023-09-19 о 21 51 39
Знімок екрана 2023-09-19 о 21 52 37
Знімок екрана 2023-09-19 о 21 53 44

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+  const filteredTodos = getFilteredTodos(todosFromServer, completionFilter, searchQuery);
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then((todoList) => {
+        setTodosFromServer(todoList);
+      })
+      .catch()
```
Moroz-Dmytro, line 57: You need at least print some message to console, that something went wrong

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,78 @@
...
```
Moroz-Dmytro, line 30:
```suggestion
    changeCompletionStatus(event.target.value as CompletionFilter);
```

Moroz-Dmytro: Have you tried this?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,38 @@
...
+
+  const {
+    id, title, userId, completed,
+  } = selectedTodo;
+
+  useEffect(() => {
+    setIsLoading(true);
+    getUser(userId)
+      .then((foundUser) => setUser(foundUser))
+      .catch()
```
Moroz-Dmytro, line 27: The same comment as above, you can print something to the console, or if you want to show some message on UI you can create `isError` state and show this message depending on this state value

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,33 +41,54 @@ export const TodoModal: React.FC = () => {
...
+              onClick={() => changeSelectedTodo(null)}
             />
           </header>
 
-          <div className="modal-card-body">
-            <p className="block" data-cy="modal-title">
-              quis ut nam facilis et officia qui
-            </p>
+          {user
+            ? (
```
Moroz-Dmytro, line 57: Try not to nest ternary, it's difficult for understanding

src/types/Todo.ts
```diff
@@ -1,6 +1,12 @@
+import { User } from './User';
+
 export interface Todo {
   id: number;
   title: string;
   completed: boolean;
   userId: number;
 }
+
+export interface PreparedTodo extends Todo {
```
Moroz-Dmytro, line 10: Why do you need this `interface`? You don't use it

Moroz-Dmytro (changes requested):
Options names should start with an upper letter("App"...)

image

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,69 @@
...
-          className="delete"
+export const TodoFilter: React.FC<Props> = ({
+  searchQuery,
+  changeSearchQuery,
+  completionStatus,
+  changeCompletionStatus,
+}) => {
+  const capitalizeFirstLetter = (string: string) => {
+    return string[0].toUpperCase() + string.slice(1);
+  };
```
BudnikOleksii, line 18: In CSS we have `text-transform: capitalize;` so you can use it instead of capitalizing with js:
https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform

Also, if you create helper functions move them outside or to utils/helpers folder

src/components/TodoModalMessage/TodoModalMessage.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    user
+      ? (
+        <div className="modal-card-body">
+          <p className="block" data-cy="modal-title">
+            {title}
+          </p>
+
+          <p className="block" data-cy="modal-user">
+            {/* <strong className="has-text-success">Done</strong> */}
```
BudnikOleksii, line 24: Remove comments

BudnikOleksii (approved):
Good work 👍 
Approved, but left a few comments, check them

PR #1849, opened 2023-09-15
Add task solution

ArtemBehter:


src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
+
+    setSelectedTodo(findTodo);
+  };
+
+  const handleModalClose = () => {
+    setSelectedTodo(null);
+  };
+
+  const handleFilter = (status: string) => {
+    if (status === 'all') {
```
OleziO, line 31: Move all cases to the enum.

OleziO (changes requested):
Fix demo link adn add it to the PR descriptiion
image

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoaded, setIsLoaded] = useState<boolean>(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [titleFilter, setTitleFilter] = useState<string>('');
+  const [filter, setFilter] = useState<string>('');
```
etojeDenys, line 18:
```suggestion
  const [isLoaded, setIsLoaded] = useState(false);
  const [todos, setTodos] = useState<Todo[]>([]);
  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
  const [titleFilter, setTitleFilter] = useState('');
  const [filter, setFilter] = useState('');
```
it's redundant to specify primitive types

src/App.tsx
etojeDenys: image
still not fixed, create some enum for status values

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
+    setSelectedTodo(null);
+  };
+
+  const handleFilter = (status: string) => {
+    if (status === 'all') {
+      setFilter('all');
+    } else if (status === 'completed') {
+      setFilter('completed');
+    } else if (status === 'active') {
+      setFilter('active');
```
etojeDenys, line 36: and why do we need these conditions if we can just write `setFilter(status)`

etojeDenys (changes requested):

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,76 @@
...
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+import { useState } from 'react';
+import { Status } from '../../types/Status';
+
+type TodoFilterProps = {
+  onFilter: (status: string) => void,
```
```suggestion
  onFilter: (status: Status) => void,
```

etojeDenys (changes requested):

src/components/TodoFilter/TodoFilter.tsx
etojeDenys: fix it everywhere
image

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,76 @@
...
+type TodoFilterProps = {
+  onFilter: (status: Status) => void,
+  onTitleFilter: (title: string) => void,
+  onClearFilter: () => void,
+};
+
+export const TodoFilter
+= ({ onFilter, onTitleFilter, onClearFilter } : TodoFilterProps) => {
+  const [query, setQuery] = useState('');
+  const [filter, setFilter] = useState(Status.all);
```
etojeDenys, line 13: btw, why do you need the separate state for the filter here, if you can just pass the filter from the app.tsx as the prop

etojeDenys (changes requested):

PR #1854, opened 2023-09-17
feat: add list of todos

FrontKid:
[DEMO LINK](https://frontkid.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,47 @@
-/* eslint-disable max-len */
-import React from 'react';
+import {
+  FC,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+
+import { getTodos } from './api';
+
+import { TodoContext } from './context/ContextTodo';
+
+import { TodoFilter, getPreparedTodos } from './components/TodoFilter';
+import { TodoModal } from './components/TodoModal';
+import { TodoList } from './components/TodoList';
+import { Loader } from './components/Loader';
+
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
-import { TodoList } from './components/TodoList';
-import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
-import { Loader } from './components/Loader';
+export const App: FC = () => {
+  const {
+    setTodos,
+    isOpenModal,
+    todos,
+    inputField,
+    filteredBy,
+  } = useContext(TodoContext);
 
-export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const preparedTodos = useMemo(() => getPreparedTodos(todos, {
+    inputField,
+    filteredBy,
+  }), [inputField, filteredBy, todos]);
+
+  const isShowTodoList = Boolean(preparedTodos.length) && !isLoading;
+
   return (
     <>
       <div className="section">
@@ -21,14 +54,19 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+
+              {isShowTodoList && (
+                <TodoList todos={preparedTodos} />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {isOpenModal && (
+        <TodoModal />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,100 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import React, { useContext } from 'react';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+import { TodoContext } from '../../context/ContextTodo';
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+import { ETodoStatus, Todo } from '../../types/Todo';
+
+interface ITodosQuery {
+  inputField?: string;
+  filteredBy?: ETodoStatus;
+}
+
+export function getPreparedTodos(todos: Todo[], {
+  inputField = '',
+  filteredBy = ETodoStatus.ALL,
+}: ITodosQuery) {
+  let clonedTodos = [...todos];
+
+  if (inputField) {
+    const searchQueryLowerCase = inputField.toLowerCase().trim();
+
+    clonedTodos = clonedTodos.filter(({ title }) => (
+      title.toLowerCase().includes(searchQueryLowerCase)
+    ));
+  }
+
+  if (filteredBy !== ETodoStatus.ALL) {
+    switch (filteredBy) {
+      case ETodoStatus.Active:
+        clonedTodos = clonedTodos.filter(({ completed }) => !completed);
+        break;
+
+      case ETodoStatus.Completed:
+        clonedTodos = clonedTodos.filter(({ completed }) => completed);
+        break;
+
+      default:
+        throw new Error('Invalid select status');
+    }
+  }
+
+  return clonedTodos;
+}
+
+export const TodoFilter = () => {
+  const {
+    inputField,
+    setInputField,
+    setFilteredBy,
+    filteredBy,
+  } = useContext(TodoContext);
+
+  const handleSelectStatus = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setFilteredBy(event.currentTarget.value as ETodoStatus);
+  };
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filteredBy}
+            onChange={handleSelectStatus}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
+          </select>
+        </span>
+      </p>
+
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          placeholder="Search..."
+          value={inputField}
+          onChange={(event) => setInputField(event.target.value)}
         />
-      </span>
-    </p>
-  </form>
-);
+
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {inputField && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              aria-label="delete button"
+              className="delete"
+              onClick={() => setInputField('')}
+            />
+          </span>
+        )}
+      </p>
+    </form>
+  );
+};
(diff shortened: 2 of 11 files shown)


src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+import '@fortawesome/fontawesome-free/css/all.css';
+
+export const App: FC = () => {
+  const {
+    setTodos,
+    isOpenModal,
+    todos,
+    inputField,
+    filteredBy,
+  } = useContext(TodoContext);
```
davinnchii, line 28: Also here, you could create ur own hook like useTodos, so u won't need to type useContext(TodoContext) every time

src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
```
davinnchii, line 35: I think it'd better to add error handling here

src/App.tsx
```diff
@@ -21,14 +54,19 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+
+              {isShowTodoList && (
```
davinnchii, line 59: isTodoListVisible or isTodoListShown

src/App.tsx
```diff
@@ -21,14 +54,19 @@ export const App: React.FC = () => {
...
+                <TodoList todos={preparedTodos} />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {isOpenModal && (
```
davinnchii, line 67: isModalOpen

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,98 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filteredBy}
+            onChange={handleSelectStatus}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
davinnchii, line 65: You have enum for this value, render options using map and keys from enum
https://stackoverflow.com/questions/41308123/map-typescript-enum

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +24,7 @@ export const TodoList: React.FC = () => (
...
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => <TodoItem key={todo.id} todo={todo} />)}
```
davinnchii, line 27: IMHO

```suggestion
      {todos.map(todo => (
     <TodoItem key={todo.id} todo={todo} />
     ))}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +40,33 @@ export const TodoModal: React.FC = () => {
...
+              {`Todo #${selectedTodo?.id}`}
             </div>
 
-            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
             <button
               type="button"
               className="delete"
               data-cy="modal-close"
+              aria-label="delete button"
+              onClick={handCloseModal}
```
davinnchii, line 51: I think u misspelled word handle here

src/context/ContextTodo.tsx
```diff
@@ -0,0 +1,76 @@
...
+  setIsOpenModal: (newValue: boolean) => void;
+  setFilteredBy: (newValue: ETodoStatus) => void;
+  setInputField: (newValues: string) => void;
+  setTodos: (newValues: Todo[]) => void;
+}
+
+const initialContext: IContextInitial = {
+  todos: [],
+  inputField: '',
+  filteredBy: ETodoStatus.ALL,
```
davinnchii, line 31: You could use object for this 2 values, in case you want to shorten ur code

davinnchii (approved):
GJ, left some comments about naming and some possible changes if you'd like to do them

src/App.tsx
roman-mirzoian: Yes, that is a valid comment.

src/App.tsx
roman-mirzoian: Yes, that is a valid comment.

src/App.tsx
roman-mirzoian: Yes, that is a valid comment.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,98 @@
...
+  if (filteredBy !== ETodoStatus.ALL) {
+    switch (filteredBy) {
+      case ETodoStatus.Active:
+        return preparedTodos.filter(({ completed }) => !completed);
+
+      case ETodoStatus.Completed:
+        return preparedTodos.filter(({ completed }) => completed);
+
+      default:
+        throw new Error('Invalid select status');
```
roman-mirzoian, line 35:
```suggestion
        throw new Error('Invalid status selected.');
```

src/components/TodoFilter/TodoFilter.tsx
roman-mirzoian: Yes, that is a valid comment.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,98 @@
...
-    </p>
-  </form>
-);
+
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {inputField && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
roman-mirzoian, line 85: Using inline styles is a bad practice, use css instead.

roman-mirzoian (changes requested):
Nice job, but there is quite a bit of work to be done.
Good comments were made during the preliminary inspection, so you should pay attention to them.

src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
+export const App: FC = () => {
+  const {
+    setTodos,
+    isModalOpen,
+    todos,
+    inputField,
+    filteredBy,
+  } = useContext(TodoContext);
+
+  const [isLoading, setIsLoading] = useState(true);
```
Moroz-Dmytro, line 30: Better to set `false` as a default value and change it to `true` before API call

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const handleSelectTodo = (chosenTodo: Todo) => {
+    setSelectedTodo(chosenTodo);
+    setIsModalOpen(true);
+  };
+
+  return (
+    <tr
+      data-cy="todo"
+      className={cn({
+        'has-background-info-light': selectedTodo?.id === id,
```
Moroz-Dmytro, line 30: DRY. Create variable for this value `selectedTodo?.id === id` and reuse it here and in other places

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,37 @@
...
+
+  const {
+    setIsModalOpen,
+    selectedTodo,
+    setSelectedTodo,
+  } = useContext(TodoContext);
+
+  useEffect(() => {
+    getUser(selectedTodo?.userId ?? 0)
+      .then(user => setSelectedUser(user));
```
Moroz-Dmytro, line 22: Add `catch`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,37 @@
...
+  const [selectedUser, setSelectedUser] = useState<User | null>(null);
+
+  const {
+    setIsModalOpen,
+    selectedTodo,
+    setSelectedTodo,
+  } = useContext(TodoContext);
+
+  useEffect(() => {
+    getUser(selectedTodo?.userId ?? 0)
```
Moroz-Dmytro, line 21: You don't need to do useless API call if `userId` doesn't exist

src/context/ContextTodo.tsx
```diff
@@ -0,0 +1,76 @@
...
+  setTodos: () => { },
+};
+
+export const TodoContext = createContext(initialContext);
+
+export const ContextTodo: FC<TContextTodoProps> = ({ children }) => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [inputField, setInputField] = useState('');
+  const [filteredBy, setFilteredBy] = useState(ETodoStatus.ALL);
+  const [isModalOpen, setIsModalOpen] = useState(false);
```
Moroz-Dmytro, line 47: This state is useless, you can render modal if `selectedTodo` is not `null`

src/types/Todo.ts
```diff
@@ -4,3 +4,9 @@ export interface Todo {
   completed: boolean;
   userId: number;
 }
+
+export enum ETodoStatus {
+  ALL = 'all',
```
Moroz-Dmytro, line 9:
```suggestion
  All = 'all',
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,37 @@
...
+    setSelectedTodo(null);
+    setIsModalOpen(false);
+  };
+
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!selectedUser ? (
```
Moroz-Dmytro, line 34: Better to add `isLoading` state as in `App` component

Moroz-Dmytro (changes requested):

src/App.tsx
Moroz-Dmytro: Not fixed

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,101 @@
...
+
+import '../../utils/custom-string-extensions';
+import styles from './TodoFilter.module.scss';
+
+interface ITodosQuery {
+  inputField?: string;
+  filteredBy?: ETodoStatus;
+}
+
+export function getPreparedTodos(todos: Todo[], {
```
Moroz-Dmytro, line 15: Why this function is inside `TodoFilter` component
Better move it to separate file as a helper function

src/components/TodoModal/TodoModal.tsx
Moroz-Dmytro: Not fixed

Moroz-Dmytro (changes requested):
Don't leave comments not fixed and without any reply 😔
Also, check what is wrong with tests, they have failed

src/App.tsx
FrontKid: I don't think it matters when you setting loader in one place

src/components/TodoModal/TodoModal.tsx
FrontKid: If you could, please explain why it's better. For me it's understandable and it works

src/App.tsx
roman-mirzoian: @FrontKid This is a rule of good code - by default, exclude what we don't need always and include only when we have to use it.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,66 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filteredBy}
+            onChange={handleSelectStatus}
+          >
+            {Object.values(ETodoStatus).map(el => (
+              <option key={el} value={el}>{el.toCapitalize()}</option>
+            ))}
```
roman-mirzoian, line 33: Do not abbreviate names to unreadable ones
```suggestion
            {Object.values(ETodoStatus).map(status => (
              <option key={status} value={status}>{status.toCapitalize()}</option>
            ))}
```

src/components/TodoModal/TodoModal.tsx
roman-mirzoian: @FrontKid You should try to structure the code in such a way that the actions in it look logical.
When you do a loading, first of all you expect that it will complete, not that the user will choose.

src/utils/todoFilter.ts
```diff
@@ -0,0 +1,36 @@
...
+interface ITodosQuery {
+  inputField?: string;
+  filteredBy?: ETodoStatus;
+}
+
+export function getPreparedTodos(todos: Todo[], {
+  inputField = '',
+  filteredBy = ETodoStatus.All,
+}: ITodosQuery) {
+  let preparedTodos = [...todos];
```
roman-mirzoian, line 12: The `filter` method returns a copy of the array to which it is called, so you don't have to do it manually

roman-mirzoian (changes requested):
Almost done!

PR #1856, opened 2023-09-17
Added React_dynamic_list solution

vserhiichuk:
[DEMO LINK](https://vserhiichuk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,35 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getPreparedTodos } from './helpers/helpers';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [selectedCategory, setSelectedCategory] = useState('');
+  const [query, setQuery] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const preparedTodos = useMemo(() => {
+    return getPreparedTodos(todos, query, selectedCategory);
+  }, [todos, query, selectedCategory]);
+
   return (
     <>
       <div className="section">
@@ -17,18 +38,34 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                value={query}
+                handleInputQuery={setQuery}
+                onSelectedCategory={setSelectedCategory}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading
+                ? <Loader />
+                : (
+                  <TodoList
+                    todos={preparedTodos}
+                    onSelectedUser={setSelectedTodo}
+                    onSelectedTodo={selectedTodo}
+                  />
+                )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal
+          selectedTodo={selectedTodo}
+          removeModal={setSelectedTodo}
+        />
+      )}
     </>
   );
 };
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -17,18 +38,34 @@ export const App: React.FC = () => {
...
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading
+                ? <Loader />
+                : (
+                  <TodoList
+                    todos={preparedTodos}
+                    onSelectedUser={setSelectedTodo}
+                    onSelectedTodo={selectedTodo}
```
polinavafik, line 55: I think it would be better to name it `selectedTodo={selectedTodo}` because with `onSelectedTodo` it seems like we expect some function there instead of some value

src/helpers/helpers.ts
```diff
@@ -0,0 +1,35 @@
...
+
+      case 'completed':
+        filteredTodos = filteredTodos.filter(({ completed }) => completed);
+        break;
+
+      default:
+        filteredTodos = filteredTodos
+          .filter(({ title }) => filterTodosByQuery(title));
+    }
+  }
```
polinavafik, line 32: instead of using `'active' 'completed'` you could create some enum (store it in types folder) with all of the options and then reuse them everywhere in your code

(so we would know for sure that if we change 'completed' to 'done' for example - it will change everywhere)

polinavafik (approved):
Everything is working as expected! Great job🎀🤗🤝

but I do have some small suggestions😅

src/App.tsx
polinavafik: i think you used that way of naming somewhere else in your code btw

src/App.tsx
```diff
@@ -1,14 +1,35 @@
...
+  const [selectedCategory, setSelectedCategory] = useState('');
+  const [query, setQuery] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
```
Moroz-Dmytro, line 26: Add `catch` block for handling errors

src/App.tsx
Moroz-Dmytro: Good point, fix this

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,21 @@
-export const TodoFilter = () => (
+interface Props {
+  value: string,
+  handleInputQuery: (value: string) => void,
+  onSelectedCategory: (category: string) => void,
```
Moroz-Dmytro, line 4:
```suggestion
  onSelectCategory: (category: string) => void,
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,21 @@
-export const TodoFilter = () => (
+interface Props {
+  value: string,
```
Moroz-Dmytro, line 2: make up more descriptive name

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,86 @@
...
-);
+              {completed ? (
+                <td className="is-vcentered">
+                  <span className="icon" data-cy="iconCompleted">
+                    <i className="fas fa-check" />
+                  </span>
+                </td>
+              ) : (
+                <td className="is-vcentered" />
+              )}
```
Moroz-Dmytro, line 52: DRY. `<td className="is-vcentered">` is the same for both blocks

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
+import React, { useEffect, useState } from 'react';
+import cn from 'classnames';
 import { Loader } from '../Loader';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  selectedTodo: Todo,
+  removeModal: (value: null) => void;
```
Moroz-Dmytro, line 10: bad name. You are not removing modal, you close it

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
+  const {
+    title,
+    id,
+    userId,
+    completed,
+  } = selectedTodo;
+
+  useEffect(() => {
+    getUser(userId)
+      .then(setUser);
```
Moroz-Dmytro, line 24: Add `catch`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
+  }, [userId]);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!user ? (
```
Moroz-Dmytro, line 31: Better to use `isLoading` state for this, as you did it in `App` component

src/helpers/helpers.ts
Moroz-Dmytro: Good point, please fix this

Moroz-Dmytro (changes requested):

Moroz-Dmytro (approved):
Looks good 👍

PR #1861, opened 2023-09-18
Done

ZadorozhnyiYevhenii:
[DEMO LINK](https://ZadorozhnyiYevhenii.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { Select } from './types/Select';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
```
anastasiiavorobiova, line 16:
```suggestion
  const [isLoading, setIsLoading] = useState<boolean>(false);
```
Consider adding generic types to keep the code consistent

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+  const [errorMessage, setErrorMessage] = useState('');
```
anastasiiavorobiova, line 17: See my comment above

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+  const [query, setQuery] = useState('');
```
anastasiiavorobiova, line 18: The same issue here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,91 @@
...
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map(todo => {
+          return (
+            <>
```
anastasiiavorobiova, line 39: There is no need to use fragments here, you have the common parent element

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,91 @@
...
+              <tr
+                data-cy="todo"
+                className={classNames({
+                  'has-background-info-light': selectedTodo !== null
+                  && selectedTodo.id === todo.id,
+                })}
```
anastasiiavorobiova, line 45: add the key prop for rendering lists

anastasiiavorobiova (changes requested):

nazarmatsevych (approved):

PR #1862, opened 2023-09-18
Develop

Yevheniia-Sid:
[DEMO LINK](https://Yevheniia-Sid.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,34 +1,129 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos()
+      .then(todoFromServer => {
+        setTodos(todoFromServer);
+      })
+      .catch(() => setIsLoading(false))
+      .finally(() => setIsLoading(false));
```
kaavka, line 68: image
In catch method would be better to throw an error

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,59 @@
...
+  query,
+}) => {
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={event => {
+              handleStatusChange(event.target.value);
```
kaavka, line 21: You could use this logic right in the handle
Here we could just put `onChange={handleStatusChange}`
And in the handler
`  const handleStatusChange = (event) => {
    setStatus(event.target.value);
  };`

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,49 @@ export const TodoList: React.FC = () => (
...
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
```
kaavka, line 31: Could use destructure here.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,28 @@
...
+  closeModal: () => void;
+};
+
+export const TodoModal: React.FC<Props> = ({ todo, closeModal }) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    getUser(todo.userId)
+      .then(setUser)
+      .catch(() => setUser(null));
```
kaavka, line 18: So, as I understood, there will be infinite loading if we don't find any user. One possible solution is to create a new state for loading.

kaavka (approved):
Good job!

src/App.tsx
```diff
@@ -1,34 +1,131 @@
...
 
+const getFilterTodos = (
+  todos: Todo[],
+  statusFilter: string,
+  queryFilter: string,
+): Todo[] => {
+  let filteredTodos: Todo[] = [];
+
+  switch (statusFilter) {
+    case 'all': {
```
Moroz-Dmytro, line 21: You should create `enum` for all these values in `case`s

src/App.tsx
```diff
@@ -1,34 +1,131 @@
...
+      filteredTodos = todos;
+      break;
+    }
+
+    case 'active': {
+      filteredTodos = todos.filter(todo => todo.completed === false);
```
Moroz-Dmytro, line 27:
```suggestion
      filteredTodos = todos.filter(({ completed })=> !completed);
```

src/App.tsx
```diff
@@ -1,34 +1,131 @@
...
+
+      break;
+    }
+
+    case 'completed': {
+      filteredTodos = todos.filter(todo => todo.completed === true);
```
Moroz-Dmytro, line 33:
```suggestion
      filteredTodos = todos.filter(({ completed })=> completed);
```

src/App.tsx
```diff
@@ -1,34 +1,131 @@
...
+    return titleLower.includes(queryLower);
+  });
+
+  return preparedTodos;
+};
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [status, setStatus] = useState('all');
```
Moroz-Dmytro, line 58: use `enum` here

src/App.tsx
```diff
@@ -1,34 +1,131 @@
...
+  const [isLoading, setIsLoading] = useState(true);
```
Moroz-Dmytro, line 59: Better to set `false` as default value and change it to `true` only before you are making API call

src/App.tsx
```diff
@@ -1,34 +1,131 @@
...
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos()
+      .then(todoFromServer => {
+        setTodos(todoFromServer);
+      })
+      .catch(() => {
+        throw new Error('Can not loading todos');
```
Moroz-Dmytro, line 68: Better to print some message to console or to show message on UI

src/App.tsx
```diff
@@ -1,34 +1,131 @@
...
+    setQuery('');
+  };
+
+  const onSelectedTodo = (todo: Todo) => {
+    setSelectedTodo(todo);
+  };
+
+  const closeModal = () => {
+    setSelectedTodo(null);
+  };
```
Moroz-Dmytro, line 93: You can combine these functions

src/components/TodoFilter/TodoFilter.tsx
Moroz-Dmytro: Good point

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,59 @@
...
+            }}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
Moroz-Dmytro, line 26: Create `enum` for these values and use `map` for rendering them

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,49 @@ export const TodoList: React.FC = () => (
...
+        <tr
+          data-cy="todo"
+          key={todo.id}
+          className={classNames(
+            { 'has-background-info-light': selectedTodo?.id === todo.id },
```
Moroz-Dmytro, line 36: DRY. Create variable for this value `selectedTodo?.id === todo.id` and reuse it in all places where you use it

Moroz-Dmytro, line 36: Move these code `<tr>...</tr>` to separate `TodoItem` component

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,31 @@
...
+
+export const TodoModal: React.FC<Props> = ({ todo, closeModal }) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    getUser(todo.userId)
+      .then(setUser)
+      .catch(() => {
+        setUser(null);
+        throw new Error('Can not loading data');
```
Moroz-Dmytro, line 20: Same comment as above

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,31 @@
...
+      });
+  }, []);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!user ? (
```
Moroz-Dmytro, line 28: Better to have state for `isLoading`, as you did it in `App` component

Moroz-Dmytro (changes requested):

PR #1865, opened 2023-09-18
implemented loading and sorting data logic

OlhaMomot:
[DEMO LINK](https://OlhaMomot.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,21 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => setLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +30,36 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                setQuery={setQuery}
+                query={query}
+                setTodos={setTodos}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {loading && <Loader />}
+              {!loading
+                && (
+                  <TodoList
+                    todos={todos}
+                    selectedTodo={selectedTodo}
+                    setSelectedTodo={setSelectedTodo}
+                  />
+                )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo
+        && (
+          <TodoModal
+            id={selectedTodo.userId}
+            todo={selectedTodo}
+            setSelectedTodo={setSelectedTodo}
+          />
+        )}
     </>
   );
 };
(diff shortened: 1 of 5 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,94 @@
...
+      getTodos(false).then(setTodos);
+      setFilter(false);
+    } else if (selectedValue === 'completed') {
+      getTodos(true).then(setTodos);
+      setFilter(true);
+    } else {
+      getTodos().then(setTodos);
+      setFilter(undefined);
+    }
+  };
```
Vannnkof, line 30: I think it's worth trying to solve it in a slightly different way. In the App component, you called the getTodos function using useEffect only once during the initial rendering. This means you can filter the received array here instead of calling getTodos every time.

Vannnkof (commented):
Keep it up!

Vannnkof (approved):

src/App.tsx
```diff
@@ -1,14 +1,27 @@
...
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => setLoading(false));
```
Moroz-Dmytro, line 22: Add `catch` for handling errors

src/App.tsx
```diff
@@ -17,18 +30,36 @@ export const App: React.FC = () => {
...
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo
+        && (
+          <TodoModal
+            id={selectedTodo.userId}
+            todo={selectedTodo}
```
Moroz-Dmytro, line 59: Do we really need both of these props? Can we get `userId` from `todo` inside the component?

src/components/TodoFilter/TodoFilter.tsx
Moroz-Dmytro: Good point, it's not a good idea to make API calls if we don't really need this. Try to refactor your solution, to make API request only 1 time and then just filter response when you need it

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,94 @@
...
+            <option
+              value="active"
+            >
+              Active
+            </option>
+            <option
+              value="completed"
+            >
+              Completed
+            </option>
```
Moroz-Dmytro, line 58: Create `enum` for all these values, and use `map` here to render them

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,94 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {query
+            && (
```
Moroz-Dmytro, line 78:
```suggestion
          {query && (
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,83 @@
...
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+          return (
+            <tr
+              data-cy="todo"
+              className=""
```
Moroz-Dmytro, line 38: Why you need empty class

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,31 @@
...
+
+export const TodoModal: React.FC<Props> = ({ id, todo, setSelectedTodo }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    setLoading(true);
+    getUser(id)
+      .then(setUser)
+      .finally(() => setLoading(false));
```
Moroz-Dmytro, line 21: Add `catch`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,31 @@
...
+
+type Props = {
+  id: number,
+  todo: Todo,
+  setSelectedTodo: (todo: Todo | null) => void,
+};
+
+export const TodoModal: React.FC<Props> = ({ id, todo, setSelectedTodo }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [loading, setLoading] = useState(true);
```
Moroz-Dmytro, line 15: Same comments to loading state as above

Moroz-Dmytro (changes requested):
I found a bug 🐛 
1. Type something in search
2. Then select "Completed" 
3. Then clear search 
I see all available todos, but "Completed" status is still selected

image

PR #1881, opened 2023-09-19
Solution

oksana-vas:
[DEMO LINK](https://oksana-vas.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,72 @@
...
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span
+            className="icon is-right"
+            style={{ pointerEvents: 'all' }}
```
VitaliyBodnar, line 58: try to avoid using inline styles

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,91 @@
...
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+          const handleOpenModal = () => {
+            setSelectedTodo(todo);
+          };
```
VitaliyBodnar, line 37: move this function outside the template

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+      })
+      .finally(() => setLoading(false));
+  }, []);
+
+  function filterTodos(
+    items: Todo[],
+    status: Status,
+    queryValue: string,
+  ) {
+    let filtered = [...items];
```
VitaliyBodnar, line 34: I suggest you to using two variables, it will make you code cleaner 
 - filteredByName
 - filteredByStatus

VitaliyBodnar (changes requested):

src/App.tsx
oksana-vas: thank you for the suggestion! but as for me now it looks more complicated

VitaliyBondarenko1982 (approved):
Looks good

PR #1884, opened 2023-09-19
Solution

jakubMikolajczykTech:
[DEMO LINK](https://<your_account>.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useMemo, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+        case Status.Active:
+          return !todo.completed;
+
+        default:
+          return true;
+      }
+    })
+      .filter(todo => todo.title.toLocaleLowerCase().includes(lowerQuery));
+  }, [todos, query, status]);
+
```
domalewski, line 50: Array is relatively small, and the filtering operation is fast, you might not see a significant performance difference by using useMemo. 
In such cases, you can choose to omit it for the sake of simplicity.

domalewski (changes requested):
please fix demo link

choeqq (approved):
@jakubMikolajczykTech please make sure to deploy your solution correctly

PR #1889, opened 2023-09-19
Solution

ivannaromanovych:
https://ivannaromanovych.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,47 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Select } from './utils/Select';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filterOption, setFilterOption] = useState(Select.All);
+  const [query, setQuery] = useState('');
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        // eslint-disable-next-line no-console
+        console.log('setTodos error');
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
+
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (filterOption) {
+        case Select.Active: return !todo.completed;
+        case Select.Completed: return todo.completed;
+        case Select.All:
+        default: return true;
+      }
+    }).filter(todo => todo.title.toLowerCase().includes(query.toLowerCase().trim()));
+  }, [todos, filterOption, query]);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
+      switch (filterOption) {
+        case Select.Active: return !todo.completed;
+        case Select.Completed: return todo.completed;
+        case Select.All:
+        default: return true;
+      }
+    }).filter(todo => todo.title.toLowerCase().includes(query.toLowerCase().trim()));
```
roman-mirzoian, line 42: To make the code more readable, it's better to split this into two separate functions.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,64 @@
...
+  setQuery = () => { },
+  query,
+}) => {
+  const handleSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => (
+    setFilterOption(event.target.value as Select)
+  );
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => (
+    setQuery(event.target.value)
+  );
+  const handleButtonClick = () => setQuery('');
```
roman-mirzoian, line 20: What button?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,64 @@
...
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {!!query.length && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
roman-mirzoian, line 52: Try to rewrite the code so that you don't need to disable the linter.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,64 @@
...
-      </span>
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {!!query.length && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
roman-mirzoian, line 51: Using inline styles is a bad practice, better use css.

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,48 @@ export const TodoList: React.FC = () => (
...
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          data-cy="todo"
+          className=""
```
roman-mirzoian, line 34: Why do you need an empty className?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +39,35 @@ export const TodoModal: React.FC = () => {
               className="modal-card-title has-text-weight-medium"
               data-cy="modal-header"
             >
-              Todo #2
+              Todo #
+              {selectedTodo.id}
             </div>
 
             {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
roman-mirzoian, line 46: Try to rewrite the code so that you don't need to disable the linter.

roman-mirzoian (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+        case Select.Completed: return todo.completed;
+        case Select.All:
+        default: return true;
+      }
+    });
+  }, [todos, filterOption]);
+
+  const filterByQuery = useCallback((todosToFilter: Todo[]) => {
+    return todosToFilter.filter(todo => todo.title.toLowerCase().includes(query.toLowerCase().trim()));
+  }, [todos, query]);
```
BudnikOleksii, line 47: Better to move these functions outside the component and use them as helpers, `useCallback` is redundant in this case too

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
         />
-      </span>
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {!!query.length && (
```
BudnikOleksii, line 51:
```suggestion
        {query && (
```
`query` is a string, and an empty string is a falsy value, so you can do like this, just check if the query exists.
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,18 @@
...
+type Props = {
+  todos: Todo[],
+  selectedTodo: Todo | null,
+  onChange: (todo: Todo) => void,
+};
+
+export const TodoList: React.FC<Props> = ({
+  todos,
+  selectedTodo,
+  onChange = () => { },
```
BudnikOleksii, line 14: `onChange` is required filed, you don't to assign default parameter

BudnikOleksii (approved):
Good work 👍 
Approved, but check the comments

PR #1891, opened 2023-09-19
add task solution

ogospadzipalina:
[DEMO LINK](https://ogospadzipalina.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,66 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filterByTitle, setFilterByTitle] = useState<string>('');
+  const [filter, setFilter] = useState<string>('');
+
+  const selectTodos = (todo: Todo) => {
+    const findTodo = todos.find(value => value === todo) as Todo;
+
+    setSelectedTodo(findTodo);
+  };
+
+  const handleModalClose = () => {
+    setSelectedTodo(null);
+  };
+
+  const handleFilter = (status: string) => {
+    if (status === 'all') {
+      setFilter('all');
+    } else if (status === 'completed') {
+      setFilter('completed');
+    } else if (status === 'active') {
+      setFilter('active');
+    }
+  };
+
+  const handleTitleFilter = (title: string) => {
+    setFilterByTitle(title.toLowerCase());
+  };
+
+  const visibleTodos = todos.filter(todo => {
+    if (filter === 'all') {
+      return todo.title.toLowerCase().includes(filterByTitle);
+    }
+
+    if (filter === 'completed') {
+      return todo.completed && todo.title.toLowerCase().includes(filterByTitle);
+    }
+
+    if (filter === 'active') {
+      return !todo.completed && todo.title.toLowerCase().includes(filterByTitle);
+    }
+
+    return todo.title.toLowerCase().includes(filterByTitle);
+  });
+
+  const handleClearFilter = () => {
+    setFilterByTitle('');
+    setFilter('all');
+  };
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos().then(setTodos).finally(() => setIsLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filterByTitle, setFilterByTitle] = useState<string>('');
+  const [filter, setFilter] = useState<string>('');
```
choeqq, line 18: we want this string to be more specific - we should accept only 3 type of strings - all, active and completed

src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
+
+  const handleFilter = (status: string) => {
+    if (status === 'all') {
+      setFilter('all');
+    } else if (status === 'completed') {
+      setFilter('completed');
+    } else if (status === 'active') {
+      setFilter('active');
+    }
+  };
```
choeqq, line 38: this should be outside of component if possible

src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filterByTitle, setFilterByTitle] = useState<string>('');
+  const [filter, setFilter] = useState<string>('');
+
+  const selectTodos = (todo: Todo) => {
+    const findTodo = todos.find(value => value === todo) as Todo;
```
choeqq, line 21:
```suggestion
    const foundTodo = todos.find(value => value === todo) as Todo;
```

choeqq, line 21:
```suggestion
    const findTodo = todos.find(value => value === todo) as Todo;
```
there is something wrong with typing, you shouldn't have to use `as`

src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
+    if (filter === 'completed') {
+      return todo.completed && todo.title.toLowerCase().includes(filterByTitle);
+    }
+
+    if (filter === 'active') {
+      return !todo.completed && todo.title.toLowerCase().includes(filterByTitle);
+    }
+
+    return todo.title.toLowerCase().includes(filterByTitle);
+  });
```
choeqq, line 58: this should also be outside of function, also please try to combine handlerFilter with visibleTodos - those are basically doing the same thing

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,76 @@
...
+type TodoFilterProps = {
+  onFilter: (status: string) => void,
+  onTitleFilter: (title: string) => void,
+  onClearFilter: () => void,
+};
+
+export const TodoFilter
+  = ({ onFilter, onTitleFilter, onClearFilter } : TodoFilterProps) => {
+    const [query, setQuery] = useState<string>('');
+    const [filter, setFilter] = useState<string>('all');
```
choeqq, line 12: please be more specific with this type

src/components/TodoList/TodoList.tsx
choeqq, line 1: please take a close look at your solution, every single Todo item should take 100% of available space, when your is taking about a half, please make sure to fix it

choeqq (changes requested):

Meg-Sowka (approved):
looks good :)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,87 @@
...
+            placeholder="Search..."
+            value={query}
+            onChange={handleInput}
+          />
+          <span className="icon is-left">
+            <i className="fas fa-magnifying-glass" />
+          </span>
+
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
pawel-peksa, line 73: Please try to rewrite the code so that you will not need this line of code

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,87 @@
...
+            className="input"
+            placeholder="Search..."
+            value={query}
+            onChange={handleInput}
+          />
+          <span className="icon is-left">
+            <i className="fas fa-magnifying-glass" />
+          </span>
+
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
pawel-peksa, line 72: Let's avoid inline styling if there is no specific need to use it here.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,87 @@
...
+    const handleSelect: React.ChangeEventHandler<HTMLSelectElement>
+    = (event) => {
+      const selectedValue = event.target.value;
+      let filterValue: Filter = 'all';
+
+      if (selectedValue === 'active') {
+        filterValue = 'active';
+      } else if (selectedValue === 'completed') {
+        filterValue = 'completed';
+      }
```
pawel-peksa, line 24: Looks like unnecessary logic, try to assign `selectedValue` without `if` statements

src/components/TodoFilter/TodoFilter.tsx
pawel-peksa: +1 `filter` cannot be any `string`

pawel-peksa (changes requested):
Really good code, good job! You are almost done with this task! 💪

pawel-peksa (changes requested):
Good code quality! 🔥  Added couple of comments.

src/components/TodoFilter/TodoFilter.tsx
ogospadzipalina: > Please try to rewrite the code so that you will not need this line of code

do you mean line 73?

pawel-peksa: yes

src/components/TodoFilter/TodoFilter.tsx
ogospadzipalina: > Let's avoid inline styling if there is no specific need to use it here.

this one was inline styling in an intial code, not sure whether I have to move styling somewhere as there is no css/scss in the task files

choeqq (approved):

PR #1894, opened 2023-09-19
Develop

Dolnys:
[DEMO LINK](https://Dolnys.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,22 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import * as todosAPI from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+
+  useEffect(() => {
+    todosAPI.getTodos()
+      .then((data) => {
+        setTodos(data);
+      })
+      .catch((error) => {
+        console.error('Error:', error);
+      });
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,75 @@
...
+
+      if (filterStatus === 'active') {
+        return !todo.completed;
+      }
+
+      if (filterStatus === 'completed') {
+        return todo.completed;
+      }
+
+      return true;
```
choeqq, line 50: that's a good place to use switch

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,69 @@
...
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type TodoFilterProps = {
+  onFilterChange: (status: string) => void;
+  handleQuery: (value: string) => void;
+};
+export const TodoFilter: React.FC<TodoFilterProps> = ({
```
choeqq, line 7:
```suggestion
};

export const TodoFilter: React.FC<TodoFilterProps> = ({
```

src/App.tsx
```diff
@@ -1,14 +1,75 @@
...
+    })
+    .filter((todo) => {
```
choeqq, line 52: please consider moving this outside of component, as utility function - we want to keep our components as lean as possible

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,69 @@
...
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => {
+                setQuery('');
+                handleQuery('');
+              }}
```
choeqq, line 62: this should be a separate function

choeqq (changes requested):

src/App.tsx
Dolnys: Moved above the app component.

src/components/TodoFilter/TodoFilter.tsx
Dolnys: done

src/components/TodoFilter/TodoFilter.tsx
Dolnys: 👍

src/App.tsx
Dolnys: right, replaced with switch

JoyCoffeeAddict (approved):

PR #1913, opened 2023-09-22
solution

patryk-babon:
[DEMO LINK](https://patryk-babon.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,65 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoaded, setIsLoaded] = useState<boolean>(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [titleFilter, setTitleFilter] = useState<string>('');
+  const [filter, setFilter] = useState<string>('');
+
+  const handleSelectedTodo = (todo: Todo) => {
+    const findTodo = todos.find(value => value === todo) as Todo;
+
+    setSelectedTodo(findTodo);
+  };
+
+  const handleModalClose = () => {
+    setSelectedTodo(null);
+  };
+
+  const handleFilter = (status: string) => {
+    if (status === 'all') {
+      setFilter('all');
+    } else if (status === 'completed') {
+      setFilter('completed');
+    } else if (status === 'active') {
+      setFilter('active');
+    }
+  };
+
+  const handleTitleFilter = (title: string) => {
+    setTitleFilter(title.toLowerCase());
+  };
+
+  const visibleTodos = todos.filter(todo => {
+    if (filter === 'all') {
+      return todo.title.toLowerCase().includes(titleFilter);
+    }
+
+    if (filter === 'completed') {
+      return todo.completed && todo.title.toLowerCase().includes(titleFilter);
+    }
+
+    if (filter === 'active') {
+      return !todo.completed && todo.title.toLowerCase().includes(titleFilter);
+    }
+
+    return todo.title.toLowerCase().includes(titleFilter);
+  });
+
+  const handleClearFilter = () => {
+    setTitleFilter('');
+    setFilter('all');
+  };
+
+  useEffect(() => {
+    getTodos().then(setTodos).finally(() => setIsLoaded(true));
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +74,26 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                onFilter={handleFilter}
+                onTitleFilter={handleTitleFilter}
+                onClearFilter={handleClearFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!isLoaded && <Loader />}
+              <TodoList
+                todos={visibleTodos}
+                handleSelectedTodo={handleSelectedTodo}
+                selectedTodo={selectedTodo}
+              />
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo !== null && <TodoModal todo={selectedTodo} handleModalClose={handleModalClose} />}
     </>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,7 +11,7 @@ import { Todo } from './types/Todo';
 import { getTodos } from './api';
 
 export const App: React.FC = () => {
-  const [isLoaded, setIsLoaded] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
   const [todos, setTodos] = useState<Todo[]>([]);
   const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
   const [titleFilter, setTitleFilter] = useState<string>('');
@@ -28,13 +28,7 @@ export const App: React.FC = () => {
   };
 
   const handleFilter = (status: string) => {
-    if (status === 'all') {
-      setFilter('all');
-    } else if (status === 'completed') {
-      setFilter('completed');
-    } else if (status === 'active') {
-      setFilter('active');
-    }
+    setFilter(status);
   };
 
   const handleTitleFilter = (title: string) => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoaded, setIsLoaded] = useState<boolean>(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [titleFilter, setTitleFilter] = useState<string>('');
+  const [filter, setFilter] = useState<string>('');
```
choeqq, line 18: we want to be more specific with this type - please create union type for all the possible cases

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
+
+  const handleFilter = (status: string) => {
+    if (status === 'all') {
+      setFilter('all');
+    } else if (status === 'completed') {
+      setFilter('completed');
+    } else if (status === 'active') {
+      setFilter('active');
+    }
+  };
```
choeqq, line 38: this can be abstracted outside of the component - please use switch with finite number of cases and make sure you can only pick 3 types of filters

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
+    if (filter === 'completed') {
+      return todo.completed && todo.title.toLowerCase().includes(titleFilter);
+    }
+
+    if (filter === 'active') {
+      return !todo.completed && todo.title.toLowerCase().includes(titleFilter);
+    }
+
+    return todo.title.toLowerCase().includes(titleFilter);
+  });
```
choeqq, line 58: I think this can be combined with `handleFilter` function

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoaded, setIsLoaded] = useState<boolean>(false);
```
choeqq, line 14: I think `isLoading` state would be more useful - if todos are present the are basically loaded, if not - they are still loading - which means loading state

src/App.tsx
```diff
@@ -17,18 +74,26 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!isLoaded && <Loader />}
+              <TodoList
+                todos={visibleTodos}
+                handleSelectedTodo={handleSelectedTodo}
+                selectedTodo={selectedTodo}
+              />
```
choeqq, line 90: this should be conditional - if is loading display loader., if not - display list

src/App.tsx
```diff
@@ -17,18 +74,26 @@ export const App: React.FC = () => {
...
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo !== null && <TodoModal todo={selectedTodo} handleModalClose={handleModalClose} />}
```
choeqq, line 96:
```suggestion
      {selectedTodo && <TodoModal todo={selectedTodo} handleModalClose={handleModalClose} />}
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,74 @@
...
-  </form>
-);
+export const TodoFilter
+  = ({ onFilter, onTitleFilter, onClearFilter } : TodoFilterProps) => {
+    const [query, setQuery] = useState<string>('');
+    const [filter, setFilter] = useState<string>('all');
+    const handleSelect: React.ChangeEventHandler<HTMLSelectElement>
+    = (event) => {
+      setFilter(event.target.value);
+      onFilter(event.target.value);
```
choeqq, line 16: who do we need separate states for that?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,67 @@
...
-    </div>
-  );
+type TodoModalProps = {
+  todo: Todo,
+  handleModalClose: () => void;
 };
+
+export const TodoModal: React.FC<TodoModalProps>
+  = ({ todo, handleModalClose }) => {
+    const [isLoaded, setIsLoaded] = useState<boolean>(false);
```
choeqq, line 14: same as above ⬆️

choeqq (changes requested):

src/App.tsx
JoyCoffeeAddict: Implement previous suggestions.

src/App.tsx
```diff
@@ -1,14 +1,62 @@
...
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [titleFilter, setTitleFilter] = useState<string>('');
+  const [filter, setFilter] = useState<string>('');
+
+  const handleSelectedTodo = (todo: Todo) => {
+    const findTodo = todos.find(value => value === todo) as Todo;
```
JoyCoffeeAddict, line 21: This type assertion `as Todo` is incorrect. `Array.prototype.find()` can return found element or `undefined`, handle that  correctly instead of asserting `as Todo`

src/components/TodoFilter/TodoFilter.tsx
JoyCoffeeAddict: implement previous suggestions.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,74 @@
...
-        />
-      </span>
-    </p>
-  </form>
-);
+export const TodoFilter
+  = ({ onFilter, onTitleFilter, onClearFilter } : TodoFilterProps) => {
+    const [query, setQuery] = useState<string>('');
+    const [filter, setFilter] = useState<string>('all');
+    const handleSelect: React.ChangeEventHandler<HTMLSelectElement>
```
JoyCoffeeAddict, line 13: add empty space between states and methods.

src/App.tsx
```diff
@@ -1,14 +1,62 @@
...
+      case 'all':
+        return true;
+      case 'completed':
+        return todo.completed;
+      case 'active':
+        return !todo.completed;
+      default:
+        return true;
+    }
+  }).filter(todo => todo.title.toLowerCase().includes(titleFilter));
```
JoyCoffeeAddict, line 49: This would greatly fit into `useMemo`

src/App.tsx
```diff
@@ -1,14 +1,62 @@
...
+
+  const handleClearFilter = () => {
+    setTitleFilter('');
+    setFilter('all');
+  };
+
+  useEffect(() => {
+    getTodos().then(setTodos).finally(() => setIsLoading(true));
```
JoyCoffeeAddict, line 57: add `catch()` block

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,75 @@
...
-      </tr>
-    </tbody>
-  </table>
-);
+        <tbody>
+          {todos.map(todo => (
+            <tr
+              data-cy="todo"
+              className={cn({
+                'has-background-info-light': todo === selectedTodo,
```
JoyCoffeeAddict, line 33: you should compare `todo.id` not the whole object.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,75 @@
...
+                <button
+                  data-cy="selectButton"
+                  className="button"
+                  type="button"
+                  onClick={() => handleSelectedTodo(todo)}
+                >
+                  <span className="icon">
+                    <i className={cn('far', {
+                      'fa-eye': todo !== selectedTodo,
+                      'fa-eye-slash': todo === selectedTodo,
```
JoyCoffeeAddict, line 64: as above, compare properties not objects

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,67 @@
...
+  handleModalClose: () => void;
 };
+
+export const TodoModal: React.FC<TodoModalProps>
+  = ({ todo, handleModalClose }) => {
+    const [isLoading, setIsLoading] = useState<boolean>(false);
+    const [user, setUser] = useState<User | null>(null);
+
+    useEffect(() => {
+      getUser(todo.userId).then(setUser).finally(() => setIsLoading(true));
```
JoyCoffeeAddict, line 18: add `catch()`

JoyCoffeeAddict (changes requested):

src/App.tsx
```diff
@@ -10,17 +10,21 @@ import { Loader } from './components/Loader';
...
 
+type PossibleCases = boolean | Todo[] | Todo | null | string;
+
 export const App: React.FC = () => {
   const [isLoading, setIsLoading] = useState<boolean>(false);
   const [todos, setTodos] = useState<Todo[]>([]);
   const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
   const [titleFilter, setTitleFilter] = useState<string>('');
-  const [filter, setFilter] = useState<string>('');
+  const [filter, setFilter] = useState<PossibleCases>('');
```
mbruhler, line 20: I think this is still an issue. 
Filter should be an `enum` with possible filtering options. You can create this `enum` in `types/Todo.ts` and then import to `App.tsx` and to `TodoFilter.tsx`.
```
export const enum Status {
  ALL = 'all',
  ACTIVE = 'active',
  COMPLETED = 'completed',
}
```

You can access enum just like that: `Status.ALL`

You can read more about enums here: https://www.typescriptlang.org/docs/handbook/enums.html
Also think of an advantage of using `const enum` rather than `enum` :)

mbruhler (changes requested):
Good work on your side!
You are almost there, just couple of fixes and I think we can mark it as accepted

choeqq (approved):
LGTM ✅

PR #1915, opened 2023-09-25
Develop

sonyaxent:
https://sonyaxent.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,66 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
+import { Todo } from './types/Todo';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [allToodo, setAllToodo] = useState<Todo[]>([]);
+  const [filterToodo, setFilterToodo] = useState(allToodo);
+  const [visibleToodo, setVisibleToodo] = useState(filterToodo);
+  const [selectUserId, setSelectUserId] = useState(0);
+  const [selectTodoId, setSelectTodoId] = useState(0);
+  const [formItems, setFormItems] = useState({ statusSelect: 'all', input: '' });
+  const selectItems = (UserId: number, todoId: number) => {
+    setSelectUserId(UserId);
+    setSelectTodoId(todoId);
+  };
+
+  const hanldleChange = (key: string, value: string) => {
+    setFormItems(prev => ({
+      ...prev,
+      [key]: value,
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [allToodo, setAllToodo] = useState<Todo[]>([]);
+  const [filterToodo, setFilterToodo] = useState(allToodo);
+  const [visibleToodo, setVisibleToodo] = useState(filterToodo);
```
etojeDenys, line 16: i think, it can be just variables instead of states

src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
+    }
+  };
+
+  const handleSearch = () => {
+    setVisibleToodo(filterToodo.filter(todo => todo.title.toLocaleLowerCase()
+      .includes(formItems.input.toLocaleLowerCase())));
+  };
+
+  useEffect(() => {
+    getTodos().then(res => setAllToodo(res));
```
etojeDenys, line 53:
```suggestion
    getTodos().then(setAllToodo);
```
we can simplify it

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,23 @@
-export const TodoFilter = () => (
+import { FilterForm } from '../../types/FilterForm';
+
+type Props = {
+  valueFormItems: FilterForm,
+  hanldleChange: (key: string, value: string) => void,
```
```suggestion
  handleChange: (key: string, value: string) => void,
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
+  selectItems: (UserId: number, todoId: number) => void,
+};
+
+export const TodoModal: React.FC<Props> = ({
+  selectUser,
+  selectTodo,
+  selectItems,
+}) => {
+  const [user, setUser] = useState<undefined | User>();
+  const [todo, setTodo] = useState<undefined | Todo>();
```
etojeDenys, line 20:
```suggestion
  const [user, setUser] = useState<User | null>(null);
  const [todo, setTodo] = useState<Todo | null>(null);
```
better to set null as the default value

etojeDenys (changes requested):

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +37,37 @@ export const TodoModal: React.FC = () => {
...
           </header>
 
           <div className="modal-card-body">
             <p className="block" data-cy="modal-title">
-              quis ut nam facilis et officia qui
+              {todo?.title}
             </p>
 
             <p className="block" data-cy="modal-user">
               {/* <strong className="has-text-success">Done</strong> */}
```
lerastarynets, line 58: remove comments

lerastarynets (approved):
You've got this, keep pushing forward!

PR #1918, opened 2023-09-26
Develop

emil-owczarek:
[DEMO LINK](https://emil-owczarek.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { getTodos } from './api';
-import { Todo } from './types/Todo';
-import { Loader } from './components/Loader';
 import { TodoModal } from './components/TodoModal';
+import { Loader } from './components/Loader';
+import { getTodos /* getUser */ } from './api';
```
choeqq, line 10:
```suggestion
import { getTodos } from './api';
```

src/types/Filter.ts
```diff
@@ -0,0 +1 @@
+export type Filter = 'all' | 'active' | 'completed';
```
choeqq, line 1:
```suggestion
export type Filters = 'all' | 'active' | 'completed';
```
I think `Filters` would be little be more explicit - we have more than one filter, but that's only a suggestion

src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
+      if (filter === 'completed' && !todo.completed) {
+        return false;
+      }
+
+      if (!todo.title.toLowerCase().includes(searchTerm.toLowerCase())) {
+        return false;
+      }
+
+      return true;
+    });
```
choeqq, line 38: that would be a perfect usecase for switch - we're having finite number of cases here

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,72 @@
...
+  setSearchTerm: (searchTerm: string) => void;
+};
+
+export const TodoFilter: React.FC<TodoFilterProps> = ({
+  setFilter,
+  setSearchTerm,
+  searchTerm,
+}) => {
+  const [isSearchEmpty, setIsSearchEmpty] = useState(true);
+  const handleFilterChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
```
choeqq, line 16:
```suggestion
  const [isSearchEmpty, setIsSearchEmpty] = useState(true);
  
  const handleFilterChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,92 +1,80 @@
...
-                {selectedTodoId === id
-                  ? <i className="far fa-eye-slash" />
-                  : <i className="far fa-eye" />}
-              </span>
-            </button>
-          </td>
+  setSelectedTodoId,
+}) => {
+  return (
+    <table className="table is-narrow is-fullwidth">
```
choeqq, line 21:
```suggestion
}) => (
    <table className="table is-narrow is-fullwidth">
```

choeqq (changes requested):

src/App.tsx
emil-owczarek: Like that:
const visibleTodos = () => {
  const visible = [...todos].filter((todo) => {
    switch (filter) {
      case 'active':
        if (todo.completed) return false;
        break;

      case 'completed':
        if (!todo.completed) return false;
        break;
    }

    if (!todo.title.toLowerCase().includes(searchTerm.toLowerCase())) {
      return false;
    }

    return true;
  });

  return visible;
};

emil-owczarek: Can this be done better? Because in this form, it's less readable.

choeqq: It might seem that way, but I think it separates the filtering logic from a search phrase and if you were to add another filter type you would know where to put it right away. And when it comes to doing it better - I think that you could possibly do a loop through all the cases but I am not sure if it would be as readable as it is right now

emil-owczarek: Well, then maybe let's leave it at the way it is ;D

choeqq (approved):
LGTM ✅

PR #1941, opened 2023-10-14
add task solution

RomanOstrous:
[DEMO LINK](https://RomanOstrous.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [query, setQuery] = useState('');
+  const [filter, setFilter] = useState(Filter.All);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos().then(setTodos);
+  }, []);
+
+  const filterTodos: Todo[] = useMemo(() => {
```
TarasHoliuk, line 24:
```suggestion
  const filteredTodos: Todo[] = useMemo(() => {
```
in this variable stored array, it's not a function

TarasHoliuk: or visibleTodos, etc

src/App.tsx
```diff
@@ -17,18 +48,34 @@ export const App: React.FC = () => {
...
+                query={query}
+                sFilter={setFilter}
+                sQuery={setQuery}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!todos.length ? (
```
TarasHoliuk, line 59: it's better to create an additional state `isLoading` for it. What if in the server no todos (and we get an empty array in response) - we will show infinity loader

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,62 @@
...
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  query: string,
+  sFilter: (filterBy: Filter) => void,
+  sQuery: (query: string) => void,
```
TarasHoliuk, line 7: what is `s`? don't use short forms :slightly_smiling_face:

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,42 @@ export const TodoList: React.FC = () => (
...
+            {todo.completed && (
+              <span className="icon" data-cy="iconCompleted">
+                <i className="fas fa-check" />
+              </span>
+            )}
+          </td>
+          <td className="is-vcentered is-expanded">
+            {todo.completed
+              ? <p className="has-text-success">{todo.title}</p>
+              : <p className="has-text-danger">{todo.title}</p>}
```
TarasHoliuk, line 48:
```suggestion
            <p className={todo.completed ? "has-text-success" : "has-text-danger"}>
              {todo.title}
            </p>
```
it's better to change className (you may use classnames library) instead of the entire tag

TarasHoliuk (changes requested):
GJ :+1: 
Added a few minor comments

VitaliyBondarenko1982 (approved):
Well done

PR #1951, opened 2023-10-23
solution

natalia-maryshko:
[DEMO LINK](https://natalia-maryshko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -17,18 +59,34 @@ export const App: React.FC = () => {
...
-              <Loader />
-              <TodoList />
+              {loading && (<Loader />)}
+              {!loading && (
+                <TodoList
+                  todos={visibleTodos}
+                  selectedTodo={selectedTodo}
+                  setSelectedTodo={setSelectedTodo}
+                />
+              )}
```
anastasiia-tilikina, line 78:
```suggestion
              {loading 
              ?  (
                <Loader />
              ) : (
                <TodoList
                  todos={visibleTodos}
                  selectedTodo={selectedTodo}
                  setSelectedTodo={setSelectedTodo}
                />
              )}
```

src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
+  const [loading, setLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [filter, setFilter] = useState(Filter.All);
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(setTodos)
+      .finally(() => setLoading(false));
```
anastasiia-tilikina, line 25: it's always a good idea to add catch brunch in case of error

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  selectedTodo,
+  setSelectedTodo,
+}) => {
+  const [loading, setLoading] = useState(false);
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    setLoading(true);
+    getUser(selectedTodo.userId)
+      .then(setUser)
```
anastasiia-tilikina, line 22: the same here

src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
+
+      case Filter.Completed:
+        visible = todos.filter(todo => todo.completed);
+        break;
+
+      default:
+      case Filter.All:
+        visible = todos;
+        break;
+    }
```
anastasiia-tilikina, line 44: let's use switch case inside filter callback to avoid duplicates

anastasiia-tilikina: also, you don't need Filter.All case. it's enough to add default.

anastasiia-tilikina (changes requested):
Good job!

let's jus fix some minor issues.

src/App.tsx
```diff
@@ -1,14 +1,57 @@
...
+  const [query, setQuery] = useState('');
+  const [filter, setFilter] = useState(Filter.All);
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        throw Error('There are no todos');
+      })
```
SanyaBratashchuk, line 27: It isn't the best idea just to throw error. It would be much better to show some user-friendly notification in that case

SanyaBratashchuk (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,57 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
```
loralevitska, line 16: Rename all booleans according [tips](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig)
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```

loralevitska (approved):
Good job!

PR #1956, opened 2023-10-27
Solution

oksanatytanych:
[DEMO LINK](https://oksanatytanych.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,62 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React from 'react';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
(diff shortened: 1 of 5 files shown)


oksanatytanych:
The page looks like a working page but some tests have failed.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,62 @@
...
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  filterParam: string,
+  changeParams?: (event: React.ChangeEvent<HTMLSelectElement>) => void,
+  querry: string,
+  querryChange?: (event: React.ChangeEvent<HTMLInputElement>) => void,
+  deletequerry?: () => void,
```
VitaliyBondarenko1982, line 9: why some props not required?
looks like no cases when they are not passed in component

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,62 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filterParam}
+            onChange={changeParams}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
VitaliyBondarenko1982, line 30: suggest to make enum for all options values and use them in all places

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments and try to fix tests.
Each test case tells what behavior it is expecting.
For example search should be case insensitive
Screenshot from 2023-10-27 13-25-14
Screenshot from 2023-10-27 13-24-59

maxim2310 (approved):
Good work!!!

PR #1961, opened 2023-10-30
solution

OlegPopovych:
Hi,
[DEMO LINK](https://OlegPopovych.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,52 @@
...
+    setErrorMessage('');
+    setIsLoading(true);
+
+    getTodos()
+      .then(todos => {
+        setTodosFromServer(todos);
+      })
+      .catch(error => {
+        setErrorMessage(error.message);
+      })
```
sviatikkondur, line 32: 👍

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,30 @@
...
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
+          value={filter}
+          data-cy="statusSelect"
+          onChange={
+            (event: React.ChangeEvent<HTMLSelectElement>) => onSetFilter(
+              event.target.value,
+            )
+          }
```
sviatikkondur, line 26: I think that setting the event type is not necessary if you are working with it directly in the event handler

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,61 @@ export const TodoList: React.FC = () => (
...
+                type="button"
+                onClick={() => onSelectTodo(todo)}
+              >
+                <span className="icon">
+                  <i className={cn(
+                    'far',
+                    { 'far fa-eye-slash': selectedTodo?.id === id },
+                    { 'far fa-eye': selectedTodo?.id !== id },
+                  )}
+                  />
```
sviatikkondur, line 79: You should check this logic, because it sets `far` twice when apply one of classes below
image

sviatikkondur (changes requested):
Great job! Just fix that small issue with class name of to-do

src/components/TodoList/TodoList.tsx
OlegPopovych: Thanks.

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,61 @@ export const TodoList: React.FC = () => (
...
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => onSelectTodo(todo)}
+              >
+                <span className="icon">
+                  <i className={cn(
+                    'far',
+                    { 'fa-eye-slash': selectedTodo?.id === id },
```
StasSokolov, line 76: let's create variable like `isSelectedTodo` and reuse

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,83 @@
...
+
+            <div className="modal-card-body">
+              <p className="block" data-cy="modal-title">
+                {selectedTodo?.title}
+              </p>
+
+              <p className="block" data-cy="modal-user">
+                {selectedTodo?.completed
+                  ? <strong className="has-text-success">Done</strong>
+                  : <strong className="has-text-danger">Planned</strong>}
```
StasSokolov, line 70: let's use single `strong` tag and only change children and className

StasSokolov (changes requested):
Almost done, let's improve your code a bit more!

src/components/TodoModal/TodoModal.tsx
OlegPopovych: Thanks.

StasSokolov (approved):
Awesome job, i'm approving it🔥

PR #1967, opened 2023-10-31
made dynamic list of todos

sintax1s:
[DEMO LINK](https://sintax1s.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,29 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { SortType } from './types/SortType';
+import { getFilteredTodos } from './utils/getFilteredTodos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [inputValue, setInputValue] = useState('');
+  const [sortBy, setSortBy] = useState(SortType.All);
+  const [showModal, setShowModal] = useState(false);
+  const [modalTodo, setModalTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const visibleTodos = getFilteredTodos(todos, inputValue, sortBy);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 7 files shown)


OlhaKoziuk (approved):
Amazing work! I have nothing to add

src/App.tsx
```diff
@@ -1,14 +1,35 @@
...
+import { getTodos } from './api';
+import { SortType } from './types/SortType';
+import { getFilteredTodos } from './utils/getFilteredTodos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [inputValue, setInputValue] = useState('');
+  const [sortBy, setSortBy] = useState(SortType.All);
+  const [showModal, setShowModal] = useState(false);
```
polosanya, line 20: I thinks it's better to name it `isTodoSelected` or `isModalVisible`.
Btw, you can just check if the `modalTodo` is not null

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,49 @@
...
+          onChange={(event: React.ChangeEvent<HTMLSelectElement>) => {
+            setSortBy(event.target.value as SortType);
+          }}
+        >
+          <option
+            value="all"
+
+          >
+            All
+          </option>
```
polosanya, line 34:
```suggestion
          <option value="all">
            All
          </option>
```
Also, would be great to use Enum instead of the hardcoded values

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+
+type Props = {
+  todo?: Todo | null;
+  setShowModal: (value: boolean) => void;
+  setModalTodo: (value: Todo | null) => void;
+};
+
+export const TodoModal: React.FC<Props> = (
+  { todo, setShowModal, setModalTodo },
+) => {
```
polosanya, line 15:
```suggestion
export const TodoModal: React.FC<Props> = ({ 
  todo, 
  setShowModal, 
  setModalTodo, 
}) => {
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +42,34 @@ export const TodoModal: React.FC = () => {
...
           <div className="modal-card-body">
             <p className="block" data-cy="modal-title">
-              quis ut nam facilis et officia qui
+              {todo?.title}
             </p>
 
             <p className="block" data-cy="modal-user">
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              {todo?.completed ? (
```
polosanya, line 63: Todo must be required in this component.
Also, you can destructure all the needed props of it

polosanya (changes requested):
Good job! Here are a few comments about how you can improve your code:

src/components/TodoModal/TodoModal.tsx
sintax1s: I couldn't destructure the object because it might be null.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,92 @@
...
             <span className="icon">
-              <i className="far fa-eye" />
+              <i className="fas fa-check" />
             </span>
-          </button>
-        </td>
-      </tr>
+          </th>
+          <th>Title</th>
+          <th> </th>
```
StasSokolov, line 36:
```suggestion
          <th />
```

if i'm not mistaken it should work also

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,92 @@
...
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => handleClick(todo)}
+              >
+                <span className="icon">
+                  <i className={classNames('far',
+                    {
+                      'fa-eye-slash': modalTodo?.id === todo.id,
```
StasSokolov, line 80: Don't repeat yourself, will be better create separated variable like `isSelectedTodo` and reuse

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,46 @@
...
+          <option
+            value={SortType.Active}
+          >
+            Active
+          </option>
+          <option
+            value={SortType.Completed}
+          >
+            Completed
+          </option>
```
StasSokolov, line 43: not critical, but will be better use `.map` for show multiple `option`

StasSokolov (changes requested):

src/components/TodoList/TodoList.tsx
sintax1s: image
image
Without aria-label doesn't work,so i added it.But in my opinion this change is useless.

StasSokolov (approved):
Looks really good🔥

PR #1969, opened 2023-11-01
add tast solution

mpohorenyi:
The block of tests related to filtering and search fails ([SCREENSHOT](https://prnt.sc/-ioARwzLXb7Q)). 
My solution works exactly the same as in the demo, I don't understand why it fails the tests.

[DEMO LINK](https://mate-academy.github.io/react_dynamic-list-of-todos/)
[MY SOLUTION LINK](https://mpohorenyi.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,80 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { useContext, useState } from 'react';
+import { DispatchContext } from '../../Store';
+import { Filter } from '../../types/Filter';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+export const TodoFilter = () => {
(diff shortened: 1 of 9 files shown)


shchipina (approved):
I've seen your code and how it works, and it looks very similar to the working version to me, so I approve it and hope that the mentor will help you find the right solution😉
Good job))

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,80 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filter}
+            onChange={handleSelect}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
StasSokolov, line 48: not critical but will be better use `.map()` for show multiple `option`, also please create `enum` for status of each option

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
 import { Loader } from '../Loader';
+import { DispatchContext, StateContext } from '../../Store';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
 
 export const TodoModal: React.FC = () => {
+  const { todoModal } = useContext(StateContext);
+  const dispatch = useContext(DispatchContext);
+  const [user, setUser] = useState<User>();
+  const [loading, setLoading] = useState(false);
```
StasSokolov, line 12: not critical but `isLoading`as for me will be more informative

StasSokolov (approved):
Looks good to me, it's really good that you use `context`, i left only few minor comments and approving it, but let's check tests on QnA🔥

mpohorenyi:
> Looks good to me, it's really good that you use `context`, i left only few minor comments and approving it, but let's check tests on QnA🔥

I've already been to two Q&A sessions, but they still couldn't solve the problem with the tests.)

PR #1970, opened 2023-11-01
first solution, two tests didn't pass

OlhaKoziuk:
[DEMO LINK](https://OlhaKoziuk.github.io/react_dynamic-list-of-todos/)
Two tests are failing, but on Q&A, we decided to create a pull request. I would appreciate advice and comments
tests

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,32 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loader, setLoader] = useState(false);
+  const [modal, setModal] = useState(false);
+  const [id, setId] = useState(0);
+  const [userId, setUserId] = useState(0);
+
+  useEffect(() => {
+    setLoader(true);
+
+    getTodos()
+      .then((todosData) => {
+        setTodos(todosData);
+      })
+      .catch(() => {})
+      .finally(() => setLoader(false));
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +35,38 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                setTodos={setTodos}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {loader && !modal && <Loader />}
+              {!loader && todos.length > 0 && (
+                <TodoList
+                  todos={todos}
+                  setModal={setModal}
+                  id={id}
+                  setId={setId}
+                  setLoader={setLoader}
+                  setUserId={setUserId}
+                />
+              )}
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {modal && (
+        <TodoModal
+          todos={todos}
+          setModal={setModal}
+          id={id}
+          setId={setId}
+          loader={loader}
+          setLoader={setLoader}
+          userId={userId}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,158 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import React, { useState, useCallback } from 'react';
+import { getTodos, getFilteredTodos } from '../../api';
+import { Todo } from '../../types/Todo';
+import { Selected } from '../../types/Enum';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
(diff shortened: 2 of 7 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,158 @@
...
+  //   handleFilterChange(selectedOption, searchValue);
+  // };
+
+  // const handleSelectChange = async (
+  //   event: React.ChangeEvent<HTMLSelectElement>,
+  // ) => {
+  //   const selectedValue = event.target.value;
+
+  //   handleFilterChange(selectedValue, value);
+  // };
```
adioskv, line 65: Remove unnecessary comments 😉

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,158 @@
...
+      }
+
+      setTodos(filteredTodos);
+    },
+    [getIncludesTodos, setTodos],
+  );
+
+  const handleInputChange = useCallback((e: React
+    .ChangeEvent<HTMLInputElement>) => {
+    const searchValue = e.target.value;
```
adioskv, line 98: I think we all know what is `e` means, but try to use more meaningful names like `event`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,158 @@
...
+
+export const TodoFilter: React.FC<Props> = ({
+  setTodos,
+}) => {
+  const [selectedOption, setSelectedOption] = useState('all');
+  const [value, setValue] = useState('');
+
+  const getIncludesTodos = useCallback(
+    (val: string, currTodos: Todo[]) => {
+      return currTodos.filter((todo) => todo.title.includes(val));
```
adioskv, line 18: I found the problem!

EXAMPLE:
image

image

Our query should ignore cases, so that's the problem!

adioskv (changes requested):
### **Amazing job 😍!** 
**Just fix some moments**

(Described them in comments to your code)

src/components/TodoFilter/TodoFilter.tsx
OlhaKoziuk: Thanks, try to do)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,118 @@
...
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable jsx-a11y/control-has-associated-label */}
+          {value && (
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={async () => {
+                setValue('');
+                handleFilterChange(selectedOption, '');
```
vasylmiskiv, line 109: Move it function

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,94 @@
...
+          <th>Title</th>
+          <th> </th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {todos.map(todo => (
+          <tr
+            data-cy="todo"
+            className=""
```
vasylmiskiv, line 58: Why is it empty

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,44 +1,89 @@
...
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    setLoader(true);
+
+    getUser(userId)
+      .then((userData) => {
+        setUser(userData);
+      })
+      .catch(() => {})
```
vasylmiskiv, line 35: Would be nice to add an error instead of an empty function

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,44 +1,89 @@
...
+      .finally(() => setLoader(false));
+  }, [setLoader, userId]);
+
+  const handleCloseModal = useCallback(() => {
+    setModal(false);
+    setId(0);
+  }, [setModal, setId]);
+
+  const getTodoById = useCallback((todoId: number) => {
```
vasylmiskiv, line 44: They're a lot of unnecessary useCallback in the code. If the function doesn't pass to another component like a prop it will be not necessary to write it. Also it decreases readability

vasylmiskiv (changes requested):

src/App.tsx
```diff
@@ -1,34 +1,103 @@
...
+import { Selected } from './types/Selected';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loader, setLoader] = useState(true);
+  const [modal, setModal] = useState(false);
+  const [id, setId] = useState(0);
+  const [userId, setUserId] = useState(0);
+  const [error, setError] = useState('');
+  const [selectedOption, setSelectedOption] = useState('all');
```
polosanya, line 20: Use Enum instead of the hardcoded value. You can just combine it with the default case

src/App.tsx
```diff
@@ -1,34 +1,103 @@
...
+
+        default:
+          return true;
+      }
+    })
+      .filter(todo => {
+        const preparedQuery = query.toLowerCase();
+
+        return todo.title.toLocaleLowerCase().includes(preparedQuery);
+      })
```
polosanya, line 49: Do you need to call filter twice? I think it's better to make everything in one

src/App.tsx
```diff
@@ -1,34 +1,103 @@
...
-import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api/api';
+import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
+import { Selected } from './types/Selected';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loader, setLoader] = useState(true);
```
polosanya, line 15:
```suggestion
  const [isLoading, setIsLoading] = useState(true);
```
image

src/App.tsx
```diff
@@ -1,34 +1,103 @@
...
+  const [modal, setModal] = useState(false);
```
polosanya, line 16: Same here

src/App.tsx
```diff
@@ -1,34 +1,103 @@
...
+      {modal && (
+        <TodoModal
+          todos={filteredTodos}
+          setModal={setModal}
+          id={id}
+          setId={setId}
+          loader={loader}
+          setLoader={setLoader}
+          userId={userId}
+          onError={setError}
```
polosanya, line 98: I'm not sure that you need all these props inside

polosanya (approved):
Nice! I'm approving it, but please pay attention to the comments)

PR #1975, opened 2023-11-01
task solution

dolphinnq:
[DEMO LINK](https://dolphinnq.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,91 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { useEffect, useState } from 'react';
+import { Todo } from '../../types/Todo';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+function filterTodos(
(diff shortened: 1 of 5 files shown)


tetyana-pol (approved):
good job

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,91 @@
...
+    setFilteredTodos(filteredTodos);
+  }, [todos, filterBy, query, setFilteredTodos]);
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={event => setFilterBy(event.target.value)}
```
vasylmiskiv, line 55: Would be better to use separate handle function instead of changing the state directly

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,91 @@
...
+      </p>
+
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          placeholder="Search..."
+          value={query}
+          onChange={event => setQuery(event.target.value)}
```
vasylmiskiv, line 71: Would be better to use separate handle function instead of changing the state directly

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,91 @@
...
+          <i className="fas fa-magnifying-glass" />
+        </span>
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => setQuery('')}
```
vasylmiskiv, line 83: Would be better to use separate handle function instead of changing the state directly

vasylmiskiv (approved):
Everything looks nice

PR #1978, opened 2023-11-01
Develop

HardCodWorker:
[DEMO LINK](https://HardCodWorker.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,47 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [loadOn, setLoadOn] = useState(false);
+  const [todos, setToDos] = useState<Todo[]>([]);
+  const [modalOn, setmodalOn] = useState(false);
+  const [modalToDo, setmodalTodo] = useState<Todo | null>(null);
+  const [sortType, setSortType] = useState<string>('');
+  const [query, setQuery] = useState<string>('');
+
+  useEffect(() => {
+    setLoadOn(true);
+    getTodos().then(todoos => setToDos(todoos)).finally(() => setLoadOn(false));
+  }, []);
+
+  function preparedTodos() {
+    let displayTodos = [...todos];
+
+    if (query) {
+      displayTodos = displayTodos.filter(
+        todo => todo.title.toLowerCase()
+          .includes(query.toLowerCase().trim()),
+      );
+    }
+
+    switch (sortType) {
+      case 'active':
+        return displayTodos.filter(todo => !todo.completed);
+
+      case 'completed':
+        return displayTodos.filter(todo => todo.completed);
+
+      case 'all':
+      default:
+        break;
+    }
+
+    return displayTodos;
+  }
+
   return (
     <>
       <div className="section">
@@ -17,18 +56,36 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                sortType={sortType}
+                setSortType={setSortType}
+                query={query}
+                setQuery={setQuery}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {loadOn ? <Loader />
+                : (
+                  <TodoList
+                    todos={preparedTodos()}
+                    setmodalOn={setmodalOn}
+                    setmodalTodo={setmodalTodo}
+                    modalToDo={modalToDo}
+                  />
+                )}
             </div>
           </div>
         </div>
       </div>
+      {modalOn && (
+        <TodoModal
+          modalToDo={modalToDo}
+          setmodalOn={setmodalOn}
+          setmodalTodo={setmodalTodo}
+        />
+      )}
 
-      <TodoModal />
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,60 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+interface Props {
+  sortType: string,
+  setSortType: (value: string) => void,
+  query: string,
+  setQuery: (value: string) => void,
+}
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+export const TodoFilter: React.FC<Props> = ({
+  sortType,
+  setSortType,
+  query,
+  setQuery,
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+}) => {
+  return (
+
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [loadOn, setLoadOn] = useState(false);
+  const [todos, setToDos] = useState<Todo[]>([]);
+  const [modalOn, setmodalOn] = useState(false);
+  const [modalToDo, setmodalTodo] = useState<Todo | null>(null);
+  const [sortType, setSortType] = useState<string>('');
+  const [query, setQuery] = useState<string>('');
```
Ayogoharo, line 19: You don't need to make these variables in the "app" component and drill it into the filter. Just create it in the filter component

src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+      case 'active':
+        return displayTodos.filter(todo => !todo.completed);
+
+      case 'completed':
+        return displayTodos.filter(todo => todo.completed);
+
+      case 'all':
+      default:
+        break;
+    }
```
Ayogoharo, line 46: you dont need to add an "all" case, it has the same logic as the default, and you don't even need default, because It doesn't do anything

Ayogoharo, line 46: This whole function is a part of "TodoFilter" component

src/components/TodoList/TodoList.tsx
```diff
@@ -16,7 +31,53 @@ export const TodoList: React.FC = () => (
     </thead>
 
     <tbody>
-      <tr data-cy="todo" className="">
+      {
+        todos.map((todo) => (
+          <tr data-cy="todo" className={cn({ 'has-background-info-light': modalToDo?.id === todo.id })}>
+            <td className="is-vcentered">{todo.id}</td>
+            <td className="is-vcentered">
```
Ayogoharo: You need to add "key" value, if using map

src/components/TodoList/TodoList.tsx
```diff
@@ -94,7 +155,7 @@ export const TodoList: React.FC = () => (
             </span>
           </button>
         </td>
-      </tr>
+      </tr> */}
```
Ayogoharo: What is this for?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
+interface Props {
+
+  modalToDo: Todo | null,
+  setmodalOn: (value: boolean) => void
+  setmodalTodo: (value: null) => void
+}
+
+export const TodoModal: React.FC<Props> = ({ modalToDo, setmodalOn, setmodalTodo }) => {
+  const [showLoader, setshowLoader] = useState(false);
+  const [currentUser, setCurrentUser] = useState<User>();
```
Ayogoharo, line 17: `| null`

Ayogoharo (approved):

src/App.tsx
polosanya: I think it's not critical as we need to pass the filtered todos to TodosList component

src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+  const [modalToDo, setmodalTodo] = useState<Todo | null>(null);
+  const [sortType, setSortType] = useState<string>('');
+  const [query, setQuery] = useState<string>('');
+
+  useEffect(() => {
+    setLoadOn(true);
+    getTodos().then(todoos => setToDos(todoos)).finally(() => setLoadOn(false));
+  }, []);
+
+  function preparedTodos() {
```
polosanya, line 26:
```suggestion
  function getPreparedTodos() {
```
Let's create a variable `preparedTodos` with the result of this function

src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+
+    if (query) {
+      displayTodos = displayTodos.filter(
+        todo => todo.title.toLowerCase()
+          .includes(query.toLowerCase().trim()),
+      );
+    }
+
+    switch (sortType) {
+      case 'active':
```
polosanya, line 37: Better use Enum instead of the hardcoded values

src/App.tsx
polosanya: Default case is always required for error handling. But yes, you can combine it with "All" case

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,60 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={sortType}
+            onChange={(event) => setSortType(event.target.value)}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
polosanya, line 27: Same here. Use Enum instead of the hardcoded values

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +37,40 @@ export const TodoModal: React.FC = () => {
...
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+
+              {
+                modalToDo?.completed
+
+                  ? (<strong className="has-text-success">Done</strong>)
+
+                  : (<strong className="has-text-danger">Planned</strong>)
+              }
```
polosanya, line 68: Fix formatting

polosanya (changes requested):
Nice! Let's improve your code a bit)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,62 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={sortType}
+            onChange={(event) => setSortType(event.target.value)}
+          >
+            <option value="all">{Sorttype.all}</option>
+            <option value="active">{Sorttype.active}</option>
+            <option value="completed">{Sorttype.completed}</option>
```
StasSokolov, line 29: will be better use `.map()` and iterate `option`

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +31,56 @@ export const TodoList: React.FC = () => (
...
+                type="button"
+                onClick={() => {
+                  setmodalOn(true);
+                  setmodalTodo(todo);
+                }}
+              >
+                <span className="icon">
+                  <i className={cn('far',
+                    {
+                      'fa-eye-slash': modalToDo?.id === todo.id,
```
StasSokolov, line 74: Don't repeat yourself, move into variable and reuse like `isSelectedToDo`

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +31,56 @@ export const TodoList: React.FC = () => (
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {
+        todos.map((todo) => (
+          <tr
+            data-cy="todo"
+            className={cn({ 'has-background-info-light': modalToDo?.id === todo.id })}
```
StasSokolov, line 38: reuse here please

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +31,56 @@ export const TodoList: React.FC = () => (
...
+                onClick={() => {
+                  setmodalOn(true);
+                  setmodalTodo(todo);
+                }}
+              >
+                <span className="icon">
+                  <i className={cn('far',
+                    {
+                      'fa-eye-slash': modalToDo?.id === todo.id,
+                      'fa-eye': modalToDo?.id !== todo.id,
```
StasSokolov, line 75: reuse here please

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +37,35 @@ export const TodoModal: React.FC = () => {
...
             </p>
 
             <p className="block" data-cy="modal-user">
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
-
+              {
+                modalToDo?.completed
+                  ? (<strong className="has-text-success">Done</strong>)
+                  : (<strong className="has-text-danger">Planned</strong>)
```
StasSokolov, line 64: let's use single tag `strong` and only change className and content

src/types/Sorttype.ts
```diff
@@ -0,0 +1,6 @@
+export enum Sorttype {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
+
```
StasSokolov, line 5: remove redundant space please

StasSokolov (changes requested):

src/components/TodoFilter/TodoFilter.tsx
HardCodWorker: Could you please specify this moment, as I vaguely get what should be done

HardCodWorker:
All the notes were implemented

src/components/TodoFilter/TodoFilter.tsx
StasSokolov: it's not critical but you can iterate by `enum` SortType

StasSokolov (approved):
Overall look good to me🔥

PR #1979, opened 2023-11-01
task solution

YePolishchuk:
[DEMO LINK](https://YePolishchuk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,57 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { SortBy } from './types/SortBy';
+
+const sortTodosBy = (
+  todos: Todo[],
+  query: string,
+  sortType: SortBy,
+) => {
+  let sortedTodos = [...todos];
+
+  if (query.trim()) {
+    sortedTodos = (todos.filter(
+      todo => todo.title.toLocaleLowerCase()
+        .includes(query.trim().toLocaleLowerCase()),
+    ));
+  }
+
+  switch (sortType) {
+    case SortBy.Active:
+      sortedTodos = sortedTodos.filter(todo => !todo.completed);
+      break;
+    case SortBy.Completed:
+      sortedTodos = sortedTodos.filter(todo => todo.completed);
+      break;
+    case SortBy.All:
+    default:
+      break;
+  }
+
+  return sortedTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loader, setLoader] = useState(false);
+  const [visibleModal, setVisibleModal] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
+  const [sortBy, setSortBy] = useState(SortBy.All);
+
+  const selectedTodos = sortTodosBy(todos, query, sortBy);
+
+  useEffect(() => {
+    setLoader(true);
+
+    getTodos()
+      .then((currentTodos) => setTodos(currentTodos))
+      .finally(() => setLoader(false));
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+  todos: Todo[],
+  query: string,
+  sortType: SortBy,
+) => {
+  let sortedTodos = [...todos];
+
+  if (query.trim()) {
+    sortedTodos = (todos.filter(
+      todo => todo.title.toLocaleLowerCase()
+        .includes(query.trim().toLocaleLowerCase()),
```
ivaZaiets, line 24: You can create new variable and write this logic there. Smth like 'normalizeQuery' with method trim() and toLowerCase() or toLocaleLowerCase

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+  const [query, setQuery] = useState('');
+  const [sortBy, setSortBy] = useState(SortBy.All);
+
+  const selectedTodos = sortTodosBy(todos, query, sortBy);
+
+  useEffect(() => {
+    setLoader(true);
+
+    getTodos()
+      .then((currentTodos) => setTodos(currentTodos))
```
ivaZaiets, line 57: You can write just 
```js
.then(setTodos)
```

src/App.tsx
```diff
@@ -17,18 +66,36 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                query={query}
+                setQuery={setQuery}
+                sortBy={sortBy}
+                setSortBy={setSortBy}
```
ivaZaiets, line 73: Check point 3 in checklist, you should use correct naming for function with key word 'on'

src/App.tsx
```diff
@@ -17,18 +66,36 @@ export const App: React.FC = () => {
...
+                    setVisibleModal={setVisibleModal}
+                  />
+                )}
             </div>
           </div>
         </div>
       </div>
-
-      <TodoModal />
+      {visibleModal && (
```
ivaZaiets, line 92: You can use only ```selectedTodo``` instead ```visibleModal```

src/components/TodoList/TodoList.tsx
```diff
@@ -14,87 +28,56 @@ export const TodoList: React.FC = () => (
...
+            >
+              {todo.title}
+            </p>
+          </td>
+          <td className="has-text-right is-vcentered">
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => {
```
ivaZaiets, line 64: You can use only ```setSelectedTodo``` without ```setVisibleModal```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { Todo } from '../../types/Todo';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
+
+interface Props {
+  selectedTodo: Todo | null,
+  setSelectedTodo: (v: Todo | null) => void,
```
ivaZaiets, line 9: I think ```v``` not good naming, use smth like ```todo``` in the first variant or ```value``` in the second

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +37,34 @@ export const TodoModal: React.FC = () => {
...
+              {`Todo #${selectedTodo?.id}`}
             </div>
 
             {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
             <button
               type="button"
               className="delete"
               data-cy="modal-close"
+              onClick={() => {
+                setSelectedTodo(null);
```
ivaZaiets, line 49: The same about ```setVisibleModal``` here

ivaZaiets (approved):
Good job! Demo look so good!  If you want, you can improve your code a bit more. Good luck!

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+    default:
+      break;
+  }
+
+  return sortedTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loader, setLoader] = useState(false);
```
vasylmiskiv, line 45:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,28 @@
-export const TodoFilter = () => (
+import { SortBy } from '../../types/SortBy';
+
+interface Props {
+  query: string,
+  setQuery: (v: string) => void,
```
vasylmiskiv, line 5: Should write more appropriate name instead of v

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
+  selectedTodo,
+  setSelectedTodo,
+  setVisibleModal,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    if (selectedTodo) {
+      getUser(selectedTodo.userId)
+        .then((currentUser) => setUser(currentUser));
```
vasylmiskiv, line 23:
```suggestion
        .then((user) => setUser(user));
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
+    }
+  }, [selectedTodo]);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!user ? (
```
vasylmiskiv, line 31: It will be showing the loader even though it receives an error instead of the user. Would be better to create a separate loading state and set the loader value as a false in [finally method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally)

src/App.tsx
vasylmiskiv, line 57:
```suggestion
      .then((todos) => setTodos(todos))
```
or
```suggestion
      .then(setTodos)
```

PR #1980, opened 2023-11-01
add solution dynamic list of todos

Kuper18:
Hi, here is my solotion

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


koros-rk (approved):
Good Job ;)

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+import { Todo } from './types/Todo';
+import { Option } from './types/Option';
+
+interface FilterOptions {
+  select: string;
+  query: string;
+}
+
+function filterBy(todos: Todo[], { select, query }: FilterOptions): Todo[] {
+  let filtered = todos;
```
vasylmiskiv, line 19: You should add a copy of the array instead of defining a reference to the array
```suggestion
  let filteredTodos = [...todos];
```

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+  const [post, setPost] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setIsLoading(false);
+    }, 300);
+
+    getTodos()
+      .then(setTodos)
+      .catch();
```
vasylmiskiv, line 62: You can pass the error and show it

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+  const [select, setSelect] = useState('');
+  const [query, setQuery] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [post, setPost] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setIsLoading(false);
+    }, 300);
```
vasylmiskiv, line 58: Instead of it should set loading as a false after catch method in [finally method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally)

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+}
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [isVisibleModal, setIsVisibleModal] = useState(false);
+  const [select, setSelect] = useState('');
+  const [query, setQuery] = useState('');
+  const [userId, setUserId] = useState(0);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [post, setPost] = useState<Todo | null>(null);
```
vasylmiskiv, line 53:
```suggestion
  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,68 @@
...
+  setQuery,
+}) => {
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            value={select}
+            data-cy="statusSelect"
+            onChange={event => setSelect(event.target.value)}
```
vasylmiskiv, line 24: Would be better to add a handle select function instead changing the state direcly

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,114 @@
...
-            </span>
-          </button>
-        </td>
-      </tr>
+export const TodoList: React.FC<Props> = ({
+  todos,
+  setIsVisibleModal,
+  setUserId,
+  setPost,
+  selectedPostId,
```
vasylmiskiv, line 24:
```suggestion
  setSelectedTodo,
  selectedTodoId,
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,114 @@
...
+}) => {
+  function handleClick(values: HandleClick) {
```
vasylmiskiv, line 26: Should write more appropriate name of the function because it helps to select a todo

vasylmiskiv (changes requested):
Overall looks fine but needs to add some changes

StasSokolov (changes requested):
Great job, let's improve your code a bit more🔥

StasSokolov (approved):
Looks good to me, i'm approving it🔥

PR #1990, opened 2023-11-02
solution

Ayogoharo:
[DEMO LINK](https://ayogoharo.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,34 +1,56 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [activeTodo, setActiveTodo] = useState<Todo | null>(null);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+
+  useEffect(() => {
+    getTodos().then(setTodos).finally(() => setIsLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
-        <div className="container">
-          <div className="box">
-            <h1 className="title">Todos:</h1>
+        <h1 className="title">Todos:</h1>
 
-            <div className="block">
-              <TodoFilter />
-            </div>
+        <div className="block">
+          <TodoFilter
+            todos={todos}
+            setFilteredTodos={setFilteredTodos}
+          />
+        </div>
 
-            <div className="block">
-              <Loader />
-              <TodoList />
-            </div>
-          </div>
+        <div className="block">
+          {isLoading
+            ? (<Loader />)
+            : (
+              <TodoList
+                todos={filteredTodos}
+                activeTodo={activeTodo}
+                setActiveTodo={setActiveTodo}
+              />
+            )}
         </div>
       </div>
 
-      <TodoModal />
+      {activeTodo && (
+        <TodoModal
+          activeTodo={activeTodo}
+          setActiveTodo={setActiveTodo}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,85 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import React, { useEffect, useState } from 'react';
+import { Todo } from '../../types/Todo';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -1,34 +1,56 @@
...
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [activeTodo, setActiveTodo] = useState<Todo | null>(null);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+
+  useEffect(() => {
+    getTodos().then(setTodos).finally(() => setIsLoading(false));
```
polosanya, line 20:
```suggestion
    getTodos()
      .then(setTodos)
      .finally(() => setIsLoading(false));
```
Don't forget to catch the possible errors

src/App.tsx
```diff
@@ -1,34 +1,56 @@
...
 import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [activeTodo, setActiveTodo] = useState<Todo | null>(null);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
polosanya, line 17: Do you really need to have two different states for filtered and nonfiltered todos?
Just use useMemo instead

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,85 @@
...
-          className="delete"
+import React, { useEffect, useState } from 'react';
+import { Todo } from '../../types/Todo';
+
+function filterTodos(
+  todos: Todo[],
+  filterMethod: string,
+  query: string,
+): Todo[] {
+  let copy = [...todos];
```
polosanya, line 9: Filter is a non-mutating method. There is no need to create a copy

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,85 @@
...
+        .toLowerCase()
+        .includes(query.toLowerCase())));
+  }
+
+  if (filterMethod === 'active') {
+    (copy = copy.filter(todo => !todo.completed));
+  }
+
+  if (filterMethod === 'completed') {
+    (copy = copy.filter(todo => todo.completed));
```
polosanya, line 23: Would be great to use Enum instead of the hardcoded values. Also, switch case might be useful here

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,85 @@
...
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={event => setFilterMethod(event.target.value)}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
polosanya, line 52: Same here

polosanya (changes requested):
Nice, good job!
Here are a few comments about how you can improve your code:

yulabakun (approved):
Well done! One small suggestion

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,73 @@
...
+}
+
+export const TodoModal: React.FC<Props> = ({
+  activeTodo,
+  setActiveTodo,
+}) => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [activeUser, setActiveUser] = useState<User | null>(null);
+
+  if (activeTodo) {
```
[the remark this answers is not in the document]
StasSokolov: @Ayogoharo check this point please

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,58 @@
...
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={event => setFilterMethod(event.target.value)}
+          >
+            <option value={Method.all}>All</option>
+            <option value={Method.active}>Active</option>
+            <option value={Method.completed}>Completed</option>
```
StasSokolov, line 25: will be better use `.map()` for iterate

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,84 @@
...
+                <button
+                  data-cy="selectButton"
+                  className="button"
+                  type="button"
+                  onClick={() => setActiveTodo(todo)}
+                >
+                  <span className="icon">
+                    <i className={classNames('far', {
+                      'fa-eye-slash': todo.id === activeTodo?.id,
+                      'fa-eye': todo.id !== activeTodo?.id,
```
StasSokolov, line 72: DRY, let's create variable like `isSelectedTodo` and reuse

src/components/TodoModal/TodoModal.tsx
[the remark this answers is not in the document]
StasSokolov: check this also

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,73 @@
...
+
+            <div className="modal-card-body">
+              <p className="block" data-cy="modal-title">
+                {activeTodo?.title}
+              </p>
+
+              <p className="block" data-cy="modal-user">
+                {activeTodo?.completed
+                  ? (<strong className="has-text-success">Done</strong>)
+                  : (<strong className="has-text-danger">Planned</strong>)}
```
StasSokolov, line 60: let's use single `strong` tag and only change content and className

StasSokolov (changes requested):

src/components/TodoFilter/TodoFilter.tsx
Ayogoharo: I can do it, but... I mean, is it really necessary? I will need to create a new array for it, and copy all the methods in it.

polosanya (approved):
👍

PR #1997, opened 2023-11-02
First push to check tests

Drach-Illya:
[DEMO LINK](https://Drach-Illya.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,11 +1,36 @@
-export const TodoFilter = () => (
+interface Props {
+  value: string;
+  changeFilter: (string: string) => void;
+  changeInput: (string: string) => void;
+}
+
+export const TodoFilter: React.FC<Props> = ({
+  value,
+  changeFilter,
+  changeInput,
+}) => (
   <form className="field has-addons">
     <p className="control">
       <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
+        <select
+          data-cy="statusSelect"
+          onChange={(event) => changeFilter(event.target.value)}
+        >
+          <option
+            value="all"
+          >
+            All
+          </option>
+          <option
+            value="active"
+          >
+            Active
+          </option>
+          <option
+            value="completed"
+          >
+            Completed
+          </option>
         </select>
       </span>
     </p>
@@ -16,6 +41,8 @@ export const TodoFilter = () => (
         type="text"
         className="input"
         placeholder="Search..."
+        value={value}
+        onChange={(event) => changeInput(event.target.value)}
       />
       <span className="icon is-left">
         <i className="fas fa-magnifying-glass" />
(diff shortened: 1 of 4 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,36 @@
...
+          <option
+            value="active"
+          >
+            Active
+          </option>
+          <option
+            value="completed"
+          >
+            Completed
+          </option>
```
StasSokolov, line 33: will be better iterate `option` using `.map()` and `enum` for value

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,18 +41,25 @@ export const TodoFilter = () => (
...
 
       <span className="icon is-right" style={{ pointerEvents: 'all' }}>
         {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
-        />
+        {value !== ''
+          ? (
```
StasSokolov, line 54:
```suggestion
          && (
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,51 @@ export const TodoList: React.FC = () => (
...
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => changeShowing(todo.id)}
+            >
+              <span className="icon">
+                <i className={cn('far',
+                  {
+                    'fa-eye-slash': displayedTodo?.id === todo.id,
```
StasSokolov, line 67: DRY, create variable like `isSelectedTodo` and reuse

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,37 @@
...
+
+export const TodoModal: React.FC<Props> = ({
+  displayedTodo,
+  changeShowing,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [loadingUser, setLoadingUsers] = useState(false);
+
+  useEffect(() => {
+    if (displayedTodo !== null) {
```
StasSokolov, line 22:
```suggestion
    if (displayedTodo) {
```

should work same but simplified

StasSokolov (changes requested):
Almost done, let's fix a few minor comments and improve your code a bit more🔥

src/components/TodoFilter/TodoFilter.tsx
Drach-Illya: @Stanislav Sokolov  "will be better iterate option using .map() and enum for value"


If I recall, enum has the behaviour of an object, while .map() is an array method. Do you propose to make something like this?


```
enum ListOfOptions {
  array = [
    {title: 'All', value: 'all'},
    {title: 'Active', value: 'active'},
    {title: 'Completed', value: 'completed'},
  ]
}

ListOfOptions.array.map(option => (
  <option
    value=option.value
  >
    {option.title}
  </option>
))
```


Because this looks EXTREMELY cumbersome

superpooperxxx (approved):
🔥

PR #2005, opened 2023-11-08
add task solution

mrogozhinskaya:
[DEMO LINK](https://mrogozhinskaya.github.io/react_dynamic-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,81 @@
...
-            </span>
-          </button>
-        </td>
-      </tr>
+          return (
+            <tr
+              data-cy="todo"
+              className={
+                isSelectedTodo ? 'has-background-info-light' : ''
+              }
```
loralevitska, line 38: Use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally
Fix all cases

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

- add [trimming](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) to search 
image

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +27,61 @@ export const TodoList: React.FC = () => (
...
-      </tr>
+      {todos.map(todo => (
+        <tr
+          data-cy="todo"
+          className={
+            todo.id === selectedTodo?.id
+              ? ('has-background-info-light'
+              ) : (
+                ''
+              )
```
etojeDenys, line 38: you can use classnames package here

etojeDenys (changes requested):

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,79 @@
...
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map(todo => (
+          <>
```
etojeDenys, line 32: btw, this fragment looks redundant here

etojeDenys (approved):
lgtm

PR #2010, opened 2023-11-12
add task solution

andrew17752985:
https://andrew17752985.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +6,27 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | undefined>(undefined);
+  const [isLoading, setIsLoading] = useState(true);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+
+  useEffect(() => {
+    getTodos()
+      .then(serverTodos => {
+        setTodos(serverTodos);
+        setFilteredTodos(serverTodos);
+        setIsLoading(false);
+      })
+      .catch(error => {
+        throw new Error(error);
+      });
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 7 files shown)


TarasHoliuk:
table head column doesn't match table body column - example title (3 place and 4 place accordingly):

image

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,95 @@
...
-              <a href="mailto:Sincere@april.biz">
-                Leanne Graham
-              </a>
-            </p>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                type="button"
+                className="delete"
+                data-cy="modal-close"
+                onClick={() => setSelectedTodo(undefined)}
```
TarasHoliuk, line 61:
```suggestion
                onClick={() => setSelectedTodo(null)}
```
it's better to explicitly say it's defined but empty value - there is difference between `null` and `undefined`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,49 +1,95 @@
...
+
+  useEffect(() => {
+    setErrorMessage('');
+    setIsLoading(true);
+
+    if (todo) {
+      getUser(todo?.userId)
+        .then(userFromServer => {
+          setUser(userFromServer);
+          setIsLoading(false);
```
TarasHoliuk, line 28:
```suggestion
```
no need to do it here if you do it in `finally`

src/App.tsx
```diff
@@ -1,14 +1,32 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | undefined>(undefined);
```
TarasHoliuk, line 14:
```suggestion
  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
```

src/App.tsx
```diff
@@ -1,14 +1,32 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then(serverTodos => {
+        setTodos(serverTodos);
+        setFilteredTodos(serverTodos);
+        setIsLoading(false);
+      })
+      .catch(error => {
+        throw new Error(error);
```
TarasHoliuk, line 26: FYI: it'll lead to page crush - it's better not to throw an error if you won't handle it somehow. In other words, this catch block does nothing

TarasHoliuk (changes requested):
GJ :+1: 
Added a few minor comments. Also, fix the columns in the table

TarasHoliuk (approved):
Well done - approved :heavy_check_mark: 
P.S. Next time do not forget to re-deploy your solution after fixes

PR #2011, opened 2023-11-12
🌭🌭🌭

denys2:
https://denys2.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+
+  const visibleTodos = useMemo(() => {
+    return filterTodos(todos, filter, query);
+  }, [todos, query, filter]);
+
+  useEffect(() => {
+    getTodos()
+      .then(response => {
+        setTodos(response);
+      });
```
etojeDenys, line 29:
```suggestion
      .then(setTodos);
```
we can simplify it

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,89 @@
...
+      <tbody>
+        {visibleTodos.map(todo => (
+          <tr
+            data-cy="todo"
+            className={
+              todo.id === selectedTodo?.id ? (
+                'has-background-info-light'
+              ) : (
+                ''
+              )
```
etojeDenys, line 39: let's use classnames package here

etojeDenys (changes requested):

src/components/TodoList/TodoList.tsx
denys2: I did it

etojeDenys (approved):
lgtm

PR #2022, opened 2023-11-18
solution

xnquaint:
[DEMO LINK](https://xnquaint.github.io/react_dynamic-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,81 @@
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+          return (
+            <tr
+              data-cy="todo"
+              className={
+                isTodoSelected ? 'has-background-info-light' : ''
```
etojeDenys, line 38: we usually use classnames package for such cases

etojeDenys (approved):
good job

PR #2035, opened 2023-11-29
add task solution

4xmplme:
[DEMO LINK](https://4xmplme.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,10 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, {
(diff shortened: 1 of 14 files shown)


loralevitska (changes requested):
Almost done) Let's improve

use [string.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) for search
image
image

etojeDenys (approved):
good job

PR #2040, opened 2023-12-03
result

errisi:
[DEMO LINK](https://errisi.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,99 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { FC, useEffect, useState } from 'react';
+import { Todo } from '../../types/Todo';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  todos: Todo[];
+  setFilteredTodos: React.Dispatch<React.SetStateAction<Todo[]>>;
+};
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+enum SortType {
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
+
+const todoFilter = (
+  todos: Todo[],
+  sortType: string,
+  query: string,
+): Todo[] => {
+  let filteredTodos = [...todos];
+
+  switch (sortType) {
+    case SortType.Active:
+      filteredTodos = filteredTodos.filter(todo => !todo.completed);
+      break;
+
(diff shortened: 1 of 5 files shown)


ArtemSydun (approved):
gj

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,99 @@
...
+    case SortType.Active:
+      filteredTodos = filteredTodos.filter(todo => !todo.completed);
+      break;
+
+    case SortType.Completed:
+      filteredTodos = filteredTodos.filter(todo => todo.completed);
+      break;
+
+    default:
+      break;
```
roman-mirzoian, line 32: Do not leave the `default` block empty, there you can throw an error about the wrong type of sorting

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,99 @@
...
-      </span>
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
roman-mirzoian, line 86: Do not use inline styling

roman-mirzoian (changes requested):

src/components/TodoFilter/TodoFilter.tsx
errisi: > Do not leave the `default` block empty, there you can throw an error about the wrong type of sorting

the code doesn't compile with
```
  default:
    throw new Error('Цrong type of sorting!');
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,99 @@
...
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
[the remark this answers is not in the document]
errisi: > Do not disable the linter, but correct the code in such a way that the linter does not fight

"A control must be associated with a text label.eslint(jsx-a11y/control-has-associated-label)"

I tried to fix this for three hours, added labels, then asked for an action on the button like “onkeyup”, nothing helped me, so I just turned off the linter in order to make a commit

yevtok (approved):
Good job!

src/components/TodoFilter/TodoFilter.tsx
roman-mirzoian: @errisi I can't tell you anything about compilation problems without error text or a screenshot, but the option to leave the default block empty is definitely wrong.

src/components/TodoFilter/TodoFilter.tsx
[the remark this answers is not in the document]
roman-mirzoian: @errisi As in the previous comment, in order to understand how to help you, I need to see the result of your actions. This linter bug is fixed by adding `aria-label="Clear Search"`
I also have a question - why did you try to do this for three hours, but did not ask for help in the chat?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,31 +40,43 @@ export const TodoModal: React.FC = () => {
               className="modal-card-title has-text-weight-medium"
               data-cy="modal-header"
             >
-              Todo #2
+              {`Todo #${todo.id}`}
             </div>
 
             {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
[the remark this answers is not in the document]
roman-mirzoian: @errisi The same answer regarding the lack of result of your actions

roman-mirzoian (changes requested):
Correct the comments that could not be corrected from the last check or add screenshots of the failed results to your comments so that it can be understood that these comments cannot be corrected

PR #2041, opened 2023-12-03
solution

yevtok:
[DEMO LINK](https://yevtok.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,10 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,14 +1,51 @@
...
+    isLoadingTodos,
+  } = useContext(StateContext);
+
+  const filterTodos = useCallback((todosToFilter: Todo[]) => {
+    const { option, query } = filter;
+
+    return filterByQuery(filterByOption(todosToFilter, option), query);
+  }, [filter]);
+
+  const [filteredTodos, setFilteredTodos] = useState(filterTodos(todos));
```
polosanya, line 36: You shouldn't recreate the initial state
https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state

src/reducers/reducer.ts
```diff
@@ -0,0 +1,65 @@
+import { State } from '../types/State';
+import { Action } from '../types/Action';
+
+export function stateReducer(state: State, action: Action): State {
+  switch (action.type) {
+    case 'setTodos':
```
polosanya, line 6: Would be great to use Enum instead of the hardcoded values

polosanya (changes requested):
Nice! Let's improve your code a bit)

src/reducers/reducer.ts
yevtok: done

polosanya (approved):
👍

PR #2054, opened 2023-12-12
task solution

VeDono:
[DEMO LINK](https://VeDono.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,42 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+
+import './App.scss';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { getFilteredTodos } from './helpers';
+import { SelectorOption } from './enum';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [currentSelector, setCurrentSelector] = useState<string>(SelectorOption.All);
+  const [searchQuery, setSearchQuery] = useState('');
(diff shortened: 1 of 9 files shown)


ViktorAfk (approved):
NIce!

src/App.tsx
```diff
@@ -1,14 +1,42 @@
...
+
+  const filteredTodos = getFilteredTodos(todos, currentSelector, searchQuery) || null;
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      // eslint-disable-next-line no-alert
+      .catch(() => alert('Error during todos reciving 😥'))
```
roman-mirzoian, line 32: Interesting decision :)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
+}) => {
+  const handlerSelector = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setCurrentSelector(event.target.value);
+  };
+
+  const handlerSearchChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setSearchQuery(value);
+  };
```
roman-mirzoian, line 23: Why do you do destructuring for the event of the first handler and not for the second?)
In this case, it can not be done, because the value value is used only once in each handler

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
-      </span>
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {searchQuery && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
roman-mirzoian, line 58: Avoid using inline styles, use classes instead

roman-mirzoian (changes requested):

src/components/TodoFilter/TodoFilter.tsx
VeDono: Cause ESLint throw error with: "Make destructuring for that event", i will change that in patch.

src/components/TodoFilter/TodoFilter.tsx
VeDono: I reimplement that through TodoFilter.scss in patch

roman-mirzoian (approved):
Well done 👍

PR #2058, opened 2023-12-13
finish task

yefimov77:
[DEMO LINK](https://yefimov77.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import './App.scss';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +8,30 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodosToRender } from './helpers';
+import { ProgressStatus } from './types/ProgressEnum';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [progress, setProgress] = useState<ProgressStatus | string>(ProgressStatus.All);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  const closeModal = () => setSelectedTodo(null);
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const todosToRender = getTodosToRender(todos, searchQuery, progress);
+
   return (
     <>
       <div className="section">
@@ -17,18 +40,33 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                searchQuery={searchQuery}
+                setSearchQuery={setSearchQuery}
+                progress={progress}
+                setProgress={setProgress}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              <TodoList
+                todos={todosToRender}
+                selectedTodo={selectedTodo}
+                setSelectedTodo={setSelectedTodo}
+              />
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo !== null
+      && (
+        <TodoModal
+          selectedTodo={selectedTodo}
+          onClose={closeModal}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,34 +1,63 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { ProgressStatus } from '../../types/ProgressEnum';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+interface Props {
+  searchQuery: string,
+  setSearchQuery: (searchValue: string) => void,
+  progress: string,
+  setProgress: (value: ProgressStatus | string) => void;
+}
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button
-          data-cy="clearSearchButton"
-          type="button"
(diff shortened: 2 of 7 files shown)


src/App.tsx
```diff
@@ -17,18 +40,33 @@ export const App: React.FC = () => {
...
+                selectedTodo={selectedTodo}
+                setSelectedTodo={setSelectedTodo}
+              />
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo !== null
```
ArtemSydun, line 63: if your selectedTodo is null than you can write !selectedTodo

ArtemSydun (approved):
gj

src/App.tsx
yefimov77: > if your selectedTodo is null than you can write !selectedTodo

sure, i was rewritting logic, coz some thigns didn`t work properly and tried to resolve dat problem in different ways, but forgot bout that.

src/App.tsx
```diff
@@ -1,14 +1,37 @@
...
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  const closeModal = () => setSelectedTodo(null);
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
```
BudnikOleksii, line 30: Don't forget to add `catch` block to handle errors

src/App.tsx
BudnikOleksii: > if your selectedTodo is null than you can write !selectedTodo

Good point 👍  But it should be `selectedTodo` as it doesn't make sense to render modal if there is no selected todo, I think you just accidentally added `!`

src/App.tsx
```diff
@@ -1,14 +1,37 @@
...
+import { getTodosToRender } from './helpers';
+import { ProgressStatus } from './types/ProgressEnum';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [progress, setProgress] = useState<ProgressStatus | string>(ProgressStatus.All);
```
BudnikOleksii, line 20: The type should be only `ProgressStatus` as random string is not valid

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,79 @@
...
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+interface Props {
+  todos: Todo[];
+  selectedTodo: Todo | null;
+  setSelectedTodo: (t: Todo) => void;
```
BudnikOleksii, line 8: Avoid one-name variables/parameters

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,40 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+import { getUser } from '../../api';
+
+interface Props {
+  selectedTodo: Todo | null,
```
BudnikOleksii, line 9:
```suggestion
  selectedTodo: Todo,
```
You render the modal only if you actually have selectedTodo

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,40 @@
...
+
+  const [
+    currentUser,
+    setCurrentUser,
+  ] = useState<User | null>(null);
+
+  useEffect(() => {
+    setIsModalloading(true);
+
+    getUser(selectedTodo?.userId ?? 0)
```
BudnikOleksii, line 25: And this one check will be redundant.
Also, you shouldn't call API with invalid userId at all

src/helpers.ts
```diff
@@ -0,0 +1,25 @@
+import { ProgressStatus } from './types/ProgressEnum';
+import { Todo } from './types/Todo';
+
+export function getTodosToRender(
+  todos: Todo[],
+  searchQuery: string,
+  progress: string,
```
BudnikOleksii, line 7: progress isn't string, you have type for this

src/helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+
+      case ProgressStatus.Active:
+        return !todo.completed;
+
+      default:
+        return true;
+    }
+  })
+
+    .filter(todo => todo.title.toLowerCase().includes(preparedQuery));
```
BudnikOleksii, line 24: Remove blank line between chained methods

BudnikOleksii (changes requested):
Good work 👍 
But could be even better, left some comments and suggestions.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {searchQuery
+          && (
+            // eslint-disable-next-line jsx-a11y/control-has-associated-label
```
roman-mirzoian, line 53: Try to rewrite the code so that it is not necessary to disable the linter

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,65 @@
...
         />
-      </span>
-    </p>
-  </form>
-);
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
roman-mirzoian, line 50: Avoid using "magic values"

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,41 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
roman-mirzoian, line 1: Try to rewrite the code so that it is not necessary to disable the linter

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +44,37 @@ export const TodoModal: React.FC = () => {
               className="modal-card-title has-text-weight-medium"
               data-cy="modal-header"
             >
-              Todo #2
+              {`Todo #${todo?.id}`}
             </div>
 
             {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
roman-mirzoian, line 50: Try to rewrite the code so that it is not necessary to disable the linter

src/helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+        return todo.completed;
+
+      case ProgressStatus.Active:
+        return !todo.completed;
+
+      default:
+        return true;
+    }
+  })
+    .filter(todo => todo.title.toLowerCase().includes(preparedQuery));
```
roman-mirzoian, line 23: Calling two filters in a row is not an optimized solution, it is better to combine it into one filter:
```suggestion
  function includesIgnoreCase(str, query) {
    return str.toLowerCase().includes(query);
  }

  return todos.filter(todo => {
    switch (progress) {
      case ProgressStatus.Completed:
        return todo.completed && includesIgnoreCase(todo.title, preparedQuery);
  
      case ProgressStatus.Active:
        return !todo.completed && includesIgnoreCase(todo.title, preparedQuery);
  
      default:
        return includesIgnoreCase(todo.title, preparedQuery);
    }
  });
```

roman-mirzoian (changes requested):
Almost done!

roman-mirzoian (changes requested):
Comments about the linter need to be fixed, if you can't - ask for help in the chats or on Q&A

BudnikOleksii (approved):
Good work 👍

PR #2060, opened 2023-12-13
solution 1

mickhakk:
[DEMO LINK](https://mickhakk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,30 +38,34 @@ export const TodoModal: React.FC = () => {
...
-              quis ut nam facilis et officia qui
+              {`${selectedTodo.title}`}
             </p>
 
             <p className="block" data-cy="modal-user">
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              <strong className="has-text-danger">
+                {statusMessage}
+              </strong>
```
Yevhen-Kolosov, line 63: `{statusMessage}` should change its color (by changing the class of the tag) depending on which status the todo has (class `has-text-danger` for an active todo and `has-text-success` for a completed one) KEKWait (2)

Yevhen-Kolosov (approved):
Well done! 👍 Although, there are points to be checked. 🤔 
 
I think there might be some mistake that I struggle to find (as there is an unusual interaction on opening a TodoModal in the demo: the modal window pops up for a brief moment right before the 'loading' element) so I'll pass the request to the mentors by approving it. KEKWait (2)

src/App.tsx
```diff
@@ -1,14 +1,54 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [query, setQuery] = useState('');
+  const [filter, setFilter] = useState<Status>(Status.All);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(data => setTodos(data))
```
BudnikOleksii, line 24:
```suggestion
      .then(setTodos)
```
You can simplify in this case

src/components/TodoModal/TodoModal.tsx
BudnikOleksii: > `{statusMessage}` should change its color (by changing the class of the tag) depending on which status the todo has (class `has-text-danger` for an active todo and `has-text-success` for a completed one) KEKWait (2)

⬆️

BudnikOleksii (changes requested):
Good job 👍 
Just a few minor fixes are required, check the comments from the previous reviewer as well

ArtemTeslenko (approved):
I am approving it, to not to block you, but some of issues are still not fixed(

PR #2065, opened 2023-12-13
almost ready

Mantserov:
[DEMO LINK](https://Mantserov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,54 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [filter, setFilter] = useState('');
+
+  const [isLoading, setIsLoading] = useState(false);
+
+  let visibleTodos = todos;
+
+  if (searchQuery) {
+    visibleTodos = visibleTodos.filter(todo => todo.title.toLowerCase().includes(searchQuery.toLowerCase()));
+  }
+
+  if (filter) {
+    switch (filter) {
+      case 'all': {
+        break;
+      }
+
+      case 'active': {
+        visibleTodos = visibleTodos.filter(todo => todo.completed === false);
+        break;
+      }
+
+      case 'completed': {
+        visibleTodos = visibleTodos.filter(todo => todo.completed === true);
+        break;
+      }
+
+      default:
+        break;
+    }
+  }
+
+  useEffect(
+    () => {
+      setIsLoading(true);
+      getTodos().then(setTodos)
+        // eslint-disable-next-line no-console
+        .catch(console.log)
+        .finally(() => (setIsLoading(false)));
+    }, [],
+  );
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
+
+      case 'completed': {
+        visibleTodos = visibleTodos.filter(todo => todo.completed === true);
+        break;
+      }
+
+      default:
+        break;
+    }
+  }
```
emilionila, line 46: I think, in this task, you can simplify code without using switch but lets see what mentors will say)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,68 @@
...
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  onSearchQuery: React.Dispatch<React.SetStateAction<string>>,
+  filter: string,
+  onFilter: React.Dispatch<React.SetStateAction<string>>
```
emilionila, line 6: im not sure about using dispatch

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,83 @@
...
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+type Props = {
+  todos: Todo[];
+  selectedTodo: Todo | null
+  onSelectedTodo: React.Dispatch<React.SetStateAction<Todo | null>>;
```
emilionila, line 8: `onSelectedTodo: React.Dispatch<React.SetStateAction<Todo | null>>` same is here, I think you can simplify it by `onSelectedTodo: (value: string) => void`

emilionila (approved):
Looks good, but check comments(not sure about few things). Im gonna aprove, but lets see what mentors are going to say 👍

src/App.tsx
BudnikOleksii, line 46: - Create a separate util function for this logic.
- Get rid of magic values such as 'all', create enum for them

src/components/TodoFilter/TodoFilter.tsx
BudnikOleksii: > im not sure about using dispatch

It's ok, this a correct type of `setState` function

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,68 @@
...
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          placeholder="Search..."
+          value={query}
+          onChange={event => {
+            setQuery(event.target.value);
+            onSearchQuery(event.target.value);
+          }}
```
BudnikOleksii, line 44: Create a separate handler for this function, and you can destructing the `value` from the `event.target` to avoid repeating them

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,68 @@
...
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => {
+                setQuery('');
+                onSearchQuery('');
+              }}
```
BudnikOleksii, line 61: The same, create separate handler

src/components/TodoList/TodoList.tsx
BudnikOleksii: > `onSelectedTodo: React.Dispatch<React.SetStateAction<Todo | null>>` same is here, I think you can simplify it by `onSelectedTodo: (value: string) => void`

That's ok, in this case, you can pass also a callback function as `setState` accepts it
https://www.freecodecamp.org/news/using-typescript-in-react-apps/#how-to-type-hooks

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,83 @@
...
+                  <i className="fas fa-check" />
+                </span>
+              </td>
+            ) : (
+              <td className="is-vcentered" />
+            )}
+            <td className="is-vcentered is-expanded">
+              <p
+                className={
+                  todo.completed === true
```
BudnikOleksii, line 51:
```suggestion
                  todo.completed
```
You don't need to compare with `true`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,32 +36,41 @@ export const TodoModal: React.FC = () => {
...
-              quis ut nam facilis et officia qui
+              {todo.title}
             </p>
 
-            <p className="block" data-cy="modal-user">
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+            {user && (
+              <p className="block" data-cy="modal-user">
+                {/* <strong className="has-text-success">Done</strong> */}
```
BudnikOleksii, line 58: Remove comments

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, improve your solution in accordance

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filter}
+            onChange={event => onFilter(event.target.value)}
+          >
+            <option value="all">{Filter.all}</option>
```
polosanya, line 36: Would be great to use Enum for value too

polosanya (approved):
Great 🔥

PR #2068, opened 2023-12-14
dynamic list of odos

PonomarchukLana:
[DEMO LINK](https://ponomarchuklana.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,37 @@
+/* eslint-disable import/no-cycle */
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect, createContext } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import { getTodos } from './api';
+import { SortBy, getCurrentData } from './helper';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import './App.scss';
+
+export const TodoContext = createContext<number | null>(null);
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import './App.scss';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | []>([]);
+  const [query, setQuery] = useState('');
+  const [sortBy, setSortBy] = useState<keyof typeof SortBy>('All');
```
Fluebubble, line 18: You can write like
`const [sortBy, setSortBy] = useState<SortBy>(SortBy);`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,71 @@
...
-          className="delete"
+import React, { useState } from 'react';
+
+type Props = {
+  getQuery: (arg: string) => void;
+  getSortBy: (arg: any) => void;
+};
+
+export const TodoFilter: React.FC<Props> = ({ getQuery, getSortBy }) => {
+  const [query, setQuery] = useState('');
```
Fluebubble, line 9: is it correct to have query in App.tsx and here?

Fluebubble (approved):

src/components/TodoFilter/TodoFilter.tsx
roman-mirzoian: Yes, it would be worth keeping this state in only one component

roman-mirzoian (changes requested):

ArtemTeslenko (approved):
Great 🔥

PR #2074, opened 2023-12-16
add task solution

Softjey:
-[DEMO LINK](https://softjey.github.io/react_dynamic-list-of-todos/)

For some reason, one test fails, but everything works correctly

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,22 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './services/getTodos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [preparedTodos, setPreparedTodos] = useState<Todo[]>([]);
+
+  useEffect(() => {
+    getTodos().then(todosFromServer => {
+      setTodos(todosFromServer);
+      setPreparedTodos(todosFromServer);
+    });
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +31,33 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                todos={todos}
+                setPreparedTodos={setPreparedTodos}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0
+                ? (<Loader />)
+                : (
+                  <TodoList
+                    todos={preparedTodos}
+                    selectedTodo={selectedTodo}
+                    setSelectedTodo={setSelectedTodo}
+                  />
+                )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal
+          todo={selectedTodo}
+          setSelectedTodo={setSelectedTodo}
+        />
+      )}
     </>
   );
 };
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,14 +1,28 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './services/getTodos';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [preparedTodos, setPreparedTodos] = useState<Todo[]>([]);
```
SanyaBratashchuk, line 17:
```suggestion
  const [todos, setTodos] = useState<Todo[]>([]);
```

src/constants/TodosStatusQuery.ts
```diff
@@ -0,0 +1,7 @@
+enum TodosStatusQuery {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
```
SanyaBratashchuk, line 4:
```suggestion
  All = 'all',
  Active = 'active',
  Completed = 'completed',
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,92 @@
...
+  const [textQuery, setTextQuery] = useState('');
+  const [statusQuery, setStatusQuery] = useState(TodosStatusQuery.all);
+
+  useEffect(() => {
+    let filtredTodos = getFiltredTodosByStatus(todos, statusQuery);
+
+    filtredTodos = getFiltredTodosByText(filtredTodos, textQuery);
+
+    setPreparedTodos(filtredTodos);
+  }, [todos, textQuery, statusQuery]);
```
SanyaBratashchuk, line 44:
```suggestion
  }, [todos, textQuery, statusQuery, setPreparedTodos]);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,92 @@
...
+
+export const TodoFilter: React.FC<Props> = ({
+  todos,
+  setPreparedTodos,
+}) => {
+  const [textQuery, setTextQuery] = useState('');
+  const [statusQuery, setStatusQuery] = useState(TodosStatusQuery.all);
+
+  useEffect(() => {
+    let filtredTodos = getFiltredTodosByStatus(todos, statusQuery);
```
SanyaBratashchuk, line 39:
```suggestion
    let filteredTodos = getFiltredTodosByStatus(todos, statusQuery);
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +27,53 @@ export const TodoList: React.FC = () => (
...
-            <span className="icon">
-              <i className="far fa-eye-slash" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          data-cy="todo"
+          className=""
```
SanyaBratashchuk, line 33:
```suggestion
```

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +27,53 @@ export const TodoList: React.FC = () => (
...
+              type="button"
+              onClick={() => setSelectedTodo(todo)}
+            >
+              <span className="icon">
+                <i
+                  className={`far
+                  ${selectedTodo?.id === todo.id
+          ? 'fa-eye-slash'
+          : 'fa-eye'
+        }`}
```
SanyaBratashchuk, line 70: It is better to use libraries like `classnames` or `clsx` for conditional classNames

SanyaBratashchuk (changes requested):

src/App.tsx
Softjey: Removing the preparedTodos state and relying solely on the todos state to manage both the complete and filtered todo lists can lead to issues. This approach can cause unwanted side effects like overwriting the original data from the server with the filtered data, leading to difficulties in managing and displaying the todos effectively. It also increases the risk of creating an infinite rendering loop, where changes in the filtered list trigger re-renders that affect the original list, causing continuous updates.

src/components/TodoList/TodoList.tsx
Softjey: I don't want to add a whole library for one case

anastasiiavorobiova (approved):
Good work

PR #2080, opened 2023-12-19
Develop

Patrykmclaren14:
[solution](https://Patrykmclaren14.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,14 +11,14 @@ import { TodoFilter } from './components/TodoFilter';
 import { getTodos, getUser } from './api';
 
 // import { Todo } from './types/Todo';
-// import { User } from './types/User';
+import { User } from './types/User';
 
 interface Todo {
   id: number;
   title: string;
   completed: boolean;
   userId: number;
-  user?: any;
+  user: User;
 }
 
 export const App: React.FC = () => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -57,6 +57,16 @@ export const App: React.FC = () => {
     }
   };
 
+  const SearchTodo = (value: string) => {
+    const modifiedValue = value.toLocaleLowerCase().trim();
+
+    const searchedValue = TodosFromServer.filter(
+      todo => todo.title.toLocaleLowerCase().includes(modifiedValue),
+    );
+
+    setTodos(searchedValue);
+  };
+
   return (
     <>
       <div className="section">
@@ -65,7 +75,10 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter completed={handleCompletedTodos} />
+              <TodoFilter
+                completed={handleCompletedTodos}
+                search={SearchTodo}
+              />
             </div>
 
             <div className="block">
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,47 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
-import { Loader } from './components/Loader';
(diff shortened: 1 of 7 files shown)


Patrykmclaren14:
image

src/api.ts
```diff
@@ -17,7 +17,7 @@ function get<T>(url: string): Promise<T> {
   const fullURL = BASE_URL + url + '.json';
 
   // we add some delay to see how the loader works
-  return wait(300)
+  return wait(1)
```
MarcinLigmann, line 20: i don't understand why...

MarcinLigmann (approved):
looks fine👌

src/api.ts
JoyCoffeeAddict: @MarcinLigmann to make wait time shorter, basically `wait` is to simulate a network request

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,58 @@
...
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+  const { searchTodo, handleCompletedTodos } = useMyContext();
+
+  useEffect(() => {
+    searchTodo(qwerty);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [qwerty]);
```
JoyCoffeeAddict, line 12: you have a place where you change your query, it's Input's `onChange` 
put `searchTodo` there and there will be no need for a `useEffect`

src/context/myContext.tsx
```diff
@@ -0,0 +1,118 @@
...
+  useEffect(() => {
+    getTodos().then((allTodos: Todo[]) => {
+      const todoPromises = allTodos.map(todo => {
+        return getUser(todo.userId)
+          .then((user: User) => {
+            const updatedTodo = { ...todo, user };
+
+            return updatedTodo;
+          });
+      });
```
JoyCoffeeAddict, line 49: you are already using `Promise.all` try to refactor youre code, using `async` and `await` to avoid nesting a loop of promises inside a `then` block
Also, you need to check if the promise throws, add a `catch` block.

src/context/myContext.tsx
```diff
@@ -0,0 +1,118 @@
...
+      Promise.all(todoPromises)
+        .then(todoWithUser => {
+          setAllTodos(todoWithUser);
+          setTodos(todoWithUser);
+          setIsLoad(false);
+        });
+    });
+  }, []);
+
+  const handleCompletedTodos = (sortType: string) => {
```
JoyCoffeeAddict, line 60: maybe a better name would be `handleVisibleTodos`

JoyCoffeeAddict, line 60: sortType should be enum

src/context/myContext.tsx
```diff
@@ -0,0 +1,118 @@
...
+    let completedTodos: Todo[] = todos;
+
+    switch (sortType) {
+      case 'all':
```
JoyCoffeeAddict, line 64: cases should be enum's values

JoyCoffeeAddict (changes requested):

mbruhler (changes requested):
There is a problem with UI. When clicking eye icon, the modal loads only first time, then there is just a flickering not a loading indication (loader is barely visible because it hides in a fraction of second). 

This is due to `isUserLoading` state handling and the way you handle this state. Propably this should be true on clicking, but somehow is not. Check also rendering state (console.log the value in modal). This should be fixed.

src/App.tsx
```diff
@@ -1,16 +1,18 @@
...
 import '@fortawesome/fontawesome-free/css/all.css';
-
+import TodoContextProvider, { useMyContext } from './context/myContext';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const { isLoading } = useMyContext();
```
danon321, line 11: i think "useMyContext" is not very descriptive

danon321 (changes requested):

mpiskadlo (approved):
Nicely done ;)

PR #2081, opened 2023-12-19
Develop

mosowiecka:
[DEMO LINK](https://mosowiecka.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -67,7 +67,9 @@ export const App: React.FC = () => {
 
             <div className="block">
               <Loader />
-              <TodoList todos={filterTodos(todos, todosActivityFilter, todosQuery)}/>
+              <TodoList
+                todos={filterTodos(todos, todosActivityFilter, todosQuery)}
+              />
             </div>
           </div>
         </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,33 +1,52 @@
-
 /* eslint-disable max-len */
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-/*  eslint-disable */
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 // import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 import { Todo } from './types/Todo';
-import { getTodos} from './api';
+import { getTodos } from './api';
+
+function filterTodos(todos: Todo[], todosActivityFilter: string, todosQuery: string) {
+  let resultTodos = [...todos];
+  const query = todosQuery.toLowerCase();
+
+  switch (todosActivityFilter) {
+    case 'Completed':
+      resultTodos = resultTodos.filter(todo => todo.completed
+      && todo.title.toLowerCase().includes(query));
+      break;
+    case 'Active':
+      resultTodos = resultTodos.filter(todo => !todo.completed
+      && todo.title.toLowerCase().includes(query));
+      break;
+    default:
+      resultTodos = resultTodos.filter(todo => todo.title.toLowerCase().includes(query));
+      break;
+  }
+
+  return resultTodos;
+}
 
 export const App: React.FC = () => {
-
   const [todos, setTodos] = useState<Todo[]>([]);
   const [todosActivityFilter, setTodosActivityFilter] = useState('All');
   const [todosQuery, setTodosQuery] = useState('');
   const [todoShownId, setTodoShownId] = useState(-1);
   const [loadingDone, setLoadingDone] = useState(false);
+
   useEffect(() => {
     const loadTodos = async () => {
       const data = await getTodos();
-    }
+
       setTodos(data);
+      setLoadingDone(true);
     };
 
     loadTodos();
   }, []);
-
 
   return (
     <>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,6 +37,10 @@ export const App: React.FC = () => {
   const [todoShownId, setTodoShownId] = useState(-1);
   const [loadingDone, setLoadingDone] = useState(false);
 
+  function setActivityFilter(filterValue: string) {
+    setTodosActivityFilter(filterValue);
+  }
+
   useEffect(() => {
     const loadTodos = async () => {
       const data = await getTodos();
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,57 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
+// import { TodoModal } from './components/TodoModal';
```
kate-prakofyeva, line 7: don't forget to delete all unnecessary comments

src/App.tsx
```diff
@@ -1,14 +1,57 @@
...
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+function filterTodos(todos: Todo[], todosActivityFilter: string, todosQuery: string) {
+  let resultTodos = [...todos];
+  const query = todosQuery.toLowerCase();
+
+  switch (todosActivityFilter) {
+    case 'Completed':
```
kate-prakofyeva, line 17:
```
export enum SelectType {
  All = 'all',
  Active = 'active',
  Completed = 'completed',
}
```

you can create an enam type for Select options and use it like this, bonus - tips 
`case SelectType.Active:`

src/App.tsx
```diff
@@ -1,14 +1,57 @@
...
+
+  return resultTodos;
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosActivityFilter, setTodosActivityFilter] = useState('All');
+  const [todosQuery, setTodosQuery] = useState('');
+  const [todoShownId, setTodoShownId] = useState(-1);
+  const [loadingDone, setLoadingDone] = useState(false);
```
kate-prakofyeva, line 38: loadingDone = isLoading
we usually use this variant of the word

kate-prakofyeva (changes requested):
the task was not completed, it turned out to review by mistake

src/App.tsx
```diff
@@ -17,18 +72,26 @@ export const App: React.FC = () => {
...
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!loadingDone
+              && (<Loader />)}
+              <TodoList
+                todos={filterTodos(todos, todosActivityFilter, todosQuery)}
+                todoFocusedOn={todoShown}
+                focusOnTodo={focusOnTodo}
+              />
```
mpiskadlo, line 89: Use a ternary operator here, so you either render the Loader or TodoList depending on `loadingDone`

PS. I suggest renaming this piece of state to `isLoading`, more common convention

mpiskadlo: Currently when the list is loading, you see loader and table at the same time

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+function filterTodos(todos: Todo[], todosActivityFilter: string, todosQuery: string) {
+  let resultTodos = [...todos];
+  const query = todosQuery.toLowerCase();
+
+  switch (todosActivityFilter) {
```
mpiskadlo, line 16: Create a type for `todosActivityFilter`. You could use enum but I would personally go for an union type ;)

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+      break;
+  }
+
+  return resultTodos;
+}
 
 export const App: React.FC = () => {
+  const [todoShown, setTodoShown] = useState<Todo>();
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosActivityFilter, setTodosActivityFilter] = useState('All');
```
mpiskadlo, line 36: Also apply the type for `todosActivityFilter` here, for this piece of state

```suggestion
  const [todosActivityFilter, setTodosActivityFilter] = useState<SomeTimeYouCreated>('All');
```

mpiskadlo (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
+      break;
+    case 'active':
+      resultTodos = resultTodos.filter(todo => !todo.completed
+      && todo.title.toLowerCase().includes(query));
+      break;
+    default:
+      resultTodos = resultTodos.filter(todo => todo.title.toLowerCase().includes(query));
+      break;
+  }
+
```
JoyCoffeeAddict, line 30: I would filter with query after the `switch` statement so that you don't repet yourself

src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
+  useEffect(() => {
+    const loadTodos = async () => {
+      const data = await getTodos();
+
+      setTodos(data);
+      setIsLoading(true);
+    };
+
+    loadTodos();
+  }, []);
```
JoyCoffeeAddict, line 66: I would put all of that into a Context

src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
+  const setQuery = (value: string) => {
+    setTodosQuery(value);
+  };
+
+  const focusOnTodo = (todo: Todo) => {
+    setTodoShown(todo);
+  };
+
+  const unFocusOnTodo = () => {
+    setTodoShown(undefined);
```
JoyCoffeeAddict, line 54: it should be `null` not `undefined` 
`null` means that there is absence of something, `undefined` means that some variable is not yet defined.
we use null when we want to delete some value and point that there is no todo
and `undefined` means that the the code didn't reach this line

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,16 @@
 import React from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoInfo } from '../TodoInfo';
 
-export const TodoList: React.FC = () => (
+type Props = {
+  todos: Todo[],
+  todoFocusedOn: Todo | undefined,
```
JoyCoffeeAddict, line 7: I would go with `Todo | null` as explained above.

JoyCoffeeAddict (changes requested):

mpiskadlo (approved):
Well done

PR #2083, opened 2023-12-20
Solution

MarcinLigmann:
[DEMO LINK](https://Marcinligmann.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -4,7 +4,7 @@ import { Todo } from '../../types/Todo';
 
 type Props = {
   todos: Todo[] | null;
-  selectedTodo: Todo;
+  selectedTodo: Todo | null;
   setSelectedTodo: (task: Todo) => void;
   selectedModel: boolean,
   setSelectedModel: (item: boolean) => void;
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,84 @@
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+        {todos?.map(task => (
+          <tr key={task.id} data-cy="todo" className="">
```
zaura333, line 41: There's a 'has-background-info-light' class missing when todo is selected

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,84 @@
...
-        </td>
-      </tr>
+export const TodoList: React.FC<Props> = (
+  {
+    todos, selectedTodo, setSelectedTodo, isModalOpen, setIsModalOpen,
+  },
+) => {
+  const handleSelectedClick = (task: Todo) => () => {
+    setSelectedTodo(task);
+    setIsModalOpen(!isModalOpen);
```
zaura333, line 20: I wonder if it could just be set to true? Because it goes back to false with close button in modal and not from toggling

zaura333 (approved):
Nice! I'm just leaving two small suggestions

mpiskadlo (changes requested):
After clicking on 'show todo', the modal opens for a fraction of a second, then the spinner appears and after some time the modal shows again. Try to debug that ;)

I'm not giving you any more instructions, because that's the request you could get from a tester (QA). Great way to practice 🙌

mvjl000 (approved):
Goooooood

mpiskadlo (approved):
Wrong account, sorry 😆 

Goooooooooooood

PR #2088, opened 2023-12-21
Solution. 

adrian200002:
I have no idea why my solution failing 3 filter tests. I checked and everything work as it should be. Works the same like in 'the working page' provided in readme.

[DEMO LINK](https://adrian200002.github.io/react_dynamic-list-of-todos/)

image


szymanski-marcin (approved):
I'm not experienced enough to tell you what's exactly going on with your filter. I would do that similar for sure. We need Mentor's help here!

mbruhler (changes requested):
- your modal flickers, when I click it (shows up quickly, hides, then loader appears, and then modal). This is due to bad state handling, something is wrong here and you have to debug it. Component rerenders and this is due to this fact.
- as of tests - it's hard to tell now, but first let's resolve the flickering issue, then we will address this

mpiskadlo (changes requested):
In your solution, both the table and the loader for it is visible (in mate demo, there isn't such a behavior. Maybe it will help with the tests). Use for instance a ternary to decide which one you want to render, basing on isLoading value/

mpiskadlo (approved):

PR #2089, opened 2023-12-21
Develop

szymanski-marcin:
[DEMO LINK](https://szymanski-marcin.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,20 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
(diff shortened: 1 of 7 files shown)


Barosz30 (approved):
Great code, i like it very much!

If you feel fancy, you could add filtering by id of post

src/App.tsx
```diff
@@ -17,18 +55,32 @@ export const App: React.FC = () => {
...
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+
+              <TodoList
+                todos={filteredTodos}
+                selectedTodo={selectedTodo}
+                setSelectedTodo={setSelectedTodo}
+              />
```
mpiskadlo, line 72: Use a ternary operator here, so you don't render empty list while loading is true

mpiskadlo (changes requested):
Great job! Just one small improvement ;)

mpiskadlo (approved):
Merry Christmas!

PR #2095, opened 2023-12-24
Develop

jacekkosiba:
[DEMO LINK](https://jacekkosiba.github.io/react_dynamic-list-of-todos/)

To the mentor reviewing this solution:
I believe that my solution works slightly slower than the working example. I would appreciate feedback on potential reasons for this.
Additionally, I solved the task using props, but I'm considering whether using context might be more advantageous?
Also, I've just realized that I forgot to implement any optimizing hooks such as memo or useCallback. While these aren't required for the task, if you believe they would be beneficial, please let me know, and I will incorporate them.
Thank you 😊

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -94,6 +94,7 @@ export const App: FC = () => {
                     <TodoList
                       todos={todos}
                       onTodoSelect={handleTodoSelected}
+                      selectedId={selectedTodo?.id}
                     />
                   )
                   : <Loader />
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,34 +1,115 @@
 /* eslint-disable max-len */
-import React from 'react';
+import { FC, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
+import { Todo, FilterType } from './types';
+import { getTodos } from './api';
 import { TodoList } from './components/TodoList';
-import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { TodoModal } from './components/TodoModal';
+import { TodoFilter } from './components/TodoFilter';
 
-export const App: React.FC = () => {
+export const App: FC = () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,34 +1,116 @@
...
+export const App: FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filterType, setFilterType] = useState<FilterType>('all');
+  const [query, setQuery] = useState<string>('');
+
+  useEffect(() => {
+    const loadTodos = async () => {
+      try {
+        const allTodos = await getTodos();
```
mpiskadlo, line 22: Your solution might seem slower, because you are awaiting getTodos with each character typed. Instead, fetch todos once, store them in the state and filter them ;)
This should already improve the performance. If not, then you can try to add some hooks like `useMemo` or something else ;)

mpiskadlo (changes requested):

WiktoriaBielec (approved):
Great code!

src/App.tsx
```diff
@@ -1,34 +1,127 @@
...
+
+      if (query.trim().length) {
+        filteredTodos = filteredTodos
+          .filter(todo => todo.title.toLocaleLowerCase()
+            .includes(query.toLocaleLowerCase()));
+      }
+
+      setVisibleTodos(filteredTodos);
+    }
+  }, [filterType, query, todos]);
```
JoyCoffeeAddict, line 61: this should be `useMemo` not `useEffect`, because you are making calculations based on another state to get a value.

src/App.tsx
```diff
@@ -1,34 +1,127 @@
...
 
+            {
+              errorMessage && <p>{errorMessage}</p>
+            }
+
             <div className="block">
-              <Loader />
-              <TodoList />
+              {
+                todos
```
JoyCoffeeAddict, line 103: I would do `todos?.length > 0`

JoyCoffeeAddict (changes requested):

mbruhler (approved):
Good job, looks great!

PR #2103, opened 2023-12-28
Solution

Timi2424:
[DEMO LINK](https://Timi2424.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,62 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { getTodos } from './api';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+
+enum Status {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,62 @@
...
+      break;
+
+    case Status.Completed:
+      selectedTodos = todos.filter(currentTodo => currentTodo.completed);
+      break;
+
+    default:
+      selectedTodos = todos;
+      break;
+  }
```
mpiskadlo, line 46: I would go for something like this:
```suggestion
  const filteredTodos = () => {
    switch (filterByStatus) {
    case Status.Active:
      return todos.filter(currentTodo => !currentTodo.completed);

    case Status.Completed:
      return todos.filter(currentTodo => currentTodo.completed);

    default:
      return todos;
  }
  };
```

And similar thing for `visibleTodos` ;)

mpiskadlo: And then in JSX you just do `filteredTodos().map`

mpiskadlo (changes requested):

src/App.tsx
Timi2424: what an oversight, thanks!

Timi2424: I don't know exactly why do you want me to map it. I hope my refactor is sufficient

Barosz30 (approved):
Seems just right :)

mbruhler (approved):
Well done!

PR #2105, opened 2023-12-29
created dynamic list of todos

OPTIMISTIXX:
[DEMO LINK](https://OPTIMISTIXX.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,8 +8,8 @@ import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 import { Todo } from './types/Todo';
-import { getTodos } from './services/getTodos';
 import { FilteringType } from './types/FilteringType';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
@@ -22,12 +22,10 @@ export const App: React.FC = () => {
 
   useEffect(() => {
     setTodosLoading(true);
-    setTimeout(() => {
-      getTodos().then((todo) => {
-        setTodos(todo);
-      })
-        .finally(() => setTodosLoading(false));
-    }, 300);
+    getTodos().then((todo) => {
+      setTodos(todo);
+    })
+      .finally(() => setTodosLoading(false));
   }, []);
 
   return (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,29 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './services/getTodos';
+import { FilteringType } from './types/FilteringType';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosLoading, setTodosLoading] = useState(false);
+  const [filter, setFilter] = useState<FilteringType>(FilteringType.all);
+  const [title, setTitle] = useState('');
+  const [todoCard, setTodoCard] = useState<Todo>();
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [iconId, setIconId] = useState(0);
+
+  useEffect(() => {
+    setTodosLoading(true);
+    setTimeout(() => {
+      getTodos().then((todo) => {
+        setTodos(todo);
+      })
+        .finally(() => setTodosLoading(false));
+    }, 300);
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +38,42 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                input={title}
+                setInput={setTitle}
+                selectedButton={filter}
+                setSelectedButton={setFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todosLoading && (
+                <Loader />
+              )}
+              {!todosLoading && todos.length > 0 && (
+                <TodoList
+                  title={title}
+                  todos={todos}
+                  filter={filter}
+                  iconId={iconId}
+                  setIsModalOpen={setIsModalOpen}
+                  setTodoCard={setTodoCard}
+                  setIconId={setIconId}
+                />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {isModalOpen && (
+        <TodoModal
+          todoCard={todoCard}
+          setIsModalOpen={setIsModalOpen}
+          setIconId={setIconId}
+        />
+      )}
+
     </>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,9 +16,7 @@ export const App: React.FC = () => {
   const [todosLoading, setTodosLoading] = useState(false);
   const [filter, setFilter] = useState<FilteringType>(FilteringType.all);
   const [title, setTitle] = useState('');
-  const [todoCard, setTodoCard] = useState<Todo>();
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,14 +1,35 @@
...
+  const [filter, setFilter] = useState<FilteringType>(FilteringType.all);
+  const [title, setTitle] = useState('');
+  const [todoCard, setTodoCard] = useState<Todo>();
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [iconId, setIconId] = useState(0);
+
+  useEffect(() => {
+    setTodosLoading(true);
+    setTimeout(() => {
+      getTodos().then((todo) => {
```
nazarmatsevych, line 26: You already have ready-made functions for requests in the API folder
image

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,142 @@
...
-  </table>
-);
+          </th>
+          <th>Title</th>
+          <th> </th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {title.trim().length > 0 ? (
```
nazarmatsevych, line 62: This component looks too complex and has a lot of duplicated code, better to create preparedTodos variable inside App component and pass prepared (filtered) todos to TodoList

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,142 @@
...
+          findMatches.map((todoItem) => (
+            <tr
+              data-cy="todo"
+              className=""
```
nazarmatsevych, line 66: Remove attributes with an empty values

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,45 +1,96 @@
...
+  todoCard,
+  setIsModalOpen,
+  setIconId,
+}) => {
+  const [selectedUser, setsSelectedUser] = useState<User>();
+  const [isUserLoad, setIsUserLoad] = useState(false);
+
+  useEffect(() => {
+    setIsUserLoad(true);
+    setTimeout(() => {
```
nazarmatsevych, line 23: Same here, you already have ready-made functions for requests in the API folder

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,45 +1,96 @@
...
 
             <p className="block" data-cy="modal-user">
               {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              {todoCard?.completed ? (
+                <strong
+                  className="has-text-success"
+                >
+                  Done
+                </strong>
```
nazarmatsevych, line 80:
```suggestion
                <strong className="has-text-success">Done</strong>
```

this code is quite short, so we can put it in one line

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,45 +1,96 @@
...
+              ) : (
+                <strong
+                  className="has-text-danger"
```
nazarmatsevych, line 83: same as above

src/services/getTodos.ts
```diff
@@ -0,0 +1,7 @@
+export function getTodos() {
+  return fetch('https://mate-academy.github.io/'
+  + 'react_dynamic-list-of-todos/api/todos.json')
+    .then((response) => {
+      return response.json();
+    });
+}
```
nazarmatsevych, line 7: Redundant, check the API folder

nazarmatsevych (changes requested):
Good start, please check the comments below

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,45 +1,82 @@
...
+  todoCard,
+  setIsModalOpen,
+  setIconId,
+}) => {
+  const [selectedUser, setsSelectedUser] = useState<User>();
+  const [isUserLoad, setIsUserLoad] = useState(false);
+
+  useEffect(() => {
+    setIsUserLoad(true);
+    getUser(todoCard?.userId).then((user) => {
```
TarasHoliuk, line 24: it's better to check `if (todoCard?.userId)` and then fetch the user using this function instead of passing `undefined`. It anyway won't return valid info with `undefined`

src/App.tsx
```diff
@@ -17,18 +36,42 @@ export const App: React.FC = () => {
...
+                  setIconId={setIconId}
+                />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {isModalOpen && (
```
TarasHoliuk, line 67: you could use `todoCard` instead of a special state variable `isModalOpen`. If `todoCard` is not `null` it means the user selected one and we should open the modal. Also, in this case, you would be sure that `todoCard` is always defined in the `TodoModal` component

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,73 @@
...
+            <option value={FilteringType.active}>Active</option>
+            <option value={FilteringType.completed}>Completed</option>
+          </select>
+        </span>
+      </p>
+
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          value={input}
+          onChange={(e) => handleInput(e)}
```
TarasHoliuk, line 48:
```suggestion
          onChange={handleInputChange}
```
additional function wrapper is redundant, also better name would be `handleInputChange`

src/App.tsx
```diff
@@ -1,14 +1,33 @@
...
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [todosLoading, setTodosLoading] = useState(false);
+  const [filter, setFilter] = useState<FilteringType>(FilteringType.all);
+  const [title, setTitle] = useState('');
+  const [todoCard, setTodoCard] = useState<Todo>();
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [iconId, setIconId] = useState(0);
```
TarasHoliuk, line 21: The name of this variable is a bit misleading - you store the id of the selected todo, not the id of some icon. But anyway, this state is redundant - you may use `todoCard.id` instead of `iconId`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,105 @@
...
-      </tr>
+  const findMatches = filtering
+    .filter((todo) => todo.title.trim().toLowerCase()
+      .includes(title.trim().toLowerCase()));
+
+  const handleEyeButton = (todo: Todo) => {
+    setTodoCard(todo);
+    setIsModalOpen(true);
+    setIconId(todo.id);
+  };
```
TarasHoliuk, line 44: it's better to define such a function in the `App.tsx` and pass it as a prop here instead of passing 3 different "setState" functions

src/App.tsx
TarasHoliuk: it could also help you to fix this test:
image

TarasHoliuk (changes requested):
Hey! GJ :+1: 

Let's fix 1 last test
Also, left a few more comments about state and props management

src/components/TodoList/TodoList.tsx
OPTIMISTIXX: can I leave this function if it passed 1 setState function?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,73 @@
...
+    setInput('');
+  };
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            value={selectedButton}
+            onChange={(value) => handleSelect(value)}
```
lerastarynets, line 35:
```suggestion
            onChange={handleSelect}
```
If a mentor points out a mistake, it means that you need to fix ALL occurrences of the

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,73 @@
...
+  selectedButton,
+  setSelectedButton,
+}) => {
+  function handleInput(e: React.BaseSyntheticEvent) {
+    setInput(e.target.value);
+  }
+
+  const handleSelect = (value: React.ChangeEvent<HTMLSelectElement>) => {
+    setSelectedButton(value.target.value as FilteringType);
+  };
```
lerastarynets, line 23:
```suggestion
  const handleSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
    setSelectedButton(event.target.value as FilteringType);
  };
```

lerastarynets (approved):
Great job overall! Left a few comments to make your code more advanced)

PR #2108, opened 2023-12-30
add dynamic list of todos

Vlad4567:
- [DEMO LINK](https://Vlad4567.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,30 +9,25 @@ import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 import { getTodos } from './api';
 import { Filter } from './types/enum/Filter';
-import { TodosState } from './types/TodosState';
-
-const defaultState: TodosState = {
-  todos: [],
-  filter: Filter.All,
-  query: '',
-};
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
-  const [todosState, setTodosState] = useState<TodosState>(defaultState);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filter, setFilter] = useState<Filter>(Filter.All);
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,41 @@
...
+import { TodosState } from './types/TodosState';
+
+const defaultState: TodosState = {
+  todos: [],
+  filter: Filter.All,
+  query: '',
+};
 
 export const App: React.FC = () => {
+  const [todosState, setTodosState] = useState<TodosState>(defaultState);
```
etojeDenys, line 21: better to split it into three states, 'cause they are not connected and can be like separate states

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,108 @@
...
+
+                case Filter.Active:
+                  return !todo.completed;
+
+                case Filter.Completed:
+                  return todo.completed;
+
+                default:
+                  return todo;
+              }
```
etojeDenys, line 52: decide before jsx what text will be rendered and just render in jsx.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+
+interface Props {
+  todoId: number
+  setTodoId: (currentTodoId: number | null) => void
+  todos: Todo[]
```
etojeDenys, line 10: why do we need to pass the whole array here, let's pass only selectedTodo

etojeDenys (changes requested):
fix you demo link

etojeDenys (approved):
gj

PR #2111, opened 2024-01-02
Solution

web3-sorcerer:
[DEMO LINK](https://web3-sorcerer.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,6 +1,18 @@
-import React from 'react';
+import classNames from 'classnames';
+import { FC } from 'react';
+import { Todo } from '../../types/Todo';
 
-export const TodoList: React.FC = () => (
+type Props = {
+  todos: Todo[],
+  selected: Todo | null,
+  onSelected: (todo: Todo | null) => void,
+};
+
+export const TodoList: FC<Props> = ({
+  todos,
+  selected,
+  onSelected,
+}) => (
   <table className="table is-narrow is-fullwidth">
     <thead>
       <tr>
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,59 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+          <td className="is-vcentered is-expanded">
+            <p className="has-text-danger">delectus aut autem</p>
```
mbruhler, line 51: you should delete this

mbruhler (changes requested):
image
Everything works but there is minor bug with "delectus ..." string in title column, it shouldnt be there according to the mockup

Timi2424 (approved):

src/components/TodoList/TodoList.tsx
danon321: Still have that additional column

danon321 (changes requested):
Still have that additional column in demo link

marcinzrodlowski:
That's weird because I removed these lines...

marcinzrodlowski:
Ok, now it works I just had to re-deploy it again...

mbruhler (changes requested):
Last one has been solved, but still there is a flickering issue when opening the modal

https://github.com/mate-academy/react_dynamic-list-of-todos/assets/21124163/05fab18d-7f5f-4c33-a8aa-fa348a49de88

Please fix this in order to approve this task as it's not proper and doesn't work as in demo app.

You propably can't see it in the video due to frame rate and the speed of this flicekring  but check this out in your app

danon321 (changes requested):
image
As u can see on photo in loading u have additional close button

mpiskadlo (approved):
All seems to be good now!

PR #2131, opened 2024-01-23
Solution

GeorgeSavostikov:
[DEMO LINK](https://GeorgeSavostikov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,44 @@
...
+    getTodos().then(setTodos).finally(() => setLoading(false));
+  }, []);
+
+  function getFilterTodos() {
+    return todos
+      .filter((todo) => todo.title.toLowerCase().includes(query.toLowerCase()))
+      .filter((todo) => {
+        switch (filter) {
+          case 'active':
+            return todo.completed === false;
```
SanyaBratashchuk, line 31:
```suggestion
            return !todo.completed;
```

src/App.tsx
```diff
@@ -1,14 +1,44 @@
...
+
+          case 'completed':
+            return todo.completed === true;
```
SanyaBratashchuk, line 34:
```suggestion
            return todo.completed;
```

src/App.tsx
```diff
@@ -1,14 +1,44 @@
...
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos().then(setTodos).finally(() => setLoading(false));
+  }, []);
+
+  function getFilterTodos() {
+    return todos
+      .filter((todo) => todo.title.toLowerCase().includes(query.toLowerCase()))
+      .filter((todo) => {
```
SanyaBratashchuk, line 28: it is better to move filter inside switch case not vice verse to avoid an extra iteration when filter is `all`

SanyaBratashchuk (changes requested):

etojeDenys (approved):
lgtm

PR #2133, opened 2024-01-25
react_dynamic-list-of-todos

IShamkii:


src/App.tsx
```diff
@@ -1,14 +1,25 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, {
+  useContext,
+  // useEffect,
+  // useState,
```
anastasiia-tilikina, line 5: delete commented code

src/components/Store/Store.tsx
```diff
@@ -0,0 +1,65 @@
...
+    setLoading(true);
+
+    setTimeout(() => {
+      getTodos()
+        .then(setTodos)
+        .catch(() => {})
+        .finally(() => {
+          setLoading(false);
+        });
+    }, 100);
```
anastasiia-tilikina: why do you need timeout here?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,81 @@
...
+  const [title, setTitle] = useState('');
+  const [filter, setFilter] = useState(FilterParams.All);
+
+  const filteredTodos = useCallback(() => {
+    return getFilteredTodos(todos, filter, title);
+  }, [todos, filter, title]);
+
+  useEffect(() => {
+    setFilteredTodos(filteredTodos());
+  }, [filter, title]);
```
anastasiia-tilikina: you don't need second state for todos - filteredTodos.
Create state filter and query in your context. in your todolist you will assign the result of getFilteredTodos into variable and render it. every time your filter or query will change the component will rerender and variable value will updated. no need to make some side effects here

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,81 @@
...
+
+  const handleTitle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value.toLowerCase());
```
anastasiia-tilikina, line 27: do not transform input value to lowerCase. Now I'm not able to type uppercased text in the input. transform to one case in your getFilteredTodos fn.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,48 @@
...
+
+  const closeModal = () => {
+    setSelectedTodo(null);
+  };
+
+  useEffect(() => {
+    if (userId) {
+      setLoadingUser(true);
+
+      setTimeout(() => { // loading too fast
```
anastasiia-tilikina, line 30: I believe we can remove this timeout as well

src/services/getfilteredTodos.ts
```diff
@@ -0,0 +1,33 @@
...
+    filteredTodos = filteredTodos.filter(todo => {
+      switch (filterParams) {
+        case FilterParams.Active:
+          return !todo.completed;
+
+        case FilterParams.Completed:
+          return todo.completed;
+
+        default:
+          return todo;
```
anastasiia-tilikina, line 21:
```suggestion
          return true;
```
the return value of filter callback must be a boolean

anastasiia-tilikina (changes requested):
well done!
let's just make some fixes here

src/components/Store/Store.tsx
IShamkii: failed first test

IShamkii: > why do you need timeout here?

failed first test

IvanFesenko (approved):
well done

PR #2145, opened 2024-01-29
dynamic list of todos

vpashko:
[DEMO LINK](https://vpashko.github.io/react_dynamic-list-of-todos/)


EdenParadise (approved):
Looks good, great job!

src/App.tsx
```diff
@@ -1,34 +1,82 @@
...
+import { TodoContext } from './contexts/TodoContext';
 
 export const App: React.FC = () => {
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const { selectedTodo } = useContext(TodoContext);
+
+  const [isLoading, setIsLoading] = useState(true);
+
+  const [query, setQuery] = useState('');
+  const [status, setStatus] = useState('all');
```
olya-shyian, line 25: use enum Status for `all`

src/App.tsx
```diff
@@ -1,34 +1,82 @@
...
+        <div className="block">
+          <TodoFilter
+            query={query}
+            setQuery={setQuery}
+            setStatus={setStatus}
+          />
+        </div>
+
+        <div className="block">
+          {isLoading && <Loader />}
```
olya-shyian, line 72: if you can avoid creating setState then do it, in this case you can do check if `preparedTodos.length > 0 ` instead of `isLoading` state

rewrite please

src/App.tsx
```diff
@@ -1,34 +1,82 @@
...
+        case Status.active:
+          return !todo.completed;
+        case Status.completed:
+          return todo.completed;
+        default:
+          return todo;
+      }
+    }),
+    [status, filteredByQuery],
+  );
```
olya-shyian, line 56: move `filteredByQuery` and  this filter function to `utils` or `helpers` folder

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,68 @@
...
-        <button
-          data-cy="clearSearchButton"
-          type="button"
-          className="delete"
+export const TodoFilter: React.FC<Props> = ({
+  query,
+  setQuery,
+  setStatus,
+}) => {
+  const selectHandler = (event: React.ChangeEvent<HTMLSelectElement>) => {
```
olya-shyian, line 14: give more clear name...what do we select?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,35 @@
...
-      <tr data-cy="todo" className="">
-        <td className="is-vcentered">1</td>
-        <td className="is-vcentered" />
-        <td className="is-vcentered is-expanded">
-          <p className="has-text-danger">delectus aut autem</p>
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
+export const TodoList: React.FC<Props> = ({ todosFromServer }) => {
+  return (
```
olya-shyian, line 10: when you have no logic in component and just return jsx you can delete the word `return` and just return jsx in round brackets

`}) => (`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,38 @@
...
+
+  const { todo } = selectedTodo;
+
+  useEffect(() => {
+    setTimeout(() => {
+      setIsLoading(false);
+    }, 300);
+
+    getUser(todo.userId).then(setUser);
+  },
```
olya-shyian, line 21: what will happen if the user's download takes longer than 300ms?
you should still show the loader...

so the logic should be like that

```
  useEffect(() => {
    if (todo.id) {
      setUserLoading(true);
    }

    getUser(todo.userId)
      .then(setUser)
      .finally(() => setUserLoading(false));
  }, [todo.userId]);
  ```

File: src/types/Status.ts [hunk 7]
```diff
@@ -0,0 +1,5 @@
+export enum Status {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
```
olya-shyian, line 4: please write all const with first upperCase letter
`active =` to `Active =`

image

olya-shyian (changes requested):
Nicely done!💥

p.s. demo link doesn't work(

vpashko:
> [DEMO LINK](https://vpashko.github.io/react_dynamic-list-of-todos/)

src/App.tsx
vpashko: I left the state with the loader because this check is used in many places, for example, if we filter the array and find nothing, we don't need to show the loader indefinitely and the check for the length of the array will not pass here.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,83 @@
...
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+          return (
+            <tr data-cy="todo" className="" key={id}>
+              <td className="is-vcentered">{id}</td>
+              <td className="is-vcentered">
+                {completed && (
```
DanilWeda, line 43:
```suggestion
                {!!completed && (
```
in rendering with &&  need to convert the value to boolean.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,72 @@
...
+  };
+
+  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    const inputQuery = e.target.value.toLowerCase().trim();
+
+    setQuery(e.target.value);
+    setFilter(prev => ({ ...prev, title: inputQuery }));
+  };
+
+  const handleClick = () => {
```
DanilWeda, line 22: handleReset maybe?

DanilWeda (approved):
Great. 
Approved with small comments. 
Thanks!

PR #2148, opened 2024-01-30
Done!(try1)

EdenParadise:
[DEMO LINK](https://EdenParadise.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,34 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { getFilteredTodos } from './services/getFilteredTodos';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState<Filter>({ status: 'all', title: '' });
+
+  const filteredTodos = getFilteredTodos(todos, filter);
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(setTodos)
+      .finally(() => setLoading(false));
+  }, []);
+
+  const handleTodoSelect = (todo: Todo) => {
+    setSelectedTodo(todo);
+  };
+
+  const resetTodo = () => {
+    setSelectedTodo(null);
+  };
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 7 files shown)


src/components/TodoFilter/TodoFilter.tsx
artem8746, line 1: filter input doesn't allow adding spaces

src/App.tsx
```diff
@@ -1,14 +1,40 @@
...
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { getFilteredTodos } from './services/getFilteredTodos';
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
```
artem8746, line 18: would be better to create TodosContext and get todos, filterOptions and selectedTodo from there instead of drilling it in props

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,72 @@
...
+
+export const TodoFilter: React.FC<Props> = ({ setFilter }) => {
+  const [query, setQuery] = useState('');
+
+  const handleSelectChange = (e: ChangeEvent<HTMLSelectElement>) => {
+    setFilter(prev => ({ ...prev, status: e.target.value }));
+  };
+
+  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    const inputQuery = e.target.value.toLowerCase().trim();
```
artem8746, line 16: do not normalize query here, do it in filter to allow adding spaces and capital letters in input

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,83 @@
...
+                        'fa-eye': selectedTodoId !== id,
+                        'fa-eye-slash': selectedTodoId === id,
+                      })}
+                    />
+                  </span>
+                </button>
+              </td>
+            </tr>
+          );
+        })}
```
artem8746, line 79: would be better to create component for this `TodoItem`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
...
   return (
-    <div className="modal is-active" data-cy="modal">
+    <div
+      className="modal is-active"
+      data-cy="modal"
+    >
       <div className="modal-background" />
 
-      {true ? (
+      {loading ? (
```
artem8746, line 31:
```suggestion
      {!selectedUser? (
```
you can just use `selectedUser` and you won't need `loading`

artem8746 (changes requested):
Amazing job!!
Just a few fixes needed

src/App.tsx
```diff
@@ -17,18 +43,18 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter setFilter={setFilter} />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {loading ? <Loader /> : (
```
olya-shyian, line 50: instead of creating additional state its better to do like Artem said, 

` {filteredTodos.length > 0 ?   <TodoList ... > : <Loader /> `

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,34 +1,72 @@
...
+
```
olya-shyian, line 17: you do .toLowerCase() also in `getFilteredTodos()`
keep this preparation in one place

image

olya-shyian (changes requested):
Great!🙂

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,86 @@
...
+              </p>
+            </td>
+            <td className="has-text-right is-vcentered">
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => {
+                  setSelectedTodo(todo);
+                  setShowModal(!showModal);
```
olya-shyian, line 69: you can avoid creating `showModal` state and use `selectedTodo` for that

like if `selectedTodo` is true (checked) then show `TodoModal`, if `null` then - don't

olya-shyian (approved):
Good!✨

PR #2164, opened 2024-02-01
add task solution

dyarechuk:
[DEMO LINK](https://dyarechuk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,6 +1,18 @@
+import classNames from 'classnames';
 import React from 'react';
+import { Todo } from '../../types/Todo';
 
-export const TodoList: React.FC = () => (
+type Props = {
+  todos: Todo[];
+  selectedTodo?: Todo | null;
+  onTodoSelected?: (todo: Todo) => void;
+};
+
+export const TodoList: React.FC<Props> = ({
+  todos = [],
+  selectedTodo = null,
+  onTodoSelected = () => { },
+}) => (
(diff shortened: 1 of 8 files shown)


YevhenOstrovskiy (approved):
Well done! You did a good job))

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,18 @@
+import classNames from 'classnames';
 import React from 'react';
+import { Todo } from '../../types/Todo';
 
-export const TodoList: React.FC = () => (
+type Props = {
+  todos: Todo[];
+  selectedTodo?: Todo | null;
```
olya-shyian, line 7: there is no need to add mark `?` because you wrote a default value ` selectedTodo = null` in `TodoList`
so anyway it will be null if we don't pass any todo

src/services/prepareTodos.ts
```diff
@@ -0,0 +1,17 @@
...
+export function prepareTodos(todos: Todo[], filterField: FilterField) {
+  const { status, title } = filterField;
+  const filteredByTitle = todos
+    .filter(todo => todo.title.toLowerCase().includes(title));
+
+  switch (status) {
+    case 'active':
+      return filteredByTitle.filter(({ completed }) => !completed);
+    case 'completed':
+      return filteredByTitle.filter(({ completed }) => completed);
```
olya-shyian, line 13: it would be great if you create enum for 'active', 'completed', 'all'

olya-shyian (approved):
Coool!💪

PR #2169, opened 2024-02-05
add task solution

PodvAx:
[DEMO LINK](https://PodvAx.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+          return true;
+        case Status.Active:
+          return !todo.completed;
+        case Status.Completed:
+          return todo.completed;
+        default:
+          return true;
+      }
+    })
+    .filter((todo) => todo.title.toLowerCase().includes(search.toLowerCase()));
```
Esceype, line 43:
```suggestion
  const filteredTodos = todos.filter(todo => {
      const includesQuery = todo.title.toLocaleLowerCase()
        .includes(query.toLocaleLowerCase());

      switch (status) {
        case Status.Active:
          return !todo.completed && includesQuery;

        case Status.Completed:
          return todo.completed && includesQuery;

        case Status.All:
        default:
          return includesQuery;
      }
    });
```

No need to use `filter` twice

Esceype (changes requested):
Good job!
Just one small comment

SanyaBratashchuk (approved):

PR #2190, opened 2024-02-20
solution

katya-sn:
https://katya-sn.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/utils/getFilteredTodos.tsx
+++ b/src/utils/getFilteredTodos.tsx
@@ -1,0 +1,23 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
+export const getFilteredTodos = (
+  todos: Todo[],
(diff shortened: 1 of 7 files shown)


src/utils/getFilteredTodos.tsx
```diff
@@ -0,0 +1,23 @@
...
+    )
+    .filter(({ completed }) => {
+      switch (filter) {
+        case Status.active:
+          return !completed;
+        case Status.completed:
+          return completed;
+        default:
+          return todos;
+      }
```
volodymyr-soltys97, line 21:
```suggestion
      switch (filter) {
        case Status.active:
          return !completed;
          
        case Status.completed:
          return completed;
          
        default:
          return todos;
      }
```

volodymyr-soltys97 (approved):
Good job 👍

PR #2200, opened 2024-03-01
todo

edkido:
[DEMO LINK](https://edkido.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useMemo, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
+
+function getPreparedTodos(todoList: Todo[], select: Status, query: string) {
+  const visibleTodos = [...todoList];
+
+  return visibleTodos
```
etojeDenys, line 17:
```suggestion
  return todoList
```
filter doesn't mutate the array, so this copy is redundant here

etojeDenys (approved):
good job

PR #2201, opened 2024-03-02
react_dynamic-list-of-todos

TetianaMD:
[DEMO LINK](https://TetianaMD.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,66 @@
 /* eslint-disable max-len */
-import React from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-
+import { useEffect, useState } from 'react';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadingTodos, setLoadingTodos] = useState(false);
+  const [query, setQuery] = useState('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [status, setStatus] = useState('');
+
+  const handleStatusChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    setStatus(e.target.value);
+  };
+
+  const filteredTodos = todos.filter(todo =>
+    todo.title.toLowerCase().includes(query),
+  );
+
+  const filteredTodosByStatus = filteredTodos.filter(todo => {
+    if (status === 'all') {
+      return true;
+    }
+
+    if (status === 'active') {
+      return todo.completed === false;
+    }
+
+    if (status === 'completed') {
+      return todo.completed === true;
+    }
+
+    return true;
+  });
+
+  const handleAllTodos = (input: string) => {
+    setQuery(input.toLowerCase());
+  };
+
+  const handleClear = () => {
+    setSelectedTodo(null);
+    setQuery('');
+  };
+
+  const handleShowModal = (todo: Todo) => {
+    setSelectedTodo(todo);
+  };
+
+  useEffect(() => {
+    setLoadingTodos(true);
+
+    getTodos()
+      .then(setTodos)
+      .finally(() => setLoadingTodos(false));
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
+
+    if (status === 'completed') {
+      return todo.completed === true;
+    }
+
+    return true;
+  });
+
+  const handleAllTodos = (input: string) => {
+    setQuery(input.toLowerCase());
```
TarasHoliuk, line 44: it means the user can't type big letters which might be a bit confusing

It's better to do `query.toLowerCase()` inside filter to get `filteredTodos` instead of doing it here

src/App.tsx
```diff
@@ -17,18 +69,27 @@ export const App: React.FC = () => {
...
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {loadingTodos && <Loader />}
+              <TodoList
+                todos={filteredTodosByStatus}
+                onShowModal={handleShowModal}
```
TarasHoliuk, line 85:
```suggestion
                onShowModal={setSelectedTodo}
```
`handleShowModal` function wrapper is redundant

Explanation: `onShowModal` prop expects a function that accepts `todo` as an argument. `setSelectedTodo` is such a function. Also, the result will be the same, because `handleShowModal` does nothing but just call `setSelectedTodo` inside

src/App.tsx
```diff
@@ -17,18 +69,27 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                onQueryChange={handleAllTodos}
```
TarasHoliuk, line 73:
```suggestion
                onQueryChange={setQuery}
```

src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadingTodos, setLoadingTodos] = useState(false);
+  const [query, setQuery] = useState('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [status, setStatus] = useState('');
```
TarasHoliuk, line 17: Let's  add an enum for `status`. Because we have only a few values acceptable for this state (`'all'`, `'active'`, and `'completed'`)

src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
+  const handleStatusChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    setStatus(e.target.value);
+  };
+
+  const filteredTodos = todos.filter(todo =>
+    todo.title.toLowerCase().includes(query),
+  );
+
+  const filteredTodosByStatus = filteredTodos.filter(todo => {
+    if (status === 'all') {
```
TarasHoliuk, line 28: it's a good case to use `switch` block

Also, use enum for `status`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,25 +42,39 @@ export const TodoModal: React.FC = () => {
...
+                <strong
+                  className="has-text-success"
+                  >
+                  Done
+                </strong>
+              )
+              : (
+                <strong className="has-text-danger">Planned</strong>
+              )
+              }
```
TarasHoliuk, line 73:
```suggestion
              ) : (
                <strong className="has-text-danger">Planned</strong>
              )}
```

src/App.tsx
```diff
@@ -1,14 +1,66 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loadingTodos, setLoadingTodos] = useState(false);
```
TarasHoliuk, line 14: it's good practice to add `is` or `has`, etc. (depends on meaning of course) prefix to boolean variable names 

https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

TarasHoliuk (changes requested):
GJ :+1: 

Added a few comments to improve your code

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,72 @@
...
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {isTodos.map(todo => (
+          <tr key={todo.id} data-cy="todo" className="">
```
etojeDenys, line 31:
```suggestion
          <tr key={todo.id} data-cy="todo">
```

src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
+import { Todo } from './types/Todo';
+
+enum Status {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [isTodos, setIsTodos] = useState<Todo[]>([]);
```
etojeDenys, line 19:
```suggestion
  const [todos, setTodos] = useState<Todo[]>([]);
```

src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
+  const [isLoadingTodos, setIsLoadingTodos] = useState(false);
+  const [hasQuery, setHasQuery] = useState('');
+  const [isSelectedTodo, setIsSelectedTodo] = useState<Todo | null>(null);
```
etojeDenys, line 22:
```suggestion
  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
```

etojeDenys (approved):
lgtm

PR #2211, opened 2024-03-08
solution

Yehorf21:
[DEMO LINK](https://Yehorf21.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


Yehorf21:
I added conversion of input to lower case, but the Demo Link didn't get updated yet

src/App.tsx
```diff
@@ -1,34 +1,106 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todos);
+  const [isVisible, setIsVisible] = useState<Visibility>({
+    visible: false,
+    id: 0,
+  });
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    async function fetchData() {
```
IvanFesenko, line 26: You dont use await, so async is redundant
```suggestion
    function fetchData() {
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,88 @@
...
+import React, { useState } from 'react';
+import { Select } from '../../types/Select';
+
+interface Props {
+  onSelect: (val: Select) => void;
+  onInput: (value?: string) => void;
+}
+
+export const TodoFilter: React.FC<Props> = ({ onSelect, onInput }) => {
+  const [currentSelect, setCurrentSelect] = useState<Select>(Select.All);
```
IvanFesenko, line 10: I suggest move this state into App and use for get filtered todos, instead of filteredTodos state

IvanFesenko (changes requested):
Looks good, Check this case. In case query not empty it should filter
https://github.com/mate-academy/react_dynamic-list-of-todos/assets/61904995/5b97f2ae-9c11-4598-b089-4c64ef7fe65a

PR #2221, opened 2024-03-14
dynamic list of todos

roman-tarasov90:
ttps://roman-tarasov90.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,100 +1,127 @@
+import classNames from 'classnames';
 import React from 'react';
+import { Todo } from '../../types/Todo';
 
-export const TodoList: React.FC = () => (
-  <table className="table is-narrow is-fullwidth">
-    <thead>
-      <tr>
-        <th>#</th>
-        <th>
-          <span className="icon">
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,127 @@
...
+                    )}
+                  </span>
+                </button>
+              </td>
+            </tr>
+          ) : (
+            <tr
+              key={todo.id}
+              data-cy="todo"
+              className={classNames('', {
```
etojeDenys, line 95:
```suggestion
              className={classNames({
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,127 @@
...
+                      <i className="far fa-eye-slash" />
+                    ) : (
+                      <i className="far fa-eye" />
+                    )}
+                  </span>
+                </button>
+              </td>
+            </tr>
+          ) : (
+            <tr
```
etojeDenys, line 92: looks like you don't need the ternary operator here

etojeDenys (approved):
lgtm

PR #2234, opened 2024-03-19
solution

NeYaroslav:
[DEMO LINK](https://NeYaroslav.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodo, setFilteredTodo] = useState(todos);
```
etojeDenys, line 15: it can be the simple var

```
const filteredTodo = filterTodo()
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,104 @@
...
+    setFilterConfig(prevFilterConfig => ({
+      ...prevFilterConfig,
+      [name]: value,
+    }));
+  };
+
+  useEffect(() => {
+    onFilter(todos.filter(createFilterFn(filterConfig)));
+  }, [filterConfig, todos, onFilter]);
+
```
etojeDenys, line 57: move all filter states to the App component, and only change them here

etojeDenys (changes requested):
feel free to ask for some help in the chat

loralevitska (approved):
Good job!
It could be improved:
 - use [trimming](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) for filtering 
image

PR #2250, opened 2024-03-26
add task solution

creonluv:
[DEMO LINK](https://creonluv.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,33 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [originalTodos, setOriginalTodos] = useState<Todo[]>([]);
+  const [isTodoModalShown, setIsTodoModalShown] = useState(false);
+
+  useEffect(() => {
+    setLoading(true);
+
+    getTodos()
+      .then(data => {
+        setTodos(data);
+        setOriginalTodos(data);
+      })
+      .finally(() => setLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,33 @@
...
 import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [originalTodos, setOriginalTodos] = useState<Todo[]>([]);
```
anastasiia-tilikina, line 17: why do you need this state? as I see you're using only todos not this one

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,87 @@
...
+    setStatusFilter(event.currentTarget.value);
+  };
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.currentTarget.value);
+  };
+
+  const handleClearButton = () => {
+    setQuery('');
+    setStatusFilter('all');
```
anastasiia-tilikina, line 23: i believe this button should only clear query not filter

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,87 @@
...
+      );
+    }
+
+    if (statusFilter === 'active') {
+      filteredTodos = filteredTodos.filter(todo => !todo.completed);
+    } else if (statusFilter === 'completed') {
+      filteredTodos = filteredTodos.filter(todo => todo.completed);
+    }
+
+    setTodos(filteredTodos);
```
anastasiia-tilikina, line 41: you don't need to set this into state. every time you change filter or query your component will be re-rendered so filtered todos could be variable which will be recalculated on each render

anastasiia-tilikina: and you don't need to put this logic into useEffect.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,44 @@
...
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  selectedTodo: Todo;
+  setSelectedTodo: (todo: Todo | null) => void;
+  setLoading: (condition: boolean) => void;
```
anastasiia-tilikina, line 10: why do you put this props if you're using local loading state?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,25 +47,34 @@ export const TodoModal: React.FC = () => {
...
+              {selectedTodo.completed ? (
+                <strong className="has-text-success">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
 
               {' by '}
 
-              <a href="mailto:Sincere@april.biz">Leanne Graham</a>
+              <a href="mailto:Sincere@april.biz">{userData?.name}</a>
```
anastasiia-tilikina, line 77: shou should set actual user's email in href not hardcoded expample

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <button
+          data-cy="selectButton"
+          className="button"
+          type="button"
+          onClick={handleSelectButton}
+        >
+          <span className="icon">
+            <i
+              className={classNames('far', {
+                'fa-eye': todo !== selectedTodo,
```
anastasiia-tilikina, line 55:
```suggestion
                'fa-eye': todo.id !== selectedTodo.id,
```
Probably it is not a good idea to compare objects. It is safer to compare primitives. Not a mistake in this case.

anastasiia-tilikina (changes requested):
Let's fix some issues here

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+
+  const filteredTodos = todos.filter(todo => {
+    if (filter === 'all') {
+      return todo.title.toLowerCase().includes(query.trim().toLowerCase());
+    } else if (filter === 'active') {
+      return (
+        !todo.completed &&
+        todo.title.toLowerCase().includes(query.trim().toLowerCase())
+      );
+    } else if (filter === 'completed') {
```
VitaliyBondarenko1982, line 39: it is good place to use switch...case instead if...else.
better to create enum for filter.

VitaliyBondarenko1982 (approved):
Good job.

PR #2252, opened 2024-03-27
add task solution

RomanPerepelkin:
https://RomanPerepelkin.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,48 @@
...
+        return (
+          (status === 'completed'
+            ? todo.completed
+            : status === 'active'
+              ? !todo.completed
+              : todo) && todo.title.toLowerCase().includes(query.toLowerCase())
+        );
+      }),
+    );
+  }, [status, todos, query]);
```
IvanFesenko, line 44: wrong useMemo usage
```suggestion
  const [query, setQuery] = useState('');

  useEffect(() => {
    setLoading(true);
    getTodos()
      .then(setTodos)
      .finally(() => setLoading(false));
  }, []);

  const handleCloseButton = () => {
    setSelectedTodo(undefined);
  };

  const visibleTodos = useMemo(() => {
    // TODO add filtarion logic. Note: no need to filter in case status is all and query is empty.
    // Also do not use nesting ternary
  }, [status, todos, query]);
```

src/App.tsx
```diff
@@ -1,14 +1,48 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [loading, setLoading] = useState(false);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo>();
+  const [status, setStatus] = useState('all');
```
IvanFesenko, line 17: create enum for status and use instead of string

IvanFesenko (changes requested):
Good job. Lets improve your code

l4st1m0za (approved):
Good job!

PR #2255, opened 2024-03-27
Solution

winniepuhh:
[DEMO LINK](https://winniepuhh.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+enum FilterType {
+  Active = 'active',
+  Completed = 'completed',
+  All = 'all',
+}
+
+const filterTodos = (todos: Todo[], filterType: string, filterText: string) => {
+  let filteredTodos: Todo[] = [...todos];
+
+  if (filterType === FilterType.Active) {
```
illya-onyshchuk, line 22: Better use `switch case`.

illya-onyshchuk (approved):
Well done! 👍

src/App.tsx
BudnikOleksii: > Better use `switch case`.

Good point

src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+enum FilterType {
+  Active = 'active',
+  Completed = 'completed',
+  All = 'all',
+}
+
+const filterTodos = (todos: Todo[], filterType: string, filterText: string) => {
```
BudnikOleksii, line 19:
```suggestion
const filterTodos = (todos: Todo[], filterType: FilterType , filterText: string) => {
```
`filterType` isn't string

src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+    );
+  }
+
+  return filteredTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodods] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [filterType, setFilterType] = useState<string>('All');
```
BudnikOleksii, line 46: Use your enum `FilterType`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,76 @@
...
+                  setSelectedTodo(todo);
+                }}
+              >
+                <span className="icon">
+                  <i
+                    className={
+                      selectedTodo?.id === todo.id
+                        ? 'far fa-eye-slash'
+                        : 'far fa-eye'
+                    }
```
BudnikOleksii, line 66: Use classnames library to set classes conditionally

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    getUser(selectedTodo.userId).then(data => {
+      setUser(data);
+      setIsLoading(false);
+    });
+  }, [selectedTodo.userId]);
```
BudnikOleksii, line 24:
```suggestion
  const [isLoading, setIsLoading] = useState(false);

  useEffect(() => {
    setIsLoading(true);
    
    getUser(selectedTodo.userId).then(data => {
      setUser(data);
    })
    .finally(() => {
      setIsLoading(false);
    });
  }, [selectedTodo.userId]);
```
You should set loader before fetching and remove it in any case after the request, not only in successful

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

PR #2256, opened 2024-03-28
add solution

Vlobdam:
[DEMO LINK](https://vlobdam.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,51 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+type FilterOptions = {
+  query: string,
+  taskStatusFilter: TaskStatus,
+}
+
+const getFilteredTodos = (todos: Todo[], options: FilterOptions) => {
+  const {query,
+    taskStatusFilter} = options;
+  let filteredTodos = [...todos];
+
+  const normalizeStr = (str: string): string => str.trim().toLowerCase()
+
+  if (query) {
+    filteredTodos = todos.filter(todo => normalizeStr(todo.title).includes(normalizeStr(query)));
+  }
+
+  if (taskStatusFilter === 'completed') {
+    filteredTodos = todos.filter(todo => todo.completed);
+  }
+
+  if (taskStatusFilter === 'active') {
+    filteredTodos = todos.filter(todo => !todo.completed);
+  }
+
+  return filteredTodos;
+}
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isTodoArrayLoaded, setIsTodoArrayLoaded] = useState(false);
+  const [query, setQuery] = useState('');
+  const [taskStatusFilter, setTaskStatusFilter] = useState<TaskStatus>('all');
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsTodoArrayLoaded(true));
+  });
+
+  const filteredTodos = getFilteredTodos(todos, { query, taskStatusFilter });
+
   return (
     <>
       <div className="section">
@@ -17,18 +60,26 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                query={query}
+                setQuery={setQuery}
+                setTaskStatusFilter={setTaskStatusFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isTodoArrayLoaded ? (
+                <TodoList
+                  todos={filteredTodos}
+                  selectedTodo={selectedTodo}
+                  selectTodo={setSelectedTodo}
+                />) : <Loader />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo && <TodoModal setSelectedTodo={setSelectedTodo} selectedTodo={selectedTodo} />}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -36,7 +36,6 @@ export const TodoFilter: React.FC<Props> = ({
       </span>
 
       <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
 
         {query && (
         <button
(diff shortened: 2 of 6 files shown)


VitaliiVashkevych (approved):
LGTM 🚀

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+type FilterOptions = {
+  query: string;
+  taskStatusFilter: TaskStatus;
+};
+
+const getFilteredTodos = (todos: Todo[], options: FilterOptions) => {
+  const { query, taskStatusFilter } = options;
+  let filteredTodos = [...todos];
+
+  const normalizeStr = (str: string): string => str.trim().toLowerCase();
```
BudnikOleksii, line 23: Move this function outside

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+    filteredTodos = todos.filter(todo =>
+      normalizeStr(todo.title).includes(normalizeStr(query)),
+    );
+  }
+
+  if (taskStatusFilter === 'completed') {
+    filteredTodos = todos.filter(todo => todo.completed);
+  }
+
+  if (taskStatusFilter === 'active') {
```
BudnikOleksii, line 35: - Use switch case instead
- Use your enum `TaskStatus`

src/types/types.ts
```diff
@@ -0,0 +1 @@
+export type TaskStatus = 'completed'| 'active' | 'all';
```
BudnikOleksii, line 1: Change it to enum

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+
+  return filteredTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isTodoArrayLoaded, setIsTodoArrayLoaded] = useState(false);
+  const [query, setQuery] = useState('');
+  const [taskStatusFilter, setTaskStatusFilter] = useState<TaskStatus>('all');
```
BudnikOleksii, line 47: `TaskStatus` should be enum

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+    filteredTodos = todos.filter(todo => !todo.completed);
+  }
+
+  return filteredTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isTodoArrayLoaded, setIsTodoArrayLoaded] = useState(false);
```
BudnikOleksii, line 45:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
To show the loader you should set this state to true when you start loading and then remove it after you get a response from the server in finally  block

src/App.tsx
```diff
@@ -17,18 +62,34 @@ export const App: React.FC = () => {
...
+              ) : (
+                <Loader />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo !== null && (
```
BudnikOleksii, line 87:
```suggestion
      {selectedTodo && (
```
You don't need to compare with null, https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,23 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
+import React from 'react';
+
+interface Props {
+  query: string;
+  setQuery: (arg) => void;
+  setTaskStatusFilter: (arg) => void;
```
BudnikOleksii, line 6: What is `arg`? And what type does it have?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,23 @@
...
+}
+
+export const TodoFilter: React.FC<Props> = ({
+  query,
+  setQuery,
+  setTaskStatusFilter,
+}) => (
+  <form className="field has-addons" onSubmit={event => event.preventDefault()}>
```
BudnikOleksii, line 14: Why? What is the point of onSubmit handler in this case?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,85 @@
 import React from 'react';
+import { Todo } from '../../types/Todo';
+import cn from 'classnames';
 
-export const TodoList: React.FC = () => (
+interface Props {
+  todos: Todo[];
+  selectedTodo: Todo;
+  selectTodo: (event) => void;
```
BudnikOleksii, line 8: - Specify types for arguments, and fix it in all places
- Prop name for handlers should start from `on` - `onSelectTodo` for example. https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364
- Why does it accept event?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  selectedTodo,
+  setSelectedTodo,
+}) => {
+  const [isUserLoaded, setIsUserLoaded] = useState(false);
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    getUser(selectedTodo.userId)
+      .then(setUser)
+      .finally(() => setIsUserLoaded(true));
```
BudnikOleksii, line 23: The same with loading state

BudnikOleksii (changes requested):
Good start, but a lot of work has to be done. Check the comments

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+import { TaskStatus } from './types/types';
+
+type FilterOptions = {
+  query: string;
+  taskStatusFilter: TaskStatus;
+};
+
+const getFilteredTodos = (todos: Todo[], options: FilterOptions) => {
+  const { query, taskStatusFilter } = options;
+  let filteredTodos = [...todos];
```
BudnikOleksii, line 21: Filter doesn't mutate an array so you don't need to create a copy

src/App.tsx
BudnikOleksii: > Move this function outside

Not fixed

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+      return filteredTodos.filter(todo => todo.completed);
+    default:
+      return filteredTodos;
+  }
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isLoaded, setIsLoaded] = useState(false);
```
BudnikOleksii, line 44:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
To show the loader you should set this state to true when you start loading and then remove it after you get a response from the server in finally block

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,27 @@
-export const TodoFilter = () => (
+import React from 'react';
+import { TaskStatus } from '../../types/types';
+
+interface Props {
+  query: string;
+  setQuery: (_: string) => void;
+  setTaskStatusFilter: (_: TaskStatus) => void;
```
BudnikOleksii, line 7: `_` is a bad name, you can use it only if you need to skip the argument and use the next one, in this case it doesn't make sense

src/components/TodoModal/TodoModal.tsx
BudnikOleksii: > The same with loading state

Not fixed

BudnikOleksii (changes requested):
Some of the comments still aren't fixed 🤷

BudnikOleksii (approved):
👍

PR #2257, opened 2024-03-28
add task solution

pawelnowicki87:
[DEMO LINK](https://pawelnowicki87.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const [filterType, setFilterType] = useState<string>('All');
+  const [filterText, setFilterText] = useState('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos().then(data => {
+      setTodods(data);
+      setIsLoading(false);
```
yevhenii-pyl, line 54: What if there will be an error? 
Good idea with adding loaders, but to make this usage fully valid lets move setting loading to false at the `finally` block and add an error catcher.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,74 @@
...
+          <th>Title</th>
+          <th> </th>
+        </tr>
+      </thead>
+      {/* "has-text-danger" */}
+      <tbody>
+        {todos.map(todo => (
+          <tr key={todo.id} data-cy="todo" className="">
+            <td className="is-vcentered">{todo.id}</td>
+            {todo.completed ? (
```
yevhenii-pyl, line 33: As a matter of personal preference, I'd suggest to split this code into 2 chunks: 
1. Todo is completed 
2. Todo is not completed 

Adding a ternary expression into mapping here may only reduce the overall readability of the code. 
But once again, it's only my opinion here.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+
+export const TodoModal: React.FC<Props> = ({
+  selectedTodo,
+  setSelectedTodo,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    getUser(selectedTodo.userId).then(data => {
```
yevhenii-pyl, line 20: Same with loaders here

yevhenii-pyl (changes requested):
Good job! 👍 
Just a few minor fixes and we are good to go.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,76 @@
...
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+export const TodoList: React.FC<TodoListProps> = ({
+  todos,
+  selectedTodo,
+  setSelectedTodo,
+}) => {
+  const completedTodo = (
```
yevhenii-pyl, line 14: We can't do it like that in React. 
If you want to go that way, there should be a separate component or a helper function, like `renderCompletedTodo =() => `. 

That's a good idea, but in this case my point was just to split the ternary expression you had into 2 chunks:
`todo.completed ? <></> : <></> `=> 
```
todo.completed && <></>
!todo.completed && <></>
```

If you want to take it step further - sure, that sounds like a plan, but we have to do it properly.

yevhenii-pyl (changes requested):
Let's simplify or create separate components for it, we shouldn't create elements like that in React.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,55 @@
...
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          value={filterText}
+          placeholder="Search..."
+          onChange={event => setFilterText(event.target.value)}
+        />
+        <span className="icon is-left">
```
BogdanMaliuta, line 38: add empty line detween sibling elements

BogdanMaliuta (approved):
GJ!Check my comments

PR #2258, opened 2024-03-28
task done

Oleksandr2108:
[DEMO LINK](https://Oleksandr2108.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -17,18 +43,32 @@ export const App: React.FC = () => {
...
+                setQuery={setQuery}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              {!isLoading && todos.length > 0 && (
+                <TodoList
```
DenNice-cloud, line 57: I think, will be correctly like this
```suggestion
              {isLoading ? (
                <Loader />
              ) : (
                <TodoList ...
```

DenNice-cloud (approved):
Everything looks good, just few comments and strange 1 error test,
```Expected to find element: `[data-cy="modal"]```,
but you got it, its weird
Keep it up! 🔥🔥🔥

File: src/components/TodoFilter/TodoFilter.tsx [hunk 2]
```diff
@@ -1,30 +1,62 @@
...
+          <select
+            data-cy="statusSelect"
+            value={filter}
+            onChange={e => {
+              setFilter(e.target.value as Status);
+            }}
+          >
+            <option value={Status.All}>{Status.All}</option>
+            <option value={Status.Active}>{Status.Active}</option>
+            <option value={Status.Completed}>{Status.Completed}</option>
```
BudnikOleksii, line 29: You can use the enum values with `Object.values` to create an array and then map through it to create options

BudnikOleksii (changes requested):
Good job 👍 

src/App.tsx
```diff
@@ -1,14 +1,43 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState<Status>(Status.All);
+  const [query, setQuery] = useState('');
+
+  let visibleTodos = getFilterTodos([...todos], filter);
+
+  if (query) {
+    visibleTodos = todosFilter(visibleTodos, query);
+  }
```
BudnikOleksii, line 32: I meant to remove filter logic from the component at all. And You don't need to do a copy manually as filter doesn't mutate an array

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+import { getUser } from '../../api';
+
+type Props = {
+  selectedTodo: Todo | null;
```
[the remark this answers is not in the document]
BudnikOleksii: > You can do it as required, as without `selectedTodo` this modal doesn't make much sense

Not fixed

BudnikOleksii (approved):
Approved, but check the comments

PR #2261, opened 2024-03-28
Solution

VitaliiVashkevych:
[DEMO LINK](https://VitaliiVashkevych.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,27 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { getPreparedTodos } from './components/services/getPreparedTodos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setLoading] = useState(true);
+
+  const [query, setQuery] = useState('');
+  const [filterOption, setFilterOption] = useState('all');
+
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  const visibleTodos = getPreparedTodos(todos, query, filterOption);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => setLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +36,32 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                query={query}
+                setQuery={setQuery}
+                filterOption={filterOption}
+                setFilterOption={setFilterOption}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              <TodoList
+                todos={visibleTodos}
+                setSelectedTodo={setSelectedTodo}
+                selectedTodo={selectedTodo}
+              />
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal
+          selectedTodo={selectedTodo}
+          setSelectedTodo={setSelectedTodo}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,30 +1,59 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import React from 'react';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  query: string;
+  setQuery: (value: string) => void;
+  filterOption: string;
+  setFilterOption: (value: string) => void;
+};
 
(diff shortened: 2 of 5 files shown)


src/components/services/getPreparedTodos.ts
```diff
@@ -0,0 +1,29 @@
...
+export const getPreparedTodos = (
+  todos: Todo[],
+  query: string,
+  filterOption: string,
+) => {
+  let todosCopy = [...todos];
+
+  if (query) {
+    todosCopy = todosCopy.filter(todo =>
+      todo.title.trim().toLowerCase().includes(query.trim().toLowerCase()),
```
vladarskyi, line 12: Search in the working version works without `.trim()`.

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,45 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+        return (
+          <tr data-cy="todo" className="" key={id}>
```
vladarskyi, line 35: You need to add a class `has-background-info-light` to the selected todo:
missed class

src/App.tsx
```diff
@@ -1,14 +1,33 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { getPreparedTodos } from './components/services/getPreparedTodos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setLoading] = useState(true);
```
vladarskyi, line 16: I am not sure about this, but both in the video and on QnA it was noted that the default loader should be set to `false`, and changes should be made already in `useEffect` (set `true` before `getTodos()`  and then again `false`)

vladarskyi (changes requested):
Great work! But there are a few small differences with the working version (second and third comments).

VitaliiVashkevych:
1. Removed `trim()` from search
2. Added `has-background-info-light` to selected todo
3. Changed loader logic, now works correctly

src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { getPreparedTodos } from './components/services/getPreparedTodos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setLoading] = useState(false);
+
+  const [query, setQuery] = useState('');
+  const [filterOption, setFilterOption] = useState('all');
```
BudnikOleksii, line 19: Create an enum for `filterOption`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,59 @@
...
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  query: string;
+  setQuery: (value: string) => void;
+  filterOption: string;
+  setFilterOption: (value: string) => void;
```
BudnikOleksii, line 7: And use this enum as a correct type in all places

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,59 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filterOption}
+            onChange={event => setFilterOption(event.target.value)}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
BudnikOleksii, line 27: Also, you can use the enum values with `Object.values` to create an array and then map through it to create options

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,59 @@
...
+          value={query}
+          onChange={event => setQuery(event.target.value)}
+        />
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
BudnikOleksii, line 47: - Don't use inline styles, use classes instead
- Don't disable linter

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,51 @@ export const TodoList: React.FC = () => (
...
-            </span>
-          </button>
-        </td>
-      </tr>
+        return (
+          <tr
+            data-cy="todo"
+            className={
+              selectedTodo?.id === id ? 'has-background-info-light' : ''
+            }
```
BudnikOleksii, line 39: - Use classnames library instead
- `selectedTodo?.id === id` you compare them a few times, create a variable

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,51 @@ export const TodoList: React.FC = () => (
...
+            <td className="is-vcentered">{id}</td>
+            <td className="is-vcentered">
+              {completed && (
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              )}
+            </td>
+            <td className="is-vcentered is-expanded">
+              <p className={completed ? 'has-text-success' : 'has-text-danger'}>
```
BudnikOleksii, line 51: Use classnames library, fix in all places

BudnikOleksii (changes requested):
Good work 👍 
Let's fix some moments, check the suggestions

VitaliiVashkevych:
1. Created enum `FilterOption`, used in all cases
2. Created `optionList` array, based on values from enum `FilterOption`, used it to create options.
3. Used classNames library in all cases
4. Created var `isSelectedTodo`, used in all case
5. [Comment](https://github.com/mate-academy/react_dynamic-list-of-todos/pull/2261#discussion_r1544189037) Fixed. (These lines were in the initial code, so I supposed, it should be there)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,25 +37,32 @@ export const TodoModal: React.FC = () => {
...
             </p>
 
             <p className="block" data-cy="modal-user">
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              {selectedTodo.completed ? (
+                <strong className="has-text-success">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
```
StasSokolov1, line 61: here you can use single `strong` tag and only change classname and children

StasSokolov1 (approved):
Great job, I'm approving it with only one minor comment🔥

PR #2265, opened 2024-03-28
add task solution

hattoriKanto:
[DEMO LINK](https://hattoriKanto.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,81 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { Status } from './types/Status';
 import { TodoList } from './components/TodoList';
-import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
+import { TodoFilter } from './components/TodoFilter';
 import { Loader } from './components/Loader';
 
+type FilteringProps = {
+  selectedStatus: Status;
+  query: string;
+};
+
+const getPreparedTodos = (
+  todos: Todo[],
+  { selectedStatus, query }: FilteringProps,
+) => {
+  let preparedTodos = [...todos];
+
+  switch (selectedStatus) {
+    case 'active':
+      preparedTodos = todos.filter(todo => !todo.completed);
+      break;
+    case 'completed':
+      preparedTodos = todos.filter(todo => todo.completed);
+      break;
+  }
+
+  if (query) {
+    const normalizedQuery = query.trim().toLowerCase();
+
+    preparedTodos = preparedTodos.filter(todo => {
+      const normalizedTitle = todo.title.toLowerCase();
+
+      return normalizedTitle.includes(normalizedQuery);
+    });
+  }
+
+  return preparedTodos;
+};
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo>();
+
+  const [isListLoading, setIsListLoading] = useState(true);
+  const [isModalShowing, setIsModalShowing] = useState(false);
+
+  const [query, setQuery] = useState('');
+  const [selectedStatus, setSelectedStatus] = useState<Status>('all');
+
+  let visibleTodos = getPreparedTodos(todos, { selectedStatus, query });
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsListLoading(false));
+  }, []);
+
+  const handleClick = (todo: Todo) => {
+    setSelectedTodo(todo);
+    setIsModalShowing(true);
+  };
+
+  const handleDelete = () => {
+    setIsModalShowing(false);
+  };
+
+  const handleReset = () => {
+    setQuery('');
+    visibleTodos = todos;
+  };
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


anastasia1383 (approved):
Good job!

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+  todos: Todo[],
+  { selectedStatus, query }: FilteringOption,
+) => {
+  let preparedTodos = [...todos];
+
+  switch (selectedStatus) {
+    case 'active':
+      preparedTodos = todos.filter(todo => !todo.completed);
+      break;
+    case 'completed':
```
BudnikOleksii, line 25: Create an enum for statuses and use it instead of magic values

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo>({
+    id: 0,
+    title: '',
+    completed: false,
+    userId: 0,
+  });
+
+  const [isListLoading, setIsListLoading] = useState(true);
+  const [isModalShowing, setIsModalShowing] = useState(false);
```
BudnikOleksii, line 53:
```suggestion
  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);

  const [isLoading, setIsLoading] = useState(false);
```
- Show the modal only when you have a valid todo, and you can get rid of useless state
- The loading state should be false initially. You should set it when you start loading and remove after request

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+
+  const [query, setQuery] = useState('');
+  const [selectedStatus, setSelectedStatus] = useState<Status>('all');
+
+  let visibleTodos = getPreparedTodos(todos, { selectedStatus, query });
```
BudnikOleksii, line 58:
```suggestion
  const [selectedStatus, setSelectedStatus] = useState<Status>(Status.All);

  const visibleTodos = getPreparedTodos(todos, { selectedStatus, query });
```
- Let's change your `Status` to enum
- Prefer const over let

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsListLoading(false));
+  }, []);
+
+  const handleClick = (todo: Todo) => {
```
BudnikOleksii, line 66: What click do you handle here?

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+  };
+
+  const handleDelete = () => {
+    setIsModalShowing(false);
+  };
+
+  const handleReset = () => {
+    setQuery('');
+    visibleTodos = todos;
+  };
```
BudnikOleksii, line 78:
```suggestion
  const handleReset = () => {
    setQuery('');
    setSelectedStatus(Status.All);
  };
```
You should set the status to all instead, and your function `getPreparedTodos` returns not filtered array

BudnikOleksii (changes requested):
Good work 👍 
Some moments have to be fixed, check the comments

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+
+  const [query, setQuery] = useState('');
+  const [selectedStatus, setSelectedStatus] = useState<Status>(Status.all);
+
+  const visibleTodos = getPreparedTodos(todos, { selectedStatus, query });
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(true));
```
BudnikOleksii, line 57:
```suggestion
  useEffect(() => {
    setIsLoading(true);
    
    getTodos()
      .then(setTodos)
      .finally(() => setIsLoading(false));
```
When the loading is finished you should hide the loader

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,33 @@
...
+import React from 'react';
+import { Status } from '../../types/Status';
+import classNames from 'classnames';
+
+type Props = {
+  query: string;
+  selectedStatus: Status;
+  onReset: () => void;
+  onQueryChange: (event: string) => void;
+  onStatusChange: (event: Status) => void;
```
BudnikOleksii, line 10: Why do you call arguments `event` if you pass a query and a status?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,33 @@
...
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
+        <select
+          data-cy="statusSelect"
+          value={selectedStatus}
+          onChange={event => onStatusChange(event.target.value as Status)}
+        >
+          <option value={Status.all}>All</option>
+          <option value={Status.active}>Active</option>
+          <option value={Status.completed}>Completed</option>
```
BudnikOleksii, line 30: You can create an array of enum values and map through it to create options

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -14,17 +36,25 @@ export const TodoFilter = () => (
...
       <span className="icon is-left">
         <i className="fas fa-magnifying-glass" />
       </span>
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
+      {query && (
+        <span className={classNames('icon is-right', { pointerEvents: 'all' })}>
```
BudnikOleksii, line 49: You use the classnames library incorrectly, https://www.npmjs.com/package/classnames

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,30 @@
...
+
+export const TodoModal: React.FC<Props> = ({ todo, setSelectedTodo }) => {
+  const [selectedUser, setSelectedUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    getUser(todo.userId)
+      .then(setSelectedUser)
+      .finally(() => setIsLoading(true));
+  }, [todo.userId]);
```
BudnikOleksii, line 21: The same

BudnikOleksii (changes requested):
Good progress, but a few fixes are still required

StasSokolov1 (approved):
Looks like a masterpiece🔥

PR #2267, opened 2024-03-29
add solution

Graf6612:
https://graf6612.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +6,54 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+enum FilterType {
+  Active = 'active',
+  Completed = 'completed',
+  All = 'all',
+}
+
+const filterTodos = (todos: Todo[], filterType: string, filterText: string) => {
+  let filteredTodos: Todo[] = [...todos];
+
+  if (filterType === FilterType.Active) {
+    filteredTodos = filteredTodos.filter(todo => {
+      return !todo.completed;
+    });
+  }
+
+  if (filterType === FilterType.Completed) {
+    filteredTodos = filteredTodos.filter(todo => {
+      return todo.completed;
+    });
+  }
+
+  if (filterText) {
+    filteredTodos = filteredTodos.filter(todo =>
+      todo.title.toLowerCase().includes(filterText.toLowerCase()),
+    );
+  }
+
+  return filteredTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodods] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [filterType, setFilterType] = useState<string>('All');
+  const [filterText, setFilterText] = useState('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos().then(data => {
+      setTodods(data);
+      setIsLoading(false);
+    });
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +62,34 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                filterText={filterText}
+                setFilterText={setFilterText}
+                setFilterType={setFilterType}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading ? (
+                <Loader />
+              ) : (
+                <TodoList
+                  todos={filterTodos(todos, filterType, filterText)}
+                  setSelectedTodo={setSelectedTodo}
+                  selectedTodo={selectedTodo}
+                />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal
+          selectedTodo={selectedTodo}
+          setSelectedTodo={setSelectedTodo}
+        />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,30 +1,55 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import React from 'react';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+interface TodoFilterProps {
+  setFilterType: (filterType: string) => void;
+  filterText: string;
+  setFilterText: (filterText: string) => void;
+}
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
-    </p>
-  </form>
-);
+export const TodoFilter: React.FC<TodoFilterProps> = ({
+  setFilterType,
+  filterText,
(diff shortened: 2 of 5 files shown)


hattoriKanto (approved):
Looks good for me!

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+enum FilterType {
+  Active = 'active',
+  Completed = 'completed',
+  All = 'all',
+}
+
+const filterTodos = (todos: Todo[], filterType: string, filterText: string) => {
```
BudnikOleksii, line 18:
```suggestion
const filterTodos = (todos: Todo[], filterType: FilterType, filterText: string) => {
```
`filterType` should have the enum type

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+  let filteredTodos: Todo[] = [...todos];
+
+  if (filterType === FilterType.Active) {
+    filteredTodos = filteredTodos.filter(todo => {
+      return !todo.completed;
+    });
+  }
+
+  if (filterType === FilterType.Completed) {
```
BudnikOleksii, line 27: - `filter` doesn't mutate any array so you don't need to create a copy
- Use switch case instead

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+    );
+  }
+
+  return filteredTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodods] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [filterType, setFilterType] = useState<string>('All');
```
BudnikOleksii, line 45: Use enum instead of magic values

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,55 @@
...
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+interface TodoFilterProps {
+  setFilterType: (filterType: string) => void;
```
BudnikOleksii, line 4: The same, fix it everywhere

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,55 @@
...
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={event => setFilterType(event.target.value)}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
BudnikOleksii, line 24: You can get an array of values from your enum with `Object.values` and map through it to create links

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,55 @@
...
+          className="input"
+          value={filterText}
+          placeholder="Search..."
+          onChange={event => setFilterText(event.target.value)}
+        />
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
BudnikOleksii, line 42: Avoid inline styles, use classes instead

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,74 @@
...
+            <td className="is-vcentered">{todo.id}</td>
+            {todo.completed ? (
+              <td className="is-vcentered">
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              </td>
+            ) : (
+              <td className="is-vcentered" />
+            )}
```
BudnikOleksii, line 41: `<td className="is-vcentered">` in both cases, only content is different

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,74 @@
...
+                type="button"
+                onClick={() => setSelectedTodo(todo)}
+              >
+                <span className="icon">
+                  <i
+                    className={
+                      selectedTodo && selectedTodo.id === todo.id
+                        ? 'far fa-eye-slash'
+                        : 'far fa-eye'
+                    }
```
BudnikOleksii, line 64: Use classnames library to set classes conditionally

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  setSelectedTodo,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    getUser(selectedTodo.userId).then(data => {
+      setUser(data);
+      setIsLoading(false);
+    });
```
BudnikOleksii, line 23: You should set the loading state when you start loading something and remove it in the finally block as you should do it in any case, not only when the request succeeded

BudnikOleksii (changes requested):
Good work 👍 
Left some comments to improve your solution, check them

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
+    getTodos()
+      .then((todos: Todo[]) => {
+        const filteredTodos = filterTodos(todos, filter);
+
+        setTodosFromServer(filteredTodos);
+      })
+      .finally(() => {
+        setIsTodosLoading(false);
+      });
+  }, [filter]);
```
BudnikOleksii, line 59: You don't need to load todos from the server every time the filter changes. Do it just once

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,75 @@
...
+
+export const TodoFilter: React.FC<Props> = ({ onFilterChange }) => {
+  const [status, setStatus] = useState<Status>(Status.All);
+  const [searchInput, setSearchInput] = useState('');
+
+  const handleStatusChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const newStatus = event.target.value as Status;
+
+    setStatus(newStatus);
+    onFilterChange({ status: newStatus, searchInput });
```
BudnikOleksii, line 17: If you need to change only part of the state and save other previous values use functional `setState`.
Fix it in all places

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,75 @@
...
+import { FilterCriteria, Status } from '../../types/otherTypes';
+import React from 'react';
+type Props = {
+  onFilterChange: (filterCriteria: FilterCriteria) => void;
+};
+
+export const TodoFilter: React.FC<Props> = ({ onFilterChange }) => {
+  const [status, setStatus] = useState<Status>(Status.All);
+  const [searchInput, setSearchInput] = useState('');
+
```
BudnikOleksii, line 12: You store these values in the parent component in `filter` state. Get rid of them here, you have to have one source of truth

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,85 @@
...
+                  type="button"
+                  onClick={() => onSelect(todo)}
+                >
+                  {' '}
+                  <span className="icon">
+                    {selectedTodo && todo.id === selectedTodo.id ? (
+                      <i className="far fa-eye-slash" />
+                    ) : (
+                      <i className="far fa-eye" />
+                    )}
```
BudnikOleksii, line 75: - In both cases, you have the `far` icon, only one class change
- Create a variable for `todo.id === selectedTodo.id` as you do it twice

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,85 @@
...
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+        <tbody>
+          {todos.map((todo, index) => (
+            <tr
+              key={index}
+              data-cy="todo"
```
BudnikOleksii, line 38: Create a todo component

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    setLoading(true);
+    getUser(selectedTodo.userId)
+      .then(userData => {
+        setUser(userData);
+      })
```
BudnikOleksii, line 26:
```suggestion
      .then(setUser)
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+  selectedTodo: Todo;
+  onSelect: (todo: Select) => void;
+}
+
+export const TodoModal: React.FC<TodoModalProps> = ({
+  selectedTodo,
+  onSelect,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [loading, setLoading] = useState(true);
```
BudnikOleksii, line 19: Boolean variables should start from `is` or `has`

src/types/otherTypes.ts
```diff
@@ -0,0 +1,14 @@
+import { Todo } from './Todo';
+
+export type Select = Todo | null;
```
BudnikOleksii, line 3: What select is it?

BudnikOleksii (changes requested):
.

PR #2270, opened 2024-03-29
add solution

Dmitrel:
https://Dmitrel.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,30 +1,60 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import React from 'react';
+import { SortFields } from '../../types/SortFields';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  filter: SortFields;
+  query: string;
+  setFilter: (value: SortFields) => void;
+  setQuery: (query: string) => void;
+};
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
-    </p>
-  </form>
-);
+export const TodoFilter: React.FC<Props> = ({
+  filter,
+  query,
+  setFilter,
+  setQuery,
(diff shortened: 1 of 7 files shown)


Math8585 (approved):
Good job!

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,60 @@
...
+          value={query}
+          onChange={e => setQuery(e.target.value)}
+        />
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
BudnikOleksii, line 48: - Avoid inline styles, use classes instead
- Don't disable linter

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,79 @@
...
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => setModalInfo(todo)}
+              >
+                <span className="icon">
+                  <i
+                    className={classNames(
+                      'far',
+                      modalInfo?.id === todo.id ? 'fa-eye-slash' : 'fa-eye',
```
BudnikOleksii, line 68: You compare `modalInfo?.id === todo.id` twice, create a variable

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,30 @@
...
+};
+
+export const TodoModal: React.FC<Props> = ({ modalInfo, setModalInfo }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    getUser(modalInfo.userId)
+      .then(setUser)
+      .finally(() => setIsLoading(true));
```
BudnikOleksii, line 20:
```suggestion
     setIsLoading(true);
    
    getUser(modalInfo.userId)
      .then(setUser)
      .finally(() => setIsLoading(false));
```
You should set the loading state when you start loading something to show loader and remove it after finish

src/utils/getPrepareTodos.ts
```diff
@@ -0,0 +1,27 @@
+import { SortFields } from '../types/SortFields';
+import { Todo } from '../types/Todo';
+
+export const getPreparedTodos = (
+  todoList: Todo[],
+  select: SortFields,
+  query: string,
+) => {
+  const visibleTodos = [...todoList];
```
BudnikOleksii, line 9: Filter creates a new array, you don't need to create a copy

src/utils/getPrepareTodos.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  return visibleTodos
+    .filter(todo =>
+      todo.title.toLowerCase().includes(query.trim().toLowerCase()),
+    )
+    .filter(({ completed }) => {
```
BudnikOleksii, line 15: - If you don't have any query you don't need to filter
- The same if the filter type is `all`

BudnikOleksii (changes requested):
Almost done, let's improve your solution slightly

PR #2276, opened 2024-04-01
Solution

okarnaukh:
[DEMO LINK](https://okarnaukh.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,55 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { StatusFilterValue, Todo } from './types/Todo';
+import { getTodos } from './api';
+import { TodoModal } from './components/TodoModal';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [statusFilter, setStatusFilter] = useState<StatusFilterValue>('all');
+  const [query, setQuery] = useState('');
+
+  const handleTogglingTodo = (todo: Todo | null) => {
+    setSelectedTodo(todo);
+  };
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(setTodos)
+      .finally(() => setLoading(false));
+  }, []);
+
+  const getPreparedTodos = () => {
+    let result = [...todos];
+
+    if (statusFilter === 'completed') {
+      result = result.filter(todo => todo.completed);
+    }
+
+    if (statusFilter === 'active') {
+      result = result.filter(todo => !todo.completed);
+    }
+
+    if (query) {
+      result = result.filter(todo =>
+        todo.title.toLowerCase().includes(query.toLowerCase()),
+      );
+    }
+
+    return result;
+  };
+
+  const preparedTodos = getPreparedTodos();
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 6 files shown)


Graf6612 (approved):
Great Job)

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { StatusFilterValue, Todo } from './types/Todo';
+import { getTodos } from './api';
+import { TodoModal } from './components/TodoModal';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
```
BudnikOleksii, line 15: Boolean variables should start from `is` or `has`

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [statusFilter, setStatusFilter] = useState<StatusFilterValue>('all');
```
BudnikOleksii, line 17: Let's change the `StatusFilterValue` into enum and use it instead

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+  const getPreparedTodos = () => {
+    let result = [...todos];
+
+    if (statusFilter === 'completed') {
+      result = result.filter(todo => todo.completed);
+    }
+
+    if (statusFilter === 'active') {
+      result = result.filter(todo => !todo.completed);
+    }
```
BudnikOleksii, line 40: - Move this function outside the component
- You don't need to create a copy as `filter` doesn't mutate an array
- Use switch case and enum instead of `if` and magic values

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,26 @@
...
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
+          data-cy="statusSelect"
+          value={statusFilter}
+          onChange={e => onStatusSelected(e.target.value as StatusFilterValue)}
+        >
           <option value="all">All</option>
           <option value="active">Active</option>
           <option value="completed">Completed</option>
```
BudnikOleksii, line 26: When you have an enum you can get an array of values and map through it to create links

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,15 +34,23 @@ export const TodoFilter = () => (
...
       <span className="icon is-left">
         <i className="fas fa-magnifying-glass" />
       </span>
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
+      {query && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
BudnikOleksii, line 45: Avoid inline styles, use classes instead

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,18 @@
 import React from 'react';
+import { Todo } from '../../types/Todo';
+import classNames from 'classnames';
 
-export const TodoList: React.FC = () => (
+type Props = {
+  todos: Todo[];
+  onTodoSelected: (_: Todo) => void;
```
BudnikOleksii, line 7: `_` is used when you need to skip the argument and use the next one. In this case, it should be a proper name for args

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,52 @@ export const TodoList: React.FC = () => (
...
+            key={id}
+          >
+            <td className="is-vcentered">{id}</td>
+            {(completed && (
+              <td className="is-vcentered">
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              </td>
+            )) || <td className="is-vcentered" />}
```
BudnikOleksii, line 49: `<td className="is-vcentered">` in both cases, only content is different

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,52 @@ export const TodoList: React.FC = () => (
...
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => onTodoSelected(todo)}
+              >
+                <span className="icon">
+                  <i
+                    className={classNames('far', {
+                      'fa-eye-slash': selectedTodo?.id === todo.id,
+                      'fa-eye': selectedTodo?.id !== todo.id,
```
BudnikOleksii, line 68: You compare `selectedTodo?.id !== todo.id` twice, create a variable

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,26 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+import { getUser } from '../../api';
+
+interface Props {
+  todo: Todo;
+  onClose: (_: Todo | null) => void;
```
BudnikOleksii, line 9: The same

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,26 @@
...
+    getUser(todo.userId).then(setUser);
+  }, [todo.userId]);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!user ? (
```
BudnikOleksii, line 23: If something happens and you don't get a user you will show the infinite loader. Create a loading state to show loader

BudnikOleksii (changes requested):
Good start 👍 
Left some suggestions, check them

BudnikOleksii (approved):
Looks great 🚀

PR #2283, opened 2024-04-02
couple tests fall

yurovych:
https://yurovych.github.io/react_dynamic-list-of-todos/


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,121 @@
...
+  const [initialCompleteOriginal, setInitialCompleteOriginale] =
+    useState(todosInitial);
+
+  const [initialInput, setInitialInput] = useState(todosInitial);
+  const [initialInputOriginal, setInitialInputOriginale] =
+    useState(todosInitial);
+
+  const [inputValue, setInputValue] = useState('');
+
+  useEffect(() => {
```
etojeDenys, line 23: you don't need all these states here, just create two state(one for filterValue, and one for the query), move them to the App.tsx, and pass here as props the values of those states and setState functions

so this component will just change the state, you will filter the todos in the App

src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos, getUser } from './api';
+import { ShowModalContext } from './components/context/stateContext';
+import { User } from './types/User';
 
 export const App: React.FC = () => {
+  const [todosInitial, setTodosInitial] = useState<Todo[]>([]);
+  const [todos, setTodos] = useState<Todo[]>(todosInitial);
```
etojeDenys, line 17: try to simplify it using only one state

src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+  const [user, setUser] = useState<User>();
+  const [userId, setUserId] = useState(1);
```
etojeDenys, line 19: you don't need that here, move the Modal and try to simplify using only one state

src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+
```
etojeDenys, line 20: add here the 
```
const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null)
```
and pass this state into TodoList

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,78 @@
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+  const handleEyeButtonClick = (id: number, userId: number) => {
+    setisModalShowed(true);
+    setCurrentId(id);
+    setCurrentUserId(userId);
+  };
```
etojeDenys, line 19: just save the whole todo to selectedTodo state here

src/App.tsx
```diff
@@ -17,18 +37,22 @@ export const App: React.FC = () => {
...
+              {todos.length === 0 && <Loader />}
+
+              <TodoList todos={todos} setCurrentUserId={setUserId} />
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {isModalShowed === true && <TodoModal user={user} todos={todos} />}
```
etojeDenys, line 55:
```suggestion
      {selectedTodo && <TodoModal todo={selectedTodo} onClose={() => setSelectedTodo(null)} />}
```
take the userId from the todo and fetch the user inside the TodoModal

src/components/context/stateContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export const ShowModalContext = React.createContext({
+  isModalShowed: false,
+  setisModalShowed: (isModalShowed: boolean) => {
+    // eslint-disable-next-line no-console
+    console.log(isModalShowed);
+  },
+});
+
+type PropsShowModal = {
```
etojeDenys, line 40: it's redundant to use the context here, 'cause you don't have the props drilling problem in your code

etojeDenys (changes requested):
Якщо раптом щось із моїх коментарів тобі незрозуміло, то, будь ласка, не чекай і звертайся у чат з менторами або приходь на Q&A сесію. Обовʼязково з усім розберемось)

ghost (approved):
Good job!

PR #2292, opened 2024-04-08
add task solution

Klimukalina:
[DEMO LINK](https://Klimukalina.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/utils/getPreparedTodos.ts
+++ b/src/utils/getPreparedTodos.ts
@@ -1,0 +1,27 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
+export const getPreparedTodos = (
+  todoList: Todo[],
+  select: Status,
+  query: string,
+) => {
+  const visibleTodos = [...todoList];
+
+  return visibleTodos
+    .filter(todo =>
+      todo.title.toLowerCase().includes(query.trim().toLowerCase()),
+    )
+    .filter(({ completed }) => {
+      switch (select) {
+        case Status.Active:
+          return !completed;
+
+        case Status.Completed:
+          return completed;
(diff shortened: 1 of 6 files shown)


etojeDenys (changes requested):
looks broken, try to fix it firstly and then re-request the review
image

feel free to ask for some help in the chat

src/utils/getPreparedTodos.ts
```diff
@@ -0,0 +1,27 @@
...
+          return !completed;
+
+        case Status.Completed:
+          return completed;
+
+        default:
+          return visibleTodos;
+      }
+    });
+};
```
IvanFesenko, line 27: No need to filter in case Status.All or query is empty - it may cause performance issues. Also pay attention, why you return all todos in default case?

IvanFesenko (changes requested):
Good job. Lets improve your code - check my comment

src/utils/getPreparedTodos.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function getPreparedTodos(
+  todoItems: Todo[],
+  filterStatus: Status,
+  query: string,
+): Todo[] {
+  let filtered = todoItems.filter(todo => {
+    switch (filterStatus) {
+      case Status.All:
+        return true;
```
vadiimvooo, line 12: As it was mentioned before you don't need to filter todos if filterStatus is All.

vadiimvooo (changes requested):
Please fix previous comment. Feel free to ask for help in fe_chat if you need so.

vadiimvooo (approved):
Well done!)

PR #2299, opened 2024-04-11
Implement fetching filtered todos functionality

olafchuszno:
[DEMO LINK](https://olafchuszno.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,39 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
-import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,39 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
```
yevhenii-pyl, line 4: 👍

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,34 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { User } from '../../types/User';
+import { getUser } from '../../api';
+import { Todo } from '../../types/Todo';
+
+export const TodoModal: React.FC<{
+  todo: Todo;
```
yevhenii-pyl, line 8: Here it is readable, but having 3+ props may become a bit messy. 

Better approach for such cases will be to simply define a Props type, like:

```
type Props = {
    todo: Todo,
    ...
}
```

and use Props at the component definition.

yevhenii-pyl (approved):
Solid work! 🥇 
Your code is cleaner and better with each PR.

PR #2301, opened 2024-04-13
cope

true-binar-White-Rabbit:
[DEMO LINK](https://true-binar-White-Rabbit.github.io/react_dynamic-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,122 @@
...
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+  React.useEffect(() => {
```
vadiimvooo, line 19: It's better to use destructuring for hooks and other react elements to make it more readable. The same for other useEffect usages.

vadiimvooo (changes requested):
Almost done. Please fix not passed tests. Feel free to ask for help in fe_chat if you need so.

VitaliyBondarenko1982 (approved):
Looks good

PR #2304, opened 2024-04-15
done

TideTwo:
[DEMO LINK](https://tidetwo.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,54 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [statusFilter, setStatusFilter] = useState<FilterEnum>('all');
+  const [titleFilter, setTitleFilter] = useState<string>('');
+
+  useEffect(() => {
+    getTodos().then(todosFromServer => {
+      setTodos(todosFromServer);
+      setIsLoading(false);
+    });
```
choeqq, line 26: and what if then fails?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,29 @@
...
+  setSelectedTodo: React.Dispatch<React.SetStateAction<Todo | null>>;
+}> = ({ selectedTodo: { completed, id, title, userId }, setSelectedTodo }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+
+  useEffect(() => {
+    getUser(userId).then(userFromServer => {
+      setUser(userFromServer);
+      setIsLoading(false);
+    });
```
choeqq, line 19: and what if then fails?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,29 @@
...
+  }, []);
```
choeqq, line 20: don't you have a missing dependency ?

choeqq (changes requested):

src/components/TodoModal/TodoModal.tsx
TideTwo: What should it be dependent on? We only need the user once, since that component gets unmounted and remounted for the next user, if that's what you mean?

choeqq (approved):

PR #2315, opened 2024-04-25
Solution

weswojciech:
https://weswojciech.github.io/react_dynamic-list-of-todos/


src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
+    }
+
+    return null;
+  }, [selectedTodoId, filteredTodos]);
+
+  const handleSelectTodo = (id: number | null) => setSelectedTodoId(id);
+  const handleCloseModal = () => setSelectedTodoId(null);
+  const handleStatusFilterChange = (status: StatusFilter) =>
+    setStatusFilter(status);
+  const handleSearchQueryChange = (newQuery: string) => setQuery(newQuery);
```
choeqq, line 68: please add empty lines between these handlers

src/App.tsx
```diff
@@ -17,18 +75,33 @@ export const App: React.FC = () => {
...
+                <TodoList
+                  todos={filteredTodos}
+                  selectedTodoId={selectedTodoId}
+                  onSelect={handleSelectTodo}
+                />
+              ) : error ? (
+                <p>Something went wrong!</p>
+              ) : (
+                <Loader />
+              )}
```
choeqq, line 96: please don't nest ternaries

choeqq (changes requested):

src/App.tsx
weswojciech: I am not sure how to format it, look into the version that I've just commited

choeqq (changes requested):
@weswojciech you can use filteredTodos and error independently, this should solve nested ternaries

PR #2319, opened 2024-04-26
add task solution

AgnieszkaWw:
[DEMO LINK](https://AgnieszkaWw.github.io/react_dynamic-list-of-todos/) and add it to the PR description.


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,31 @@
...
+
+export const TodoModal: React.FC<Props> = ({ todo, onClose, userId }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getUser(userId)
+      .then(setUser)
+      .finally(() => setIsLoading(false));
```
yevhenii-pyl, line 21: 👍

src/components/TodoList/TodoList.tsx
```diff
@@ -14,87 +26,52 @@ export const TodoList: React.FC = () => (
...
+            </p>
+          </td>
+          <td className="has-text-right is-vcentered">
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => onSelect(todo)}
+            >
+              {todo && (
```
yevhenii-pyl, line 62: Not sure why do you need this `todo &&` check. 
You are already mapping through todos array and probably a todo, as it's item, should be there :)

PR #2323, opened 2024-04-27
solution

antonshtef:
[DEMO LINK](https://antonshtef.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoModal/TodoModal.tsx
+++ b/src/components/TodoModal/TodoModal.tsx
@@ -16,14 +16,10 @@ export const TodoModal: React.FC<Props> = ({
   const [user, setUser] = useState<User | null>(null);
   const [loading, setLoading] = useState(true);
 
-  getUser(selectedTodo.userId).then(setUser);
-
   useEffect(() => {
-    setTimeout(() => {
-      if (selectedTodo) {
-        setLoading(false);
-      }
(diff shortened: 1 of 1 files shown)


SanyaBratashchuk (changes requested):
Demo link is not working Try to run `npm run deploy` once again 
image

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,36 @@
...
+  const [loading, setLoading] = useState(true);
+
+  getUser(selectedTodo.userId).then(setUser);
+
+  useEffect(() => {
+    setTimeout(() => {
+      if (selectedTodo) {
+        setLoading(false);
+      }
+    }, 1000);
```
etojeDenys, line 26: looks strange
1. you need to fetch the user in the useEffect
2. setTimeout is redundant here
```suggestion
  

  useEffect(() => {
    getUser(selectedTodo.userId)
       .then(setUser)
       .finally(() => setLoading(false))
```

etojeDenys (approved):

etojeDenys (changes requested):
better to fix that

etojeDenys (approved):
good job

PR #2324, opened 2024-04-28
Develop

ArtemZakhar:


DanilWeda (changes requested):
Great job. 
I left some comments. 
Also, you can split code (more) into components if you wont and have time for it. 
Many thanks)

src/App.tsx
```diff
@@ -1,14 +1,57 @@
...
+import { sortTodos } from './helpers/sortTodos';
 
 export const App: React.FC = () => {
+  const initialTodo = {
+    id: -1,
+    title: '',
+    completed: false,
+    userId: -1,
+  };
+  const initialUser = { id: -1, name: '', email: '', phone: '' };
```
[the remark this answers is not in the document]
ArtemZakhar: Where should I store this information? I mean what the name of the folder should be?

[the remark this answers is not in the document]
DanilWeda: You can use initialData, defaultDate, etc.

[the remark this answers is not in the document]
ArtemZakhar: Thank you) Give me a second for change)

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [loadingModal, setLoadingModal] = useState(false);
+  const [modalShow, setModalShow] = useState(false);
+  const [activeTodo, setActiveTodo] = useState(initialTodo);
```
undeletable, line 20: It makes sense to indicate type here too (`useState<Todo>`), because it's not obvious which type is `initialTodo` of.

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+  const [userTodo, setUserTodo] = useState(initialUser);
```
undeletable, line 21: Better choice for name would be `todoUser`, `todoUsername`, `todoOwner` or `todoAssignee`, depending on the meaning of user entity in the app. `userTodo` sound more like user's TODO.
Also it makes sense to indicate `User` type here.

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+  const [sortStatus, setSortStatus] = useState('all');
+  const [querry, setQuerry] = useState('');
```
undeletable, line 23: Typo: query, not quer**r**y

src/App.tsx
ArtemZakhar: Thank you)

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,80 @@
...
-        <th>
-          <span className="icon">
-            <i className="fas fa-check" />
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+type Props = {
```
undeletable, line 4: It's better to give this interface a more clear name, e.g. `TodoListProps`

ArtemZakhar: Got it) Thanks)

src/helpers/getSortedTodos.ts
```diff
@@ -0,0 +1,39 @@
+/* pretie-disable */
```
undeletable, line 1: It should be `/* prettier-ignore */` according to https://prettier.io/docs/en/ignore.html. But it's better not to disable code style checks if there's no strong reason for that.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,43 +1,75 @@
 import React from 'react';
 import { Loader } from '../Loader';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
 
-export const TodoModal: React.FC = () => {
+type Props = {
```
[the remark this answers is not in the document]
ArtemZakhar: Will match naming more thoughtfully in the future) thank you)

src/App.tsx
```diff
@@ -1,14 +1,50 @@
...
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [loadingModal, setLoadingModal] = useState(false);
+  const [modalShow, setModalShow] = useState(false);
+  const [activeTodo, setActiveTodo] = useState(initialTodo);
+  const [userTodo, setUserTodo] = useState(initialUser);
+  const [sortStatus, setSortStatus] = useState('all');
```
undeletable, line 22: I'd recommend to rename this to smth like `filterValue`, because it's about filtering of TODOs, not sorting them.

undeletable (approved):
That's a good work! I just have some recommendations mostly on naming of code parts - take them into account during your further work.

src/helpers/getSortedTodos.ts
ArtemZakhar: Got it) I did not have the possibility to deploy the project because of this error. So I turned it off before deploy)

src/components/TodoModal/TodoModal.tsx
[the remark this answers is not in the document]
ArtemZakhar: image

PR #2326, opened 2024-04-29
solution

etuzimek:
[DEMO LINK](https://etuzimek.github.io/react_dynamic-list-of-todos/)


choeqq (changes requested):
@etuzimek please make sure to deploy your task solution

etuzimek:
> @etuzimek please make sure to deploy your task solution

Done

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  const [userLoading, setUserLoading] = useState(false);
+
+  useEffect(() => {
+    if (activeTodo.id) {
+      setUserLoading(true);
+    }
+
+    getUser(activeTodo.userId)
+      .then(setUser)
+      .finally(() => setUserLoading(false));
```
choeqq, line 23: and what if the request fails? 
please make sure to handle error case

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  }, [activeTodo]);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {userLoading || !user ? (
```
choeqq, line 30: I think I would reverse the conditions, we should check if user is present and user is not loading, it will be more clear that way. As a developer I want to check component which should be displayed on success instead of loader

choeqq (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+          return todo;
+      }
+    })
+    .filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()));
+
+  useEffect(() => {
+    setTodoLoading(true);
+    getTodos()
+      .then(setTodos)
+      .finally(() => setTodoLoading(false));
```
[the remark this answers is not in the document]
etuzimek: error added

src/components/TodoModal/TodoModal.tsx
etuzimek: error added

src/components/TodoModal/TodoModal.tsx
etuzimek: conditions reversed

PR #2333, opened 2024-04-30
Develop

GIroSTarK:
[DEMO LINK](https://girostark.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,11 +9,14 @@ import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 import { getTodos } from './api';
 import { Todo } from './types/Todo';
+import { FilterOptions } from './types/FilterOptions';
 
 export const App: React.FC = () => {
   const [isLoading, setIsLoading] = useState(false);
   const [todos, setTodos] = useState<Todo[]>([]);
   const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [query, setQuery] = useState('');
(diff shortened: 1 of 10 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +32,55 @@ export const TodoList: React.FC = () => (
...
-              <i className="far fa-eye-slash" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos
+        .filter(todo => {
+          switch (filterOption) {
+            case 'all':
+              return true;
```
etojeDenys, line 39: move this filtering logic to the App.tsx and after that remove query and filterOption prop from this component

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,49 @@
...
+          }
+        }),
+      )
+      .then(tasks =>
+        tasks.filter(task =>
+          task.title.toLowerCase().includes(query.toLowerCase()),
+        ),
+      )
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
```
etojeDenys, line 44: looks too complex, you will make every time on query or filterOption change new request to BE

just filter the array like that in the component
`const filteredTodos = getFilteredTodos(todos, query, filterOption)`
don't forget to create that function

etojeDenys (changes requested):

l4st1m0za (approved):
Good job!

PR #2336, opened 2024-05-05
Developbr

Allanpt:
[DEMO LINK](https://Allanpt.github.io/react_dynamic-list-of-todos/)


jv-aquino (approved):
Perfect job Allan, you are doing a really good job in React!!! The only thing i'd maybe suggest is to put more code into functions as a convetion to make code cleaner / tighter, but it's not something impacting anything in your code

PR #2340, opened 2024-05-07
add_solution

metrakonon:
Have a nice day! :tada: 
[DEMO](https://metrakonon.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,16 +1,31 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
-import { Loader } from './components/Loader';
+import { TodoContext } from './Contexts/TodoContext';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [loader, setLoader] = useState(true);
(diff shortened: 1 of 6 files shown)


yanamomot (approved):
Interesting solution! Approved!

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,118 @@
...
+    ) {
+      return false;
+    }
+
+    return true;
+  });
+
+  {
+    /* eslint-disable react-hooks/exhaustive-deps */
+  }
```
alexander-ignatow, line 38: most often this means you are doing something wrong, as this lint rule is here for the reason

src/App.tsx
```diff
@@ -1,16 +1,30 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
-import { Loader } from './components/Loader';
+import { TodoContext } from './Contexts/TodoContext';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [filterBy, setFilterBy] = useState('all');
+  const [searchQuery, setSearchQuery] = useState('');
+  const [currentTodo, setCurrentTodo] = useState<Todo | null>(null);
```
alexander-ignatow, line 15: Not critical, but sometimes it is better to 
1. Encapsulate these states in the context
2. Delegate loading/filtering TODOs to the context too

alexander-ignatow (approved):
Good usage of the context 👍 
BTW - you could go further and delegate all TODO operations and manipulations to the context and make it the data layer of your app (see my comments for the future)

PR #2345, opened 2024-05-11
Solution

ivankovbohdan:
[DEMO LINK](https://ivankovbohdan.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,84 @@
...
+//       case FILTER_FIELD_COMPLETED:
+//         return readyTodos.filter(todo => todo.completed);
+
+//       default:
+//         return readyTodos;
+//     }
+//   }
+
+//   return readyTodos;
+// };
```
vadiimvooo, line 44:
```suggestion
```

src/tools/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const FILTER_FIELD_ALL = 'all';
+export const FILTER_FIELD_ACTIVE = 'active';
+export const FILTER_FIELD_COMPLETED = 'completed';
```
vadiimvooo, line 3: It's better to use enum here.

src/App.tsx
```diff
@@ -1,14 +1,84 @@
...
+
+          default:
+            return readyTodos;
+        }
+      })
+      .then(setTodos)
+      .finally(() => setLoading(false));
+  }, [title, filterField]);
+
+  // const readyTodos = prepareTodos(todos, title, filterField);
```
vadiimvooo, line 80:
```suggestion
```

src/App.tsx
```diff
@@ -1,14 +1,84 @@
...
+
+          case FILTER_FIELD_ACTIVE:
+            return readyTodos.filter(todo => !todo.completed);
+
+          case FILTER_FIELD_COMPLETED:
+            return readyTodos.filter(todo => todo.completed);
+
+          default:
+            return readyTodos;
+        }
```
vadiimvooo, line 74: Both filters can be merged into one and it would be better if you make helper like prepareTodos which will be defined in separate file and be called here.

vadiimvooo:
```suggestion
      .then(readyTodos => {
        return readyTodos.filter(todo => {
          const titleCondition = todo.title
            .toLowerCase()
            .includes(title.toLowerCase());

          switch (filterType) {
            case FilterType.All:
              return firstCondition;

            case  FilterType.Active:
              return firstCondition && !todo.completed;

            case FilterType.Completed:
              return firstCondition && todo.completed;

            default:
              return true;
          }
        });
      })
```

vadiimvooo: This is the example of such function but as I mentioned before you have to create helper function for it and just call it when you receive the data.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,31 @@
...
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
+        <select
+          data-cy="statusSelect"
+          onChange={event => setFilterField(event.target.value)}
+        >
+          <option value={FILTER_FIELD_ALL}>All</option>
+          <option value={FILTER_FIELD_ACTIVE}>Active</option>
+          <option value={FILTER_FIELD_COMPLETED}>Completed</option>
```
vadiimvooo, line 28: By using enum you can map this fields.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,25 +36,34 @@ export const TodoModal: React.FC = () => {
...
+              {selectedTodo.completed ? (
+                <strong className="has-text-danger">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
 
               {' by '}
 
-              <a href="mailto:Sincere@april.biz">Leanne Graham</a>
+              <a href={`mailto:${user?.email}`}>{user?.name}</a>
```
vadiimvooo, line 66: it's better to use conditional operator here and render link only if user exists

vadiimvooo (changes requested):
.

src/tools/prepareTodos.tsx
```diff
@@ -0,0 +1,30 @@
...
+      case FilterField.COMPLETED:
+        return filteredTodos.filter(todo => todo.completed);
+
+      default:
+        return filteredTodos;
+    }
+  }
+
+  return filteredTodos;
+};
```
vadiimvooo, line 30:
```suggestion
export const prepareTodos = (
  readyTodos: Todo[],
  title: string,
  filterField: FilterField,
) => {
  return readyTodos.filter((todo: Todo) => {
     const titleCondition = todo.title
            .toLowerCase()
            .includes(title.toLowerCase());
            
     switch (filterField) {
       case FilterField.All:
         return firstCondition;
         
       case  FilterField.Active:
         return firstCondition && !todo.completed;
         
       case FilterField.Completed:
         return firstCondition && todo.completed;
         
       default:
         return true;
  });
};
```

vadiimvooo: As I mentioned in previous review you can use 1 filter for getting the same result. Please fix it.

vadiimvooo (changes requested):
Almost done, just last fix.

SanyaBratashchuk (approved):

PR #2356, opened 2024-05-15
Solution

MarchAlbion:
https://MarchAlbion.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


GrigoryyKach (approved):
God job!

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+import { InitialStates, Todo } from './types/Todo';
+import { getTodos, getUser } from './api';
+import { User } from './types/User';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [toggleModal, setToggleModal] = useState<boolean>(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo>();
+  const [user, setUser] = useState<User>();
+  const [filterTodos, setFilterTodos] = useState<InitialStates>({
```
alexander-ignatow, line 19: `filterTodos` sounds like the method name

Please rename (`todosFilter`, `filter`, etc)

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+      setSelectedTodo(seleceted);
+    }
+  };
+
+  useEffect(() => {
+    getTodos().then(response => filterResponse(response));
+
+    if (selectedTodo) {
+      getUser(selectedTodo.userId).then(response => setUser(response));
+    }
```
alexander-ignatow, line 69: Don't combine loading todos and the user as they are not related

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+        setTodos(filteredTodos);
+      } else if (filterTodos.select === 'active') {
+        filteredTodos = filteredTodos.filter(todo => todo.completed === false);
+        setTodos(filteredTodos);
+      } else {
+        setTodos(filteredTodos);
+      }
+    },
+    [filterTodos.query, filterTodos.select],
+  );
```
alexander-ignatow, line 46: It makes sense to move this method away from the component

IMO the best would be creating sort of the hook `useTodos` by combining this and the effect hook, you have below

src/types/Todo.ts
```diff
@@ -4,3 +4,8 @@ export interface Todo {
   completed: boolean;
   userId: number;
 }
+
+export type InitialStates = {
```
alexander-ignatow, line 8: This is a not ok name, especially for the type

Please rename it (Filter, TodoFilter, TodoFilterOptions, etc)

alexander-ignatow (changes requested):
Overal - looks good, but the code needs some improvements.

Also, please fix this (when no Todos found, loader remains forever)
image

src/App.tsx
```diff
@@ -79,12 +44,12 @@ export const App: React.FC = () => {
...
               <TodoFilter
                 handleFilterTodos={handleFilterTodos}
-                filterTodos={filterTodos}
+                filterTodos={todosFilter}
               />
             </div>
 
             <div className="block">
-              {todos.length > 0 ? (
+              {isLoading ? (
```
undeletable, line 52: The naming is incorrect here.
I suggest:
- either to rename this constant to `isLoaded` (here and in the hook)
- or to switch to opposite values for it in the hook (set true instead of false and vice versa) and rewrite conditional here (`isLoading ? <Loader /> : <TodoList ... />`)

undeletable (changes requested):
The solution is fine. Just a single naming-related thing has to be improved. Please check the comment.

alexander-ignatow (approved):
👍

PR #2357, opened 2024-05-15
solution

GrigoryyKach:
[DEMO LINK](https://GrigoryyKach.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,69 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import { getTodos, getUser } from './api';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [selectedTodoId, setSelectedTodoId] = useState<number | null>(null);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filteredStatus, setFilteredStatus] = useState('all');
+  const [searchQuery, setSearchQuery] = useState('');
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    const loadTodos = async () => {
+      const fetchedTodos = await getTodos();
+
+      setTodos(fetchedTodos);
+      setIsLoading(false);
+    };
+
+    loadTodos();
+  }, []);
+
+  const handleShowTodo = async (todo: Todo) => {
+    setIsLoading(true);
+    setSelectedTodoId(todo.id);
+    setSelectedTodo(todo);
+    const fetchedUser = await getUser(todo.userId);
+
+    setUser(fetchedUser);
+    setIsLoading(false);
+  };
+
+  const handleHideTodo = () => {
+    setSelectedTodoId(null);
+    setSelectedTodo(null);
+    setUser(null);
+  };
+
+  const filteredTodos = todos
+    .filter(todo => {
+      if (filteredStatus === 'completed') {
+        return todo.completed;
+      }
+
+      if (filteredStatus === 'active') {
+        return !todo.completed;
+      }
+
+      return true;
+    })
+    .filter(todo =>
+      todo.title.toLowerCase().includes(searchQuery.toLowerCase().trimStart()),
+    );
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
```
sevenlays, line 17: If you click todo you can see two loaders.
twoloaders
There should be isLoading state when todos is loading and state into Modal for users is loading.

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+  const [selectedTodoId, setSelectedTodoId] = useState<number | null>(null);
```
sevenlays, line 18: May be can just get only to. No need todoId

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [filteredStatus, setFilteredStatus] = useState('all');
+  const [searchQuery, setSearchQuery] = useState('');
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    const loadTodos = async () => {
```
sevenlays, line 25: Here can be only getTodos().
no need async

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+      const fetchedTodos = await getTodos();
+
+      setTodos(fetchedTodos);
+      setIsLoading(false);
+    };
+
+    loadTodos();
+  }, []);
+
+  const handleShowTodo = async (todo: Todo) => {
```
sevenlays, line 35: Same here.

You only need async if you have to make an asynchronous request, for example, to get both todos and users in one request. This means you need to wait for todos (await getTodos()) to get everything.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,43 +1,59 @@
...
-          </header>
+export const TodoModal: React.FC<Props> = ({ todo, user, onClose }) => (
+  <div className="modal is-active" data-cy="modal">
+    <div className="modal-background" />
 
-          <div className="modal-card-body">
-            <p className="block" data-cy="modal-title">
-              quis ut nam facilis et officia qui
-            </p>
+    {user ? (
```
sevenlays, line 17: No reason for that because you should show loader when user is loading.

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+    setIsLoading(false);
+  };
+
+  const handleHideTodo = () => {
+    setSelectedTodoId(null);
+    setSelectedTodo(null);
+    setUser(null);
+  };
+
+  const filteredTodos = todos
```
sevenlays, line 51: Create separete file for getFilteredTodos func

sevenlays (changes requested):

src/utils/getFilteredTodos.ts
```diff
@@ -0,0 +1,23 @@
...
+
+      if (filteredStatus === 'active') {
+        return !todo.completed;
+      }
+
+      return true;
+    })
+    .filter(todo =>
+      todo.title.toLowerCase().includes(searchQuery.toLowerCase().trimStart()),
+    );
```
alexander-ignatow, line 22: Not critical here, but this can be done with a single `filter` call

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,25 +24,34 @@ export const TodoModal: React.FC = () => {
...
           </header>
 
           <div className="modal-card-body">
             <p className="block" data-cy="modal-title">
-              quis ut nam facilis et officia qui
+              {todo.title}
             </p>
 
             <p className="block" data-cy="modal-user">
               {/* <strong className="has-text-success">Done</strong> */}
```
alexander-ignatow, line 45: Please, don't leave the commented out code

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,19 @@
 import React from 'react';
+import { Todo } from '../../types/Todo';
 
-export const TodoList: React.FC = () => (
+type Props = {
+  todos: Todo[];
+  selectedTodoId: number | null;
+  onShowTodo: (todo: Todo) => void;
+  onHideTodo: () => void;
```
alexander-ignatow, line 8: please combine these two into one method

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +29,48 @@ export const TodoList: React.FC = () => (
...
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => onShowTodo(todo)}
+              >
+                <span className="icon">
+                  <i className={'far fa-eye'} />
+                </span>
+              </button>
+            )}
```
alexander-ignatow, line 70: This looks too bulky

```suggestion
  <button
                data-cy="selectButton"
                className="button"
                type="button"
                onClick=**please use single callback here**
              >
                <span className="icon">
                  <i className={and calculate the classes here} />
                </span>
              </button>
```

src/App.tsx
```diff
@@ -1,14 +1,52 @@
...
+    setSelectedTodo(todo);
+    getUser(todo.userId).then(res => setUser(res));
+
+    setIsLoading(false);
+  };
+
+  const handleHideTodo = () => {
+    setSelectedTodo(null);
+    setUser(null);
+  };
```
alexander-ignatow, line 46: You can use single function - something like this:


```suggestion
  const toggleTodo = (todo?: Todo) => {
  if (todo)
    {setIsLoading(true);
    setSelectedTodo(todo);
    getUser(todo.userId).then(res => setUser(res));

    setIsLoading(false);
    } else  {
    setSelectedTodo(null);
    setUser(null);
  };
  }
```

alexander-ignatow: JFYI `?` in arguments stands for the optional argument

alexander-ignatow (approved):
Good progress on the improving 👍 

However there are some more issues to fix

PR #2358, opened 2024-05-15
Develop

sevenlays:
[DEMO LINK](https://sevenlays.github.io/react_dynamic-list-of-todos/)
With Cypress everything is ok)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,38 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+
+import { Todo } from './types/Todo';
+import { FilterType } from './types/FilterType';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { getTodos } from './api';
+
+import { getFilteredTodos } from './utils/getFilteredTodos';
+
+type CurrentTodo = Todo | null;
+
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<FilterType>(FilterType.All);
+  const [query, setQuery] = useState('');
+  const [todo, setTodo] = useState<CurrentTodo>(null);
+
+  const filteredTodos = getFilteredTodos(todos, filterBy, query);
+
+  useEffect(() => {
+    getTodos()
+      .then(data => setTodos(data))
+      .finally(() => setIsLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,14 +1,38 @@
...
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<FilterType>(FilterType.All);
+  const [query, setQuery] = useState('');
+  const [todo, setTodo] = useState<CurrentTodo>(null);
+
+  const filteredTodos = getFilteredTodos(todos, filterBy, query);
+
+  useEffect(() => {
```
iAlfreX, line 30: Think about using custom hook for this

src/App.tsx
```diff
@@ -1,14 +1,38 @@
...
+type CurrentTodo = Todo | null;
+
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filterBy, setFilterBy] = useState<FilterType>(FilterType.All);
+  const [query, setQuery] = useState('');
+  const [todo, setTodo] = useState<CurrentTodo>(null);
+
+  const filteredTodos = getFilteredTodos(todos, filterBy, query);
```
iAlfreX, line 28: I would use for the TodoList component React.memo and useMemo for this

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,32 @@
...
+type Props = {
+  todo: Todo;
+  setTodo: (todo: Todo | null) => void;
+};
+
+export const TodoModal: React.FC<Props> = ({ todo, setTodo }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
```
iAlfreX, line 19: the same, I think solution will be better using custom hooks for this

src/App.tsx
alexander-ignatow: For the current case it is ok to keep all as it is, but if you had to reload Todos from server on each filter change, it would make total sense to combine effect hook + Todos filtering in the custom hook (like `useTodos`)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,60 @@
...
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
+  setFilterBy: Dispatch<SetStateAction<FilterType>>;
```
alexander-ignatow, line 5: Not critical, but I would use 

```suggestion
  setFilterBy: (newFilterType: FilterType) => void;
```

alexander-ignatow: It is because in fact you don't need everything from React.Dispatch

alexander-ignatow (approved):
Good job< just see some comments for the future

PR #2359, opened 2024-05-15
solution

AYupko:
[DEMO LINK](https://AYupko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,30 +1,67 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { ChangeEvent } from 'react';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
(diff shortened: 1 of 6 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,67 @@
...
+};
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
-    </p>
-  </form>
-);
+export const TodoFilter: React.FC<Props> = ({
```
alexander-ignatow, line 10: JFYI You can safely skip React.FC 
https://github.com/typescript-cheatsheets/react?tab=readme-ov-file#section-2-getting-started

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,29 @@
...
+      .finally(() => setLoading(false));
+  }, [todo.userId]);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {loading ? (
```
alexander-ignatow, line 26: I don't quite like this ternary, so for the future you may consider following:

```suggestion
    if (loading) {
      return <Loader />
    }
    
    return (
      <div className="modal-card">
       ...
       ... render your modal in normal mode
```

alexander-ignatow (approved):
Good just see my comments for the future

PR #2360, opened 2024-05-15
Solution

luyx0:
[DEMO LINK](https://luyx0.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/Filter/Filter.tsx
+++ b/src/components/Filter/Filter.tsx
@@ -1,0 +1,14 @@
+import { Todo } from "../../types/Todo";
+import { IQuery, SortField } from "../../types/Filter";
+
+export const getFilteredTodos = (initialTodos: Todo[], query: IQuery) => {
+  const { status, query: searchQuery } = query;
+
+  return initialTodos.filter(todo => {
+    const isMatchingStatus =
+      status === SortField.Active ? !todo.completed : status === SortField.Completed ? todo.completed : true;
+    const isMatchingSearchQuery = !searchQuery || todo.title.toLowerCase().includes(searchQuery.toLowerCase());
+
+    return isMatchingStatus && isMatchingSearchQuery;
+  });
+};
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,30 +1,75 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
+import React, {useState} from "react";
+import {IQuery, SortField} from "../../types/Filter";
+
+interface ITodoFilter {
+  setQuery: React.Dispatch<React.SetStateAction<IQuery>>;
+}
+
+export const TodoFilter: React.FC<ITodoFilter> = ({setQuery}) => {
+  const [inputValue, setInputValue] = useState('')
+
+  const handleSelect = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const status = event.target.value as SortField
+
+    setQuery(prev => {
+      return {...prev, status}
+    })
+  }
+
+  const handleInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setInputValue(event.target.value);
+
+    setQuery(prevQuery => {
+      return { ...prevQuery, query: event.target.value };
+    });
+  };
+
+  const handleDeleteButton = () => {
+    setQuery({ status: 'all', query: '' });
+    setInputValue('');
+  };
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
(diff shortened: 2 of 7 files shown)


JohnyJDX (approved):

src/types/Filter.ts
```diff
@@ -0,0 +1,10 @@
+export enum SortField {
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
+
+export interface IQuery {
+  status: string;
```
alexander-ignatow, line 8: why the `status` is string and not `SortField`?

src/components/Filter/Filter.tsx
alexander-ignatow, line 1: It is not a component, so 
1. Not `tsx` but `ts`
2. Lets rename this file and move it - for instance to `utils/todoUtils`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,70 @@
...
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
-    </p>
-  </form>
-);
+import React, { useState } from 'react';
+import { IQuery, SortField } from '../../types/Filter';
+
+interface ITodoFilter {
+  setQuery: React.Dispatch<React.SetStateAction<IQuery>>;
```
alexander-ignatow, line 5: I don't quite like the idea of exposing such types

why not:
```suggestion
  setQuery: (query: IQuery) => void
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,70 @@
...
+}
+
+export const TodoFilter: React.FC<ITodoFilter> = ({ setQuery }) => {
+  const [inputValue, setInputValue] = useState('');
```
alexander-ignatow, line 9: this is the problem

See - you have at least two places to store similar state: this one and one in the `App`
You should always have one source of truth

I suggest you 
1. Remove this state
2. pass query here as a property from the App

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,70 @@
...
+    setInputValue(event.target.value);
+
+    setQuery(prevQuery => {
+      return { ...prevQuery, query: event.target.value };
+    });
+  };
+
+  const handleDeleteButton = () => {
+    setQuery({ status: 'all', query: '' });
+    setInputValue('');
```
alexander-ignatow, line 29: See - you perform a single clear operation, but have to set data in two places. It is the potential source of the bugs.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,70 @@
...
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select onChange={handleSelect} data-cy="statusSelect">
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
+          </select>
```
alexander-ignatow, line 40: This select goes in uncontrolled mode, while the input is used in controlled mode

It is not critical, but not perfect as well - if you pass query as prop, you can make it consistent

alexander-ignatow (changes requested):
Good, but let's make some changes

src/types/Filter.ts
luyx0: I set the initial values
useState<IQuery>({ status: 'all', query: '' });

undeletable: > useState({ status: 'all', query: '' });

`'all'` is one of  `SortField` enum values, so it makes sense to define `status` type as `SortField`.
Also, you should use enum for referring to value: `status: SortField.All` instead of `status: 'all'`.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,38 @@
...
+}
+
+const useUser = (userId: number) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    getUser(userId)
+      .then(setUser)
+      .catch(e => alert(e))
```
undeletable, line 20: It's not the best idea to use `alert()`, because it blocks the whole browser interface.
Usually, in such cases some custom warning component is rendered. I'm not sure if it's the part of your task, so you can just display console message - like you do in `App.tsx`.

undeletable (changes requested):
The solution looks fine after refactoring. Just please take care of the review comment below, and also of two things:
- use enum for `status` type definition and its usages (see [comment](https://github.com/mate-academy/react_dynamic-list-of-todos/pull/2360#discussion_r1606569839))
- please fix eslint warnings (you can see them in the console when running `npm start`)

PR #2362, opened 2024-05-15
add solution

D-Kondratiev:
[DEMO LINK](https://D-Kondratiev.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,60 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
+import { FilteredOptions } from './types/FilteredOption';
+import { getTodos } from './api';
+
+function handleFilteredTodos(
+  todos: Todo[],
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+  switch (selectedOption) {
+    case FilteredOptions.active:
+      return filteredTodos.filter(todo => !todo.completed);
+    case FilteredOptions.completed:
+      return filteredTodos.filter(todo => todo.completed);
+
+    default:
+      return filteredTodos;
+  }
+}
```
AYupko, line 36: better to move out function to its own file

AYupko (approved):
Good job! Approved!

src/App.tsx
alexander-ignatow: Makes sense

src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+
+  switch (selectedOption) {
+    case FilteredOptions.active:
+      return filteredTodos.filter(todo => !todo.completed);
+    case FilteredOptions.completed:
+      return filteredTodos.filter(todo => todo.completed);
+
+    default:
+      return filteredTodos;
+  }
```
alexander-ignatow, line 35: 1. No need to copy anything
2. Please try implement it with a single `filter` call


```suggestion
 return filteredTodos.filter(todo => {
 
 switch (selectedOption) {
    case FilteredOptions.active:
      return !todo.completed'
    case FilteredOptions.completed:
      return todo.completed;

    default:
      return true;
  }
 
 // TODO: implement check on query
 
 })
```

alexander-ignatow (changes requested):
Good 👍 , just one minor thing to fix

alexander-ignatow (approved):
Good!

PR #2367, opened 2024-05-15
solution

Olek-91:
[DEMO LINK](https://Olek-91.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/utils/getFilteredTodos.ts
+++ b/src/utils/getFilteredTodos.ts
@@ -1,0 +1,33 @@
+import { SortField } from '../types/SortField';
+import { Todo } from '../types/Todo';
+
+interface Filters {
+  sortField: SortField;
+  query: string;
+}
+
+export function getFilteredTodos(
+  todos: Todo[],
+  { sortField, query }: Filters,
+): Todo[] {
+  // const copyTodos = [...todos];
+
+  if (query) {
+    return todos.filter(todo => {
+      todo.title.trim().toLowerCase().includes(query.trim().toLowerCase());
+    });
+  }
+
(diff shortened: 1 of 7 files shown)


luyx0 (changes requested):
Good work but can be seen that tests do not go through because no functionality is implemented in the file Todofilter.tsx

Olek-91:
> Good work but can be seen that tests do not go through because no functionality is implemented in the file Todofilter.tsx

I know, thank you, I'm doing it so that Coach will see that I've started my homework.

Olek-91:
> Good work but can be seen that tests do not go through because no functionality is implemented in the file Todofilter.tsx

you can help me to fix this)))

src/utils/getFilteredTodos.ts
```diff
@@ -0,0 +1,38 @@
...
+      case SortField.completed:
+        copy = copy.filter(todo => todo.completed);
+        break;
+      default:
+        break;
+    }
+  }
+
+  return copy;
+}
```
alexander-ignatow, line 38: Not critical - but it can be re-written with just one `filter` call that is way more effective

```suggestion
export const getFilteredTodos = (todos: Todo[],  { sortField, query }: Filters) => todos.filter(
  theTodo => {
    if (query && !todo.toLowerCase().includes(query.toLowerCase()) { 
       return false
    }
    
    // check completed ...
    
    return true
  }
)
```

alexander-ignatow (changes requested):
Looks like something does not work

image

image

The code is good, but see also my comments

PR #2368, opened 2024-05-16
add task solution

TetianaVeremchuk:
[DEMO LINK](https://tetianaveremchuk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


JohnyJDX (approved):

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import type { Todo } from './types/Todo';
+import { StatusFilter } from './types/StatusFilter';
 
 export const App: React.FC = () => {
```
alexander-ignatow, line 14: React.FC is not needed
https://github.com/typescript-cheatsheets/react?tab=readme-ov-file#section-2-getting-started

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+    getTodos()
+      .then(setTodos)
+      .catch(error => error.message)
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const todosToRender = useMemo(
+    () =>
+      todos.filter(todo => {
+        let hasToBeRendered = true;
```
alexander-ignatow, line 31: the code will be cleaner if you remove `hasToBeRendered` and will just return `false` if the item should be skipped

See below

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+
+        switch (statusFilter) {
+          case StatusFilter.ALL:
+            break;
+
+          case StatusFilter.ACTIVE:
+            hasToBeRendered = !todo.completed;
```
alexander-ignatow, line 38:
```suggestion
   if (todo.completed) { return false; }
```

TetianaVeremchuk:
> Very good implementation 👍 !
> 
> See my comments as a note for the future

Thank you!))

PR #2371, opened 2024-05-16
add task solution

SlawikH:
[DEMO LINK](https://SlawikH.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable max-len */
 import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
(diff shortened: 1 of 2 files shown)


ilyasolo881 (approved):
Nice work!

src/App.tsx
```diff
@@ -1,14 +1,54 @@
...
+
+const FILTER_MAP = {
+  [SelectOptions.All]: () => true,
+  [SelectOptions.Active]: (todo: Todo) => !todo.completed,
+  [SelectOptions.Completed]: (todo: Todo) => todo.completed,
+};
+
+function applyFilter(arr: Todo[], selectedOption: SelectOptions): Todo[] {
+  return arr.filter(FILTER_MAP[selectedOption]);
+}
```
alexander-ignatow, line 21: This idea with map is interesting, but you break filtering logic in several parts:

one is in `applyFilter` and another is inlined in the `App` component

You may keep the map and several methods, but I let's be consistent and put all parts of the filtering logic in one place

I recommend you placing them in a separate file, say `utils/todoFilter` or `utils/todoUtils`

alexander-ignatow (changes requested):
Good, but one minor change is needed

mrkovaliv (approved):

PR #2375, opened 2024-05-16
add soluion

stepanyukvova200:
[DEMO LINK](https://stepanyukvova200.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,100 +1,78 @@
 import React from 'react';
+import {Todo} from "../../types/Todo";
+import classNames from "classnames";
 
-export const TodoList: React.FC = () => (
-  <table className="table is-narrow is-fullwidth">
-    <thead>
+type Props = {
+  todos: Todo[];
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,82 @@
...
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+type Props = {
+  todos: Todo[];
+  onActiveModal: React.Dispatch<React.SetStateAction<boolean>>;
+  onSetTodo: React.Dispatch<React.SetStateAction<Todo | undefined>>;
```
denyssheremeta, line 8: will be better to use null instead undefined

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,82 @@
...
+  isModalActive: boolean;
+  activeTodo: Todo | undefined;
```
denyssheremeta, line 10: will be better to use null instead undefined

denyssheremeta (approved):
Good job!
Check some issues.

PR #2397, opened 2024-05-30
react-list-todo

vlad-fil:
[DEMO LINK](https://vlad-fil.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -8,8 +8,8 @@ type Props = {
 
 export const TodoFilter: React.FC<Props> = ({
   query,
-  onQueryChange = () => {},
-  onStatusChange = () => {},
+  onQueryChange,
+  onStatusChange,
 }) => (
   <form className="field has-addons">
     <p className="control">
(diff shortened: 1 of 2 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,74 @@
...
-            <i className="fas fa-check" />
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+type Props = {
+  todos: Todo[];
+  onSelect?: (todo: Todo | null) => void;
```
etojeDenys, line 6:
```suggestion
  onSelect: (todo: Todo | null) => void;
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,23 @@
-export const TodoFilter = () => (
+import React from 'react';
+
+type Props = {
+  query: string;
+  onQueryChange?: (query: string) => void;
+  onStatusChange?: (status: string) => void;
```
etojeDenys, line 6:
```suggestion
  onQueryChange: (query: string) => void;
  onStatusChange: (status: string) => void;
```

etojeDenys (changes requested):
why do you make those function optional, let's fix that

vadiimvooo (approved):
Well done

PR #2406, opened 2024-06-06
solution

VladimirKolesnikov:
[DEMO LINK](https://VladimirKolesnikov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/types/TodoStatus.ts
+++ b/src/types/TodoStatus.ts
@@ -1,0 +1,1 @@
+export type TodoStatus = 'all' | 'completed' | 'active';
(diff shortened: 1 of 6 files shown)


src/types/TodoStatus.ts
```diff
@@ -0,0 +1 @@
+export type TodoStatus = 'all' | 'completed' | 'active';
```
DarkMistyRoom, line 1: better to use Enum

DarkMistyRoom (approved):
Good job!

PR #2415, opened 2024-06-14
solution

YuliiaCat:
https://yuliiacat.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,100 +1,80 @@
 import React from 'react';
+import { Todo } from '../../types/Todo';
+import classNames from 'classnames';
 
-export const TodoList: React.FC = () => (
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,80 @@
...
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              )}
+            </td>
+            <td className="is-vcentered is-expanded">
+              <p
+                className={
+                  todo.completed ? 'has-text-success' : 'has-text-danger'
+                }
```
SanyaBratashchuk, line 56: it is better to use libraries like classnames or clsx for coditiona classnames

SanyaBratashchuk (approved):

PR #2426, opened 2024-06-19
solution

KaterinaZhlobinskaya:
[DEMO LINK](https://KaterinaZhlobinskaya.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,7 +16,7 @@ type Props = {
   status: TodoStatus;
 };
 
-const filteredTodos = (todos: Todo[], { query, status }: Props) => {
+const getFilteredTodos = (todos: Todo[], { query, status }: Props) => {
   let filterTodos = todos;
 
   if (query.trim()) {
@@ -52,15 +52,15 @@ export const App: React.FC = () => {
       .finally(() => setLoading(false));
   }, []);
 
-  const filterTodos = filteredTodos(todos, { query, status });
+  const filterTodos = getFilteredTodos(todos, { query, status });
 
-  const handleSelect = (todo: Todo | null) => {
-    setSelectedTodo(todo);
-  };
+  // const handleSelect = (todo: Todo | null) => {
+  //   setSelectedTodo(todo);
+  // };
 
-  const handleClose = () => {
-    setSelectedTodo(null);
-  };
+  // const handleClose = () => {
+  //   setSelectedTodo(null);
+  // };
 
   return (
     <>
@@ -84,7 +84,7 @@ export const App: React.FC = () => {
                 <TodoList
                   todos={filterTodos}
                   selectedTodo={selectedTodo}
-                  onSelectedTodo={handleSelect}
+                  onSelectedTodo={setSelectedTodo}
                 />
               )}
             </div>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+import { Todo } from './types/Todo';
+import { TodoStatus } from './types/TodoStatus';
+import { getTodos } from './api';
+
+type Props = {
+  query: string;
+  status: TodoStatus;
+};
+
+const filteredTodos = (todos: Todo[], { query, status }: Props) => {
```
TarasHoliuk, line 19:
```suggestion
const getFilteredTodos = (todos: Todo[], { query, status }: Props) => {
```

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+  const [loading, setLoading] = useState(true);
+  const [query, setQuery] = useState('');
+  const [status, setStatus] = useState(TodoStatus.all);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        return (
```
TarasHoliuk, line 48: it won't work

You can't return JSX inside the catch callback to render it somewhere
You should change some state value (for example errorMessage) which will cause a rerender

Just FYI

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+          <p style={{ color: 'red' }}>Something went wrong, try again later</p>
+        );
+      })
+      .finally(() => setLoading(false));
+  }, []);
+
+  const filterTodos = filteredTodos(todos, { query, status });
+
+  const handleSelect = (todo: Todo | null) => {
```
TarasHoliuk, line 57: redundant handleSelect function wrapper. Just use `setSelectedTodo` everywhere instead

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,29 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { User } from '../../types/User';
+import { Todo } from '../../types/Todo';
+import { getUser } from '../../api';
+
+type Props = {
+  selectedTodo: Todo;
+  onCloswModal: () => void;
```
TarasHoliuk, line 9:
```suggestion
  onCloseModal: () => void;
```
typo

TarasHoliuk (changes requested):
GJ 👍 

Added a few small comments

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+      .finally(() => setLoading(false));
+  }, []);
+
+  const filterTodos = getFilteredTodos(todos, { query, status });
+
+  // const handleSelect = (todo: Todo | null) => {
+  //   setSelectedTodo(todo);
+  // };
+
+  // const handleClose = () => {
```
TarasHoliuk, line 61: it's better not to leave comments - just delete code you don't need anymore

BTW - handleClose function makes sense (it doesn't accept any arguments so you can just pass it as prop), when handleSelect just duplicated the original function

FYI only

TarasHoliuk (approved):
GJ 👍

PR #2435, opened 2024-06-27
need to fix filtering

Viktoria-Moroz:
[DEMO LINK](https://Viktoria-Moroz.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,94 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { FilterStatus } from './types/filter';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectTodos, setSelectTodos] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState(FilterStatus.All);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const delay = 300;
+
+  const filteredTodos = useMemo(() => {
+    let filteredTodo = todos;
+
+    if (filter === FilterStatus.Active) {
+      filteredTodo = filteredTodo.filter(todo => !todo.completed);
+    }
+
+    if (filter === FilterStatus.Completed) {
+      filteredTodo = filteredTodo.filter(todo => todo.completed);
+    }
+
+    if (appliedQuery) {
+      filteredTodo = filteredTodo.filter(todo =>
+        todo.title.toLowerCase().includes(appliedQuery.toLowerCase().trim()),
+      );
+    }
+
+    return filteredTodo;
+  }, [todos, filter, appliedQuery]);
+
+  const debounce = (callback: (args: string) => void) => {
+    let timerId = 0;
+
+    return (args: string) => {
+      window.clearTimeout(timerId);
+
+      timerId = window.setTimeout(() => {
+        callback(args);
+      }, delay);
+    };
+  };
+
+  const applyDebounceQuery = useCallback(
+    debounce((value: string) => {
+      setAppliedQuery(value);
+    }),
+    [debounce],
+  );
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    applyDebounceQuery(event.target.value);
+    setQuery(event.target.value);
+  };
+
+  useEffect(() => {
+    getTodos().then(todosFromServer => {
+      setTodos(todosFromServer);
+    });
+  }, []);
+
+  const handleShowModal = (userId: number) => {
+    const foundTodo = todos.find(todo => todo.id === userId);
+
+    if (foundTodo) {
+      setSelectTodos(foundTodo);
+    }
+  };
+
+  function changeStatusOfTodos(value: FilterStatus) {
+    setFilter(value);
+  }
+
+  const handleClearSearchBtn = () => {
+    if (appliedQuery && query) {
+      setQuery('');
+      setAppliedQuery('');
+    }
+  };
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,14 +1,94 @@
...
+    return (args: string) => {
+      window.clearTimeout(timerId);
+
+      timerId = window.setTimeout(() => {
+        callback(args);
+      }, delay);
+    };
+  };
+
+  const applyDebounceQuery = useCallback(
```
TarasHoliuk, line 55: you can use `useMemo` here
or just disable this lint rule for this line

Please, create the debounce function outside of the component and remove it from the dependency array

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,60 @@
...
+  const handleChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    changeStatusOfTodos(event.target.value as FilterStatus);
+  };
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select data-cy="statusSelect" onChange={handleChange}>
+            <option value={FilterStatus.All}>All</option>
```
TarasHoliuk, line 26: It looks better with `All`, but tests require `all`:

image

Change it to pas tests

src/App.tsx
```diff
@@ -1,14 +1,94 @@
...
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectTodos, setSelectTodos] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState(FilterStatus.All);
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+
+  const delay = 300;
+
+  const filteredTodos = useMemo(() => {
+    let filteredTodo = todos;
```
TarasHoliuk, line 24:
```suggestion
    let filteredTodos = todos;
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,73 @@
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye-slash" />
-            </span>
-          </button>
-        </td>
-      </tr>
+type Props = {
+  filteredTodos: Todo[] | undefined;
```
TarasHoliuk, line 7:
```suggestion
  todos: Todo[];
```
Minor naming suggestion: prop name can be just todos - this component just renders a list of any todos (and doesn't "know" if it is filtered or sorted, etc.)

TarasHoliuk (changes requested):
You did a good job 👍 
Added a few minor comments

Try to fix all tests 🙂

src/App.tsx
volodymyr-soltys97: Not fixed 
image

src/App.tsx
volodymyr-soltys97: Not fixed 
image

src/App.tsx
```diff
@@ -1,14 +1,83 @@
...
+
+  const handleShowModal = (userId: number) => {
+    const foundTodo = todos.find(todo => todo.id === userId);
+
+    if (foundTodo) {
+      setSelectTodos(foundTodo);
+    }
+  };
+
+  function changeStatusOfTodos(value: FilterStatus) {
```
volodymyr-soltys97, line 72: It is important to follow the same way of creating functions everywhere
```suggestion
  const changeStatusOfTodos = (value: FilterStatus) => {
```

src/App.tsx
```diff
@@ -17,18 +86,31 @@ export const App: React.FC = () => {
...
+                handleQueryChange={handleQueryChange}
+                handleClearSearchBtn={handleClearSearchBtn}
+                query={query}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length > 0 ? (
```
volodymyr-soltys97, line 98:
```suggestion
              {!!todos.length ? (
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,73 @@
...
+                )}
+              </td>
+              <td className="is-vcentered is-expanded">
+                <p
+                  className={cn({
+                    'has-text-danger': !todo.completed,
+                    'has-text-success': todo.completed,
+                  })}
+                >
+                  {todo.title}
```
volodymyr-soltys97, line 51: Use destruction for `todo`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,73 @@
...
+              <td className="has-text-right is-vcentered">
+                <button
+                  data-cy="selectButton"
+                  className="button"
+                  type="button"
+                  onClick={() => handleShowModal(todo.id)}
+                >
+                  <span className="icon">
+                    <i
+                      className={`far ${selectTodos === todo ? 'fa-eye-slash' : 'fa-eye'}`}
```
volodymyr-soltys97, line 63: Use the classnames library rules for add classes with condition

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/App.tsx
```diff
@@ -8,7 +8,77 @@ import { TodoFilter } from './components/TodoFilter';
...
 import { Loader } from './components/Loader';
 
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { FilterStatus } from './types/filter';
+import { debounce } from './utils/debounce';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectTodos, setSelectTodos] = useState<Todo | null>(null);
```
etojeDenys, line 18:
```suggestion
  const [selectTodo, setSelectTodo] = useState<Todo | null>(null);
```
it's only one todo

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+import { getUser } from '../../api';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  selectTodos: Todo;
+  setSelectTodos: (todo: Todo | null) => void;
+};
+
+export const TodoModal: React.FC<Props> = ({ selectTodos, setSelectTodos }) => {
+  const [users, setUsers] = useState<User | undefined>();
```
etojeDenys, line 14:
```suggestion
  const [user, setUser] = useState<User | null>();
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+  const [isClosed, setIsClosed] = useState(true);
```
etojeDenys, line 15: looks strange, why do you need this state

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+
+  useEffect(() => {
+    getUser(selectTodos.userId).then(usersFromServer => {
+      setUsers(usersFromServer);
+    });
+  }, [selectTodos.userId]);
+
+  useEffect(() => {
+    setIsClosed(false);
+  }, [selectTodos]);
```
etojeDenys, line 25: same here

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+
+  const handleDeleteBtn = () => {
+    setIsClosed(true);
+    setSelectTodos(null);
+  };
```
etojeDenys, line 30:
```suggestion
  const handleDeleteBtn = () => {
    setSelectTodos(null);
  };
```
should be enough

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+
+export const TodoModal: React.FC<Props> = ({ selectTodos, setSelectTodos }) => {
+  const [users, setUsers] = useState<User | undefined>();
+  const [isClosed, setIsClosed] = useState(true);
+
+  useEffect(() => {
+    getUser(selectTodos.userId).then(usersFromServer => {
+      setUsers(usersFromServer);
+    });
+  }, [selectTodos.userId]);
```
etojeDenys, line 21: just create the isLoading state, and set false after the user has loaded

volodymyr-soltys97 (approved):
Great work 🔥

PR #2437, opened 2024-06-28
Add dynamic list of todos

GoVolodya:
[DEMO LINK](https://GoVolodya.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,77 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [filter, setFilter] = useState('all');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [activeTodo, setActiveTodo] = useState<Todo | undefined>(undefined);
+  const [query, setQuery] = useState('');
+  const [loadingTodos, setLoadingTodos] = useState(true);
+
+  useEffect(() => {
+    setLoadingTodos(true);
+
+    switch (filter) {
+      case 'all':
+        getTodos()
+          .then(allTodos =>
+            allTodos.filter(todo =>
+              todo.title.toLowerCase().includes(query.toLowerCase()),
+            ),
+          )
+          .then(setTodos)
+          .finally(() => setLoadingTodos(false));
+        break;
+      case 'active':
+        getTodos()
+          .then(allTodos =>
+            allTodos.filter(
+              todo =>
+                !todo.completed &&
+                todo.title.toLowerCase().includes(query.toLowerCase()),
+            ),
+          )
+          .then(setTodos)
+          .finally(() => setLoadingTodos(false));
+        break;
+      case 'completed':
+        getTodos()
+          .then(allTodos =>
+            allTodos.filter(
+              todo =>
+                todo.completed &&
+                todo.title.toLowerCase().includes(query.toLowerCase()),
+            ),
+          )
+          .then(setTodos)
+          .finally(() => setLoadingTodos(false));
+        break;
+      default:
+        setTodos([]);
+        setLoadingTodos(false);
+        break;
+    }
+  }, [filter, query]);
+
+  const handleFilter = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setFilter(event.target.value);
+  };
+
+  const handleModalOpen = (activeTodoItem: Todo | undefined) => {
+    setActiveTodo(activeTodoItem);
+  };
+
+  const handleModalClose = () => {
+    setActiveTodo(undefined);
+  };
+
+  const handleQueryChange = (newQuery: string) => {
+    setQuery(newQuery);
+  };
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,83 @@
...
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [filter, setFilter] = useState('all');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [activeTodo, setActiveTodo] = useState<Todo | undefined>(undefined);
+  const [query, setQuery] = useState('');
+  const [loadingTodos, setLoadingTodos] = useState(true);
```
TarasHoliuk, line 18:
```suggestion
  const [isLoadingTodos, setIsLoadingTodos] = useState(true);
```
for boolean variables: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/App.tsx
```diff
@@ -1,14 +1,83 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [filter, setFilter] = useState('all');
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [activeTodo, setActiveTodo] = useState<Todo | undefined>(undefined);
```
TarasHoliuk, line 16:
```suggestion
  const [activeTodo, setActiveTodo] = useState<Todo | null>(null);
```
Not critical, but it's more precise - we know that active todo is just not selected

src/App.tsx
```diff
@@ -1,14 +1,83 @@
...
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [filter, setFilter] = useState('all');
```
TarasHoliuk, line 14: Would be nice to use enum or union type for filter values

src/App.tsx
```diff
@@ -1,14 +1,83 @@
...
+  const [activeTodo, setActiveTodo] = useState<Todo | undefined>(undefined);
+  const [query, setQuery] = useState('');
+  const [loadingTodos, setLoadingTodos] = useState(true);
+
+  useEffect(() => {
+    setLoadingTodos(true);
+
+    switch (filter) {
+      case 'all':
+        getTodos()
```
TarasHoliuk, line 25: No need to fetch todos every time filter changes
Fetch them once. Then just save in the common variable (const) filtered todos. It's better to use  the `useMemo` hook to memorize value and recompute filtered todos only on filter or query change

src/App.tsx
```diff
@@ -1,14 +1,83 @@
...
+        setLoadingTodos(false);
+        break;
+    }
+  }, [filter, query]);
+
+  const handleFilter = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setFilter(event.target.value);
+  };
+
+  const handleModalOpen = (activeTodoItem: Todo | undefined) => {
```
TarasHoliuk, line 69: this wrapper is redundant - just use `setActiveTodo` instead of `handleModalOpen` everywhere

src/App.tsx
```diff
@@ -1,14 +1,83 @@
...
+    setActiveTodo(activeTodoItem);
+  };
+
+  const handleModalClose = () => {
+    setActiveTodo(undefined);
+  };
+
+  const handleQueryChange = (newQuery: string) => {
```
TarasHoliuk, line 77: same as `handleModalOpen` - redundant function wrapper

TarasHoliuk (changes requested):
Great job! Added a few suggestions to refactor your code

src/App.tsx
GoVolodya: Done, will keep that post in mind.

src/App.tsx
GoVolodya: Done, that should make more sense now.

src/App.tsx
GoVolodya: oh, yes, you are correct, removed this.

src/App.tsx
GoVolodya: Good point, less code easier to maintain.

src/App.tsx
GoVolodya: Okay, for this task this will work as we have static todos. We can change it later to update todos periodically or something if needed. Also will save some resources for Mate Academy =)

src/App.tsx
GoVolodya: Cool suggestion, like it and need to use it more often for some static values.

GoVolodya:
Thanks for suggestions, I really appreciate this.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,75 @@
...
-  </table>
-);
+      <tbody>
+        {todos.map(todo => (
+          <tr
+            key={todo.id}
+            data-cy="todo"
+            className={classNames({
+              'has-background-info-light':
+                activeTodo && activeTodo.id === todo.id,
```
etojeDenys, line 32:
```suggestion
                activeTodo?.id === todo.id,
```

src/App.tsx
```diff
@@ -1,14 +1,78 @@
...
+        break;
+      case Filter.Completed:
+        cachedTodos
+          .then(allTodos =>
+            allTodos.filter(
+              todo =>
+                todo.completed &&
+                todo.title.toLowerCase().includes(query.toLowerCase()),
+            ),
+          )
```
etojeDenys, line 57: it looks strange and complex, just move it to some function that will return the filtered array
`const visibleTodos = getVisibleTodos(todos, filter, query)`

and then use that array everywhere

etojeDenys (changes requested):

src/components/TodoList/TodoList.tsx
GoVolodya: Sure, must have missed some while refactoring.

GoVolodya:
New version of dynamic todos is ready. Please have a look.

SanyaBratashchuk (approved):

PR #2443, opened 2024-07-03
Develop

DanielLinker17:
https://DanielLinker17.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -9,19 +9,16 @@ export const TodoList: React.FC<TodoListProps> = ({
   filterCategory,
 }) => {
   const preparedTodos = todos.filter(todo => {
-    if (filterCategory === 'all') {
-      return true;
+    switch (filterCategory) {
+      case 'all':
+        return true;
+      case 'completed':
+        return todo.completed;
+      case 'active':
+        return !todo.completed;
+      default:
+        return true;
     }
-
-    if (filterCategory === 'completed') {
-      return todo.completed;
-    }
-
-    if (filterCategory === 'active') {
-      return !todo.completed;
-    }
-
-    return true;
   });
 
   const filteredTodos = preparedTodos.filter(todo =>
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,100 +1,87 @@
 import React from 'react';
+import { TodoListProps } from './TodoList.types';
 
-export const TodoList: React.FC = () => (
-  <table className="table is-narrow is-fullwidth">
-    <thead>
-      <tr>
-        <th>#</th>
-        <th>
-          <span className="icon">
-            <i className="fas fa-check" />
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+export const TodoList: React.FC<TodoListProps> = ({
+  todos,
+  filter,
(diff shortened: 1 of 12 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,41 @@
...
+  setSelectedTodo,
+}) => {
+  const [userName, setUserName] = useState('');
+  const [userIsLoading, setUserIsLoading] = useState(false);
+  const targetUserId = selectedTodo?.userId;
+
+  useEffect(() => {
+    if (targetUserId) {
+      setUserIsLoading(true);
+      setTimeout(() => {
```
etojeDenys, line 18: why do we need timeout here, looks redundant

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,87 @@
...
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye-slash" />
-            </span>
-          </button>
-        </td>
-      </tr>
+    if (filterCategory === 'completed') {
+      return todo.completed;
+    }
```
etojeDenys, line 18: better to use switch case here, instead of multiple if

etojeDenys (changes requested):

src/components/TodoModal/TodoModal.tsx
DanielLinker17: Didn't pass tests otherwise 
Screenshot 2024-07-03 at 17 37 39

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+      setUserIsLoading(true);
+      getUser(targetUserId)
+        .then(user => {
+          setUserName(user.name);
+        })
+        .finally(() => {
+          setUserIsLoading(false);
+        });
+    }
+  }, []);
```
TarasHoliuk, line 26: just add it to the dependency array:
image

src/components/TodoModal/TodoModal.types.ts
```diff
@@ -0,0 +1,6 @@
+import { Todo } from '../../types/Todo';
+
+export type TodoModalProps = {
+  selectedTodo: Todo | null;
```
TarasHoliuk, line 4:
```suggestion
  selectedTodo: Todo;
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
 import { Loader } from '../Loader';
+import { TodoModalProps } from './TodoModal.types';
+import { getUser } from '../../utils/getUser';
+
+export const TodoModal: React.FC<TodoModalProps> = ({
+  selectedTodo,
+  setSelectedTodo,
+}) => {
+  const [userName, setUserName] = useState('');
+  const [userIsLoading, setUserIsLoading] = useState(false);
```
TarasHoliuk, line 12: a bit strange that the test with the loader fails w/o timeout
try to set it true by default

TarasHoliuk (changes requested):

src/components/TodoModal/TodoModal.tsx
DanielLinker17: Doesn't work

TarasHoliuk (approved):

PR #2448, opened 2024-07-03
add initial solution for the react_dynamic-list-of-todos task

lesdyman:
Hey there! I'm happy to introduce my hard-won solution for the react_dynamic-list-of-todos task. Can't denny, this was really difficult piece. But here we are though. As usual you shall find demo link below. Hope my solution is't that bad.

[DEMO LINK](https://lesdyman.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,71 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import { getTodos } from './api';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+
+  const [filter, setFilter] = useState('all');
+  const [searchInput, setSearchInput] = useState('');
+
+  const [isLoading, setIsLoading] = useState(true);
+
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [modalTodo, setModalTodo] = useState<Todo | null>(null);
+
+  const [selectedTodoId, setSelectedTodoId] = useState<number | null>(null);
+
+  useEffect(() => {
+    getTodos().then(todosFromServer => {
+      setTodos(todosFromServer);
+      setIsLoading(false);
+    });
+  }, []);
+
+  const filteredTodos = useMemo(() => {
+    let result = todos;
+
+    switch (filter) {
+      case 'active':
+        result = result.filter(todo => !todo.completed);
+        break;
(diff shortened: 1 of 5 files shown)


yaroslav1177 (approved):
Your solution is not bad at all and i agree that this task wasn`t easy. But from my point of view you did it great! Definitely approve!

lesdyman:
After the class, I realised that I needed to improve some code fragments a little bit. So I did that and pushed the updated solution.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,36 @@
...
+export const TodoModal: React.FC<Props> = ({ todo, closeModal }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getUser(todo.userId)
+      .then(userData => {
+        setUser(userData);
+        setIsLoading(false);
```
artemforsoff, line 21: you can move the call to this function into a `finally` block like this:
```
getUser(todo.userId)
  .then(userData => {
    setUser(userData);
  })
  .catch(error => {
    throw new Error(error);
  })
  .finally(() => {
     setIsLoading(false);
  });
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,68 @@
...
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
-    </p>
-  </form>
-);
+import React, { useState } from 'react';
+import { FilterType } from '../../types/FilterType';
+
+type Props = {
+  setFilter: (filter: string) => void;
```
artemforsoff, line 5: `setFilter: (filter: FilterType) => void;`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,68 @@
...
+    setFilter(FilterType.All);
+    setSearchInput('');
+  };
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select data-cy="statusSelect" onChange={handleSelect}>
+            <option value="all">All</option>
```
artemforsoff, line 34: also better use `FilterType` here:
`<option value={FilterType.all}>All</option>`

src/App.tsx
```diff
@@ -1,14 +1,72 @@
...
 import { TodoFilter } from './components/TodoFilter';
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { FilterType } from './types/FilterType';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+
+  const [filter, setFilter] = useState('all');
```
artemforsoff, line 17: `useState(FilterType.all);`

artemforsoff (approved):
overall great beautiful work 🔥

lesdyman:
@artemforsoff Thanks for your review! I'll certainly use those advices to improve my solution.

PR #2451, opened 2024-07-04
solution

yaroslav1177:
[DEMO LINK](https://yaroslav1177.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Todo } from '../../types/Todo';
-import cn from 'classnames';
+import { TodoItem } from '../TodoItem/TodoItem';
 
 type Props = {
   todos: Todo[];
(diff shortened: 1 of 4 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,60 @@
...
+  setQuery,
+}) => {
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={filterBy}
+            onChange={event => setFilterBy(event.target.value as Filter)}
```
lesdyman, line 24: It works. But I'm not sure about this way of force casting to be a good practice. I mean this part "event.target.value as Filter". What is gonna to happened if event.taget.valuse will not comply with type Filter? Again, I'm not sure about this one, but maybe you should give it another look)

lesdyman (approved):
Well done! I left only one comment. Maybe it will be useful for you. Or maybe not)). But in either way, have a look. Other than that, from my point of view, you've done a good job.

yaroslav1177:
thanks for the comment, I'll see what I can do

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,80 @@
...
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map(todo => (
+          <tr
+            key={todo.id}
+            data-cy="todo"
+            className={cn({
+              'has-background-info-light': selectedTodoId === todo.id,
```
Moroz-Dmytro, line 37: Create variable for `selectedTodoId === todo.id` you use it 3 times

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,80 @@
...
+            })}
+          >
```
Moroz-Dmytro, line 39: Would be good to move `<tr>...</tr>` to a new component

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,26 @@
...
+import { Todo } from '../../types/Todo';
+import { User } from '../../types/User';
+import { getUser } from '../../api';
+
+type Props = {
+  selectedTodo: Todo;
+  onReset: () => void;
+};
+
+export const TodoModal: React.FC<Props> = ({ selectedTodo, onReset }) => {
```
Moroz-Dmytro, line 12:
```suggestion
export const TodoModal: React.FC<Props> = ({ selectedTodo, onClose }) => {
```

Moroz-Dmytro (changes requested):
Good work 👍 
Really small comments below

Moroz-Dmytro (approved):
Good work 👍 
Small advice below

PR #2454, opened 2024-07-04
add task solution

V-Shut:
[DEMO LINK](https://v-shut.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,38 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [openedTodo, setOpenedTodo] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState('all');
+  const [search, setSearch] = useState('');
+  const [timerIsActive, setTimerIsActive] = useState(false);
+  const [list, setList] = useState<Todo[]>([]);
+
+  const reset = () => {
+    setOpenedTodo(null);
+  };
+
+  useEffect(() => {
+    const fetchedTodos = async () => {
+      setTimerIsActive(true);
+      const todos = await getTodos();
+
+      setList(todos);
+      setTimerIsActive(false);
+    };
+
+    fetchedTodos();
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +41,30 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                setFilter={setFilter}
+                setSearch={setSearch}
+                search={search}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {timerIsActive && <Loader />}
+              {!timerIsActive && (
+                <TodoList
+                  openedTodo={openedTodo}
+                  setOpenedTodo={setOpenedTodo}
+                  filter={filter}
+                  search={search}
+                  standartList={list}
+                />
+              )}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {openedTodo && <TodoModal todo={openedTodo} reset={reset} />}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,8 +1,23 @@
-export const TodoFilter = () => (
+import React from 'react';
+
+interface Props {
+  setFilter: (value: string) => void;
+  setSearch: (value: string) => void;
+  search: string;
+}
+
+export const TodoFilter: React.FC<Props> = ({
+  setFilter,
+  setSearch,
+  search,
+}) => (
   <form className="field has-addons">
     <p className="control">
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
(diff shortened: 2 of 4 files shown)


HumenVAsya (approved):
good job)

src/App.tsx
```diff
@@ -1,14 +1,38 @@
...
+import { Todo } from './types/Todo';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [openedTodo, setOpenedTodo] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState('all');
+  const [search, setSearch] = useState('');
+  const [timerIsActive, setTimerIsActive] = useState(false);
+  const [list, setList] = useState<Todo[]>([]);
```
artemforsoff, line 18: better rename to `todos` or `todoList`
because the current name of the variable does not indicate what it stores

src/App.tsx
```diff
@@ -1,14 +1,38 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [openedTodo, setOpenedTodo] = useState<Todo | null>(null);
+  const [filter, setFilter] = useState('all');
```
artemforsoff, line 15: it would be nice to put the value for this variable in enam

src/App.tsx
```diff
@@ -1,14 +1,38 @@
...
+  const [search, setSearch] = useState('');
+  const [timerIsActive, setTimerIsActive] = useState(false);
```
artemforsoff, line 17: As I understand it, this state indicates that the request is loading. Therefore, it is better to rename it to `isLoading`

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,23 @@
-export const TodoFilter = () => (
+import React from 'react';
+
+interface Props {
+  setFilter: (value: string) => void;
```
artemforsoff, line 4: if we pass a function to a component that will be called on some event, then the best name for it is to start with a prefix `on`
like this `onSetFilter` or `onChangeFilter`, `onChangeSearchValue`, etc

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,101 @@
...
+  standartList: Todo[];
+}
+export const TodoList: React.FC<Props> = ({
+  openedTodo,
+  setOpenedTodo,
+  filter,
+  search,
+  standartList,
+}) => {
+  const [list, setList] = useState<Todo[]>(standartList);
```
artemforsoff, line 19: repeating the same state in different components is not good. The solution to this problem is to pass a function that will change the state in the parent component

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,101 @@
...
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye-slash" />
-            </span>
-          </button>
-        </td>
-      </tr>
+  const res = [...standartList].filter(el =>
```
artemforsoff, line 21: you should move it to `useEffect`
```
useEffect(() => {
  const res = [...standartList].filter(el =>
    el.title.toLowerCase().includes(search.trim().toLowerCase()),
  );
  switch (filter) {
    case 'all':
      setList(res);
      break;
    case 'active':
      setList(res.filter(el => el.completed === false));
      break;
    case 'active':
      setList(res.filter(el => el.completed === false));
      break;
   case 'completed':
      setList(res.filter(el => el.completed === true));
      break;
  }
}, [filter, search, standartList]);

File: src/components/TodoList/TodoList.tsx [hunk 7]
```diff
@@ -1,100 +1,101 @@
...
-        <td className="is-vcentered is-expanded">
-          <p className="has-text-success">quo adipisci enim quam ut ab</p>
-        </td>
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
+      case 'completed':
+        setList(res.filter(el => el.completed === true));
+        break;
+    }
+  }, [filter, search]);
```
artemforsoff, line 39: image

_React Hook useEffect has a missing dependency: 'res'. Either include it or remove the dependency array_

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,101 @@
...
+          <th> </th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {list.map(el => (
+          <tr data-cy="todo" className="" key={el.id}>
+            <td className="is-vcentered">{el.id}</td>
+            <td className="is-vcentered">
+              {el.completed ? (
```
artemforsoff, line 61: better use operator `&&`
like this:
`{isTrue && <>yes, its true</>}`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,101 @@
...
-);
+          </th>
+          <th>Title</th>
+          <th> </th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {list.map(el => (
+          <tr data-cy="todo" className="" key={el.id}>
```
artemforsoff, line 58: empty `className`
need to remove

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,36 @@
...
+    const fetchUser = async () => {
+      setIsLoading(true);
+      const person = await getUser(todo.userId);
+
+      setUser(person);
+      setIsLoading(false);
+    };
+
+    fetchUser();
+  }, []);
```
artemforsoff, line 27: }, [todo.userId]);

artemforsoff: image

artemforsoff (approved):
overall great beautiful work 🔥
please pay attention to my comments and recommend using them in the future

PR #2471, opened 2024-07-09
partial solution

gazebo88fire:
https://gazebo88fire.github.io/react_dynamic-list-of-todos/


djtyrf312 (changes requested):
Task is not a part of qa course

PR #2476, opened 2024-07-11
add solution

om1gse:
[DEMO LINK](https://om1gse.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


DSB-18 (approved):
Well done!

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [tasks, setTasks] = useState<Todo[]>([]);
```
polosanya, line 14: I guess its better to name it `todos`

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const [activeTodo, setActiveTodo] = useState<Todo | null>(null);
+  const [statusFilter, setStatusFilter] = useState('all');
```
polosanya, line 17: Would be great to use Enum instead of hardcoded value

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(setTasks)
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const filteredTasks = tasks
+    .filter(task => {
+      if (statusFilter === 'completed') {
```
polosanya, line 29: Let's use switch...case instead with Enum values

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+  const [searchQuery, setSearchQuery] = useState('');
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(setTasks)
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const filteredTasks = tasks
```
polosanya, line 27: useMemo could be used here to avoid calling filter after every change

src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+      }
+
+      if (statusFilter === 'active') {
+        return !task.completed;
+      }
+
+      return true;
+    })
+    .filter(task =>
+      task.title.toLowerCase().includes(searchQuery.toLowerCase()),
```
polosanya, line 40: Do you really need to call `filter` method twice?

polosanya (changes requested):
Nice! Let's improve your code a bit

PR #2478, opened 2024-07-12
add task solution

ropanefarias:
[DEMO LINK](https://ropanefarias.github.io/react_dynamic-list-of-todos/)


jv-aquino (approved):
Nice Rodrigo, there's only one 'logic' problem and that is: when you click to visualize the todo, the entire page is setting to loading mode, and I think is not the best approach. This is just simulating loading, but when you do it for real, try to create a loading for each of the  components if they're not dependant on each other

PR #2480, opened 2024-07-13
add solution

reznik-denis:
[DEMO LINK](https://reznik-denis.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,37 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
-import { TodoList } from './components/TodoList';
-import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
+import { MemoTodoList } from './components/TodoList';
+import { MemoTodoFilter } from './components/TodoFilter';
+import { MemoTodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { filterTodos } from './services/filteredTodos';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [userTodo, setUserTodo] = useState<Todo | null>(null);
+  const [inputQuery, setInputQuery] = useState('');
+  const [selQuery, setSelQuery] = useState('all');
+
+  useEffect(() => {
+    getTodos().then(data => {
+      setTodos(data);
+    });
+  }, []);
+
+  const handleModal = () => {
+    setUserTodo(null);
+  };
+
+  const filter = useMemo(() => {
+    return filterTodos(todos, selQuery, inputQuery);
+  }, [todos, selQuery, inputQuery]);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,37 @@
...
 
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { filterTodos } from './services/filteredTodos';
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [userTodo, setUserTodo] = useState<Todo | null>(null);
+  const [inputQuery, setInputQuery] = useState('');
+  const [selQuery, setSelQuery] = useState('all');
```
TarasHoliuk, line 19: selQuery is a bad name. It could be unclear what is `sel`. So never use short forms

But even select query is not a good choice. You can describe what exactly it is - it is status (todo status)

Also, create enum or union type for it (there are a few allowed values - it's not any string)

src/App.tsx
```diff
@@ -1,14 +1,37 @@
...
+    getTodos().then(data => {
+      setTodos(data);
+    });
+  }, []);
+
+  const handleModal = () => {
+    setUserTodo(null);
+  };
+
+  const filter = useMemo(() => {
```
TarasHoliuk, line 31:
```suggestion
  const filteredTodos = useMemo(() => {
```
filter is misleading variable name here

src/App.tsx
```diff
@@ -17,18 +40,28 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <MemoTodoFilter
+                onSelect={query => setSelQuery(query)}
```
TarasHoliuk, line 44:
```suggestion
                onSelect={setSelQuery}
```
No need to create additional function wrapper
Fix everywhere

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,84 @@
...
-            <i className="fas fa-check" />
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+type Props = {
+  todos: Todo[];
+  buttonSwitch: Todo | null;
```
TarasHoliuk, line 7: also about naming
It's not clear what is `buttonSwitch` here. It has Todo type (or null), so actually it's todo. So you should name this prop accordingly - `selectedTodo`. `selected` describes what kind of todo it is

naming is very important part of clear and easy to read code 🙂

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,84 @@
...
+              })}
+              key={todo.id}
+            >
+              <td className="is-vcentered">{todo.id}</td>
+              {todo.completed ? (
+                <td className="is-vcentered">
+                  <span className="icon" data-cy="iconCompleted">
+                    <i className="fas fa-check" />
+                  </span>
+                </td>
```
TarasHoliuk, line 45:
```suggestion
                <td className="is-vcentered">
                   {todo.completed && (
                     <span className="icon" data-cy="iconCompleted">
                      <i className="fas fa-check" />
                    </span>
                  )}
                </td>
```
td is necessary anyway so you can just render or not content iside

TarasHoliuk (changes requested):

reznik-denis:
done thanks for your comments

TarasHoliuk (approved):

PR #2482, opened 2024-07-15
add task solution

Oleksandr-Vol:
https://oleksandr-vol.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [completedStatus, setCompletedStatus] = useState('all');
```
TarasHoliuk, line 16: Create enum or union type for it
```suggestion
  const [completedStatus, setCompletedStatus] = useState('all');
```
```suggestion
  const [completedStatus, setCompletedStatus] = useState<Status>('all');
```

src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
```
TarasHoliuk, line 14:
```suggestion
  const [todos, setTodos] = useState<Todo[]>([]);
```
For array values you can set empty array as default value
It's easier to handle it properly in any case and allows to get rid of some redundant checks

src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
+
+        default:
+          return true;
+      }
+    });
+
+    filteredTodos = filterByStatus.filter(todo =>
+      todo.title.toLowerCase().includes(query.toLowerCase()),
+    );
+  }
```
TarasHoliuk, line 38:
```
  // logic here and return filtered array
}, [// dependencies])
```
use `useMemo` hook in such cases

TarasHoliuk (changes requested):
Great job!
Added a few small comments 🙂

anastasiiavorobiova (approved):
Great work!

PR #2491, opened 2024-07-21
My solution

Karanelus:
[DEMO LINK](https://karanelus.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos, getUser } from './api';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+
+export type FilterProperties = {
+  filterType: 'all' | 'active' | 'completed';
+  filterValue: string;
+};
```
Zibi95, line 17: Should be kept in separate file if used elsewhere.

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
+    if (type === 'completed' && todos) {
+      const filterCompleted = todos.filter(el => el.completed);
+
+      return isValueEmpty
+        ? filterCompleted
+        : filterCompleted.filter(el => filterCallback(el));
+    }
+
+    return todos;
+  }, [todos, filterProperties]);
```
Zibi95, line 91: This can be way more simpler. Just firstly filter by status and then by query. That's all.

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
+
+export type FilterProperties = {
+  filterType: 'all' | 'active' | 'completed';
+  filterValue: string;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null>(null);
+  const [userTodo, setUserTodo] = useState<Todo | null>(null);
+  const [user, setUser] = useState<User | null>(null);
```
Zibi95, line 22: You could keep the logic and state of user inside Modal and load i using `useEffect` there.

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
+  const [isInfoClicked, setIsInfoClick] = useState<boolean>(false);
```
Zibi95, line 23: This is not necessary. You can track whether modal is open with `userTodo`

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
+
+  const loadUser = async (userId: number) => {
+    const todosFromServer = await getUser(userId);
+
+    setUser(todosFromServer);
+  };
+
+  useEffect(() => {
+    loadTodos();
+    setUser(null);
```
Zibi95, line 49: user is null at this point its default value.

src/App.tsx
```diff
@@ -1,14 +1,95 @@
...
+  };
+
+  const loadUser = async (userId: number) => {
+    const todosFromServer = await getUser(userId);
+
+    setUser(todosFromServer);
+  };
+
+  useEffect(() => {
+    loadTodos();
```
Zibi95, line 48: If you use this function only inside than declare it inside effect body

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,26 @@
...
+  chosedTodo: Todo | null;
+};
+
+// completed: false
+// ​​​
+// id: 1
+// ​​​
+// title: "delectus aut autem"
+// ​​​
+// userId: 1
```
Zibi95, line 17: remove comments

src/components/TodoList/TodoInfo.tsx
```diff
@@ -0,0 +1,56 @@
+import classNames from 'classnames';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo;
+  onClickDetails: (todo: Todo) => void;
+  chosedTodo: Todo | null;
```
Zibi95, line 7: I think it would be better if you compute the value to boolean in TodoList component when you are mapping and just pass as prop isTodoChoosed

Zibi95 (approved):

Zibi95:
Good job i left some comments. There are some things to improve with state managing and with filter logic. Always try to pass as less complicated prop (i mean primitive values here) as you can that way its easier to optimize it in the future,

PR #2497, opened 2024-07-30
add solution

OlehPhw:
[DEMO LINK](https://OlehPhw.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
+    getTodos().then(todosFromServer => {
+      setTodos(todosFromServer);
+    });
+  }, []);
+
+  useEffect(() => {
+    if (selectedTodo) {
+      getUser(selectedTodo.userId).then(userFromServer => {
+        setUser(userFromServer);
+      });
```
volodymyr-soltys97, line 35:
```suggestion
      getUser(selectedTodo.userId)
        .then(setUser);
```

src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
+  ) => {
+    setFilterMethod(event.target.value);
+  };
+
+  const resetQuery = () => setQuery('');
+
+  const filtredTodos = todos.filter(todo => {
+    const hasQuery = todo.title.includes(query.trim().toLowerCase());
+
+    if (filterMethod === 'active') {
```
volodymyr-soltys97, line 54: Create a enum for 'active', 'completed', 'all' and use it everywhere

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +27,45 @@ export const TodoList: React.FC = () => (
...
+              className="button"
+              type="button"
+              onClick={() => onClickHandler(todo)}
+            >
+              <span className="icon">
+                <i
+                  className={
+                    selectedTodo?.id === todo.id
+                      ? 'far fa-eye-slash'
+                      : 'far fa-eye'
```
volodymyr-soltys97, line 61: Use the classnames library rules for add classes with condition

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
 import { Loader } from './components/Loader';
+import { getTodos, getUser } from './api';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
+import { FilterTypes } from './types/FilterTypes';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [user, setUser] = useState<User | null>(null);
```
etojeDenys, line 18: does App.tsx need to save the user state? I don't think so, you can just move all user related logic inside the TodoModal

src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+      return !todo.completed && hasQuery;
+    }
+
+    if (filterMethod === FilterTypes.completed) {
+      return todo.completed && hasQuery;
+    }
+
+    if (filterMethod === FilterTypes.all) {
+      return true && hasQuery;
+    }
```
etojeDenys, line 67: you can simplify that using the switch case

etojeDenys (changes requested):

src/App.tsx
anastasiia-tilikina: not fixed

anastasiia-tilikina (changes requested):
Almost done!

one comment is not fixed from prev review

image

src/App.tsx
OlehPhw: Isn't it?
image
Mb your are looking at old commit? Because I fixed it

SanyaBratashchuk (approved):

PR #2510, opened 2024-08-08
add solution

RainerWerner44:
[DEMO LINK](https://RainerWerner44.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,61 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos, getUser } from './api';
+import { Todo } from './types/Todo';
+import { User } from './types/User';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [showTodoId, setShowTodoId] = useState(0);
+  const [isModalActive, setIsModalActive] = useState(false);
+  const [todoOwner, setTodoOwner] = useState<User | null>(null);
+  const [selectedUsers, setSelectedUsers] = useState('all');
+  const [query, setQuery] = useState('');
+  const [modalLoading, setModalLoading] = useState(false);
+
+  const todo = todos.find(t => t.id === showTodoId) || null;
+  const findUser = () => {
+    setModalLoading(true);
+    if (todo?.userId) {
+      getUser(todo.userId)
+        .then(({ email, id, name, phone }) => {
+          setTodoOwner({ email, id, name, phone });
+        })
+        .finally(() => {
+          setModalLoading(false);
+        });
+    }
+  };
+
+  useEffect(() => {
+    findUser();
+  }, [todo]);
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(setTodos)
+      .finally(() => setLoading(false));
+  }, []);
+
+  let filteredTodos = [...todos];
+
+  if (selectedUsers === 'active') {
+    filteredTodos = filteredTodos.filter(t => !t.completed);
+  }
+
+  if (selectedUsers === 'completed') {
+    filteredTodos = filteredTodos.filter(t => t.completed);
+  }
+
+  if (query) {
+    filteredTodos = filteredTodos.filter(el =>
+      el.title.toLocaleLowerCase().includes(query.toLocaleLowerCase()),
+    );
+  }
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+import { Todo } from './types/Todo';
+import { User } from './types/User';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [showTodoId, setShowTodoId] = useState(0);
+  const [isModalActive, setIsModalActive] = useState(false);
+  const [todoOwner, setTodoOwner] = useState<User | null>(null);
+  const [selectedUsers, setSelectedUsers] = useState('all');
```
BudnikOleksii, line 20: If it's a filter type, it should be enum instead of magic value

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+  const [query, setQuery] = useState('');
+  const [modalLoading, setModalLoading] = useState(false);
```
BudnikOleksii, line 22: Local state related to modal should in the modal component

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+
+  const todo = todos.find(t => t.id === showTodoId) || null;
```
BudnikOleksii, line 24: You can store a selected todo in the state directly instead of id

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+        })
+        .finally(() => {
+          setModalLoading(false);
+        });
+    }
+  };
+
+  useEffect(() => {
+    findUser();
+  }, [todo]);
```
BudnikOleksii, line 40:
```suggestion
  const findUser = () => {
    setModalLoading(true);
    if (todo?.userId) {
      getUser(todo.userId)
        .then(setTodoOwner)
        .finally(() => {
          setModalLoading(false);
        });
    }
  };

  useEffect(() => {
    findUser();
  }, [todo]);
```
- You don't use this in this component so move it modal
- catch method is missing, in catch you should handle possible errors
- React Hook useEffect has a missing dependency: 'findUser'. Either include it or remove the dependency array

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(setTodos)
+      .finally(() => setLoading(false));
```
BudnikOleksii, line 46: catch method is missing

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+
+  if (selectedUsers === 'completed') {
+    filteredTodos = filteredTodos.filter(t => t.completed);
+  }
+
+  if (query) {
+    filteredTodos = filteredTodos.filter(el =>
+      el.title.toLocaleLowerCase().includes(query.toLocaleLowerCase()),
+    );
+  }
```
BudnikOleksii, line 63: - Create a function to filter todos
- `filter` method doesn't mutate an array so you don't need to create a copy
- Use switch case when you have fixed number of condition values, and use enum instead of magic values
- avoid one-letter variables/arguments, `el` is a bad name too

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,53 @@
...
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+type Props = {
+  setSelectedUsers: (a: string) => void;
+  query: string;
+  setQuery: (a: string) => void;
```
BudnikOleksii, line 4: What is `a`?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,53 @@
...
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={event => setSelectedUsers(event.target.value)}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
BudnikOleksii, line 22: Use enum for this select

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,53 @@
...
+          className="input"
+          placeholder="Search..."
+          value={query}
+          onChange={event => setQuery(event.target.value)}
+        />
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
BudnikOleksii, line 40: Avoid inline styles, use classes instead

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,91 @@
...
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+type Props = {
+  filteredTodos: Todo[];
+  showTodoId: number;
+  setShowTodoId: (id: number) => void;
+  setIsModalActive: (a: boolean) => void;
```
BudnikOleksii, line 9: `a`... fix naming in all places

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,91 @@
...
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {filteredTodos.map(todo => {
+          return (
+            <tr
+              data-cy="todo"
```
BudnikOleksii, line 39: Create a todo component

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,43 +1,77 @@
...
+  showTodoId,
+  isModalActive,
+  todo,
+  setIsModalActive,
+  todoOwner,
+  modalLoading,
+}) => {
+  return (
+    <>
+      {isModalActive && (
```
BudnikOleksii, line 26: You can do this check in the parent component as you render nothing in this case

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,43 +1,77 @@
...
+                  className="modal-card-title has-text-weight-medium"
+                  data-cy="modal-header"
+                >
+                  Todo #{todo?.id}
+                </div>
 
-            <p className="block" data-cy="modal-user">
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+                {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
BudnikOleksii, line 45: Don't disable linter

BudnikOleksii (changes requested):
Good start 👍 
Let's fix some moments

mrkovaliv (approved):
👍 👍 👍

PR #2520, opened 2024-08-18
solution

coldshein:
[DEMO LINK](https://coldshein.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


RimmaBuzuluk (approved):
good job ))

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+  const [showModal, setShowModal] = useState(false);
+
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [filterQuery, setFilterQuery] = useState('');
+
+  const getAllTodos = () => {
+    getTodos()
+      .then(data => setAllTodos(data))
```
BudnikOleksii, line 27:
```suggestion
      .then(setAllTodos)
```
- Could be simplified
- Don't forget to use `catch` to handle possible errors

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { debounce } from './utils';
 
 export const App: React.FC = () => {
+  const [allTodos, setAllTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  const [isLoading, setIsLoading] = useState(true);
+  const [showModal, setShowModal] = useState(false);
```
BudnikOleksii, line 19:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
- You should set isLoading only when actually start loading something
- `showModal` is redundant, you can show modal only in case of `selectedTodo`

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [filterQuery, setFilterQuery] = useState('');
+
+  const getAllTodos = () => {
+    getTodos()
+      .then(data => setAllTodos(data))
+      .finally(() => setIsLoading(false));
+  };
+
+  const applyQuery = useCallback(debounce(setAppliedQuery, 300), []);
```
BudnikOleksii, line 31: React Hook useCallback received a function whose dependencies are unknown. Pass an inline function instead

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
 export const App: React.FC = () => {
+  const [allTodos, setAllTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  const [isLoading, setIsLoading] = useState(true);
+  const [showModal, setShowModal] = useState(false);
+
+  const [query, setQuery] = useState('');
+  const [appliedQuery, setAppliedQuery] = useState('');
+  const [filterQuery, setFilterQuery] = useState('');
```
BudnikOleksii, line 23: Create an enum or a type for the filter. And probably it's not a query, but a filter type or something

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+  };
+
+  useEffect(() => {
+    getAllTodos();
+  }, []);
+
+  const getVisibleTodos = useCallback(() => {
+    let filteredTodos = [...allTodos];
+
+    if (filterQuery === 'active' || filterQuery === 'completed') {
```
BudnikOleksii, line 59:
```suggestion
    if (filterQuery) {
```
Could be simplified

src/App.tsx
```diff
@@ -1,14 +1,82 @@
...
+      filteredTodos = allTodos.filter(({ completed }) => {
+        if (filterQuery === 'active') {
+          return !completed;
+        } else {
+          return completed;
+        }
```
BudnikOleksii, line 65: This could be simplified as well, try to do it

src/components/TodoCard/TodoCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+          onClick={() => {
+            selectTodo(todo);
+            showModalWindow(true);
+          }}
+        >
+          <span className="icon">
+            <i
+              className={cn('far', {
+                'fa-eye': selectedTodo?.id !== todo.id,
+                'fa-eye-slash': selectedTodo?.id === todo.id,
```
BudnikOleksii, line 61: You compare twice, create a variable

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,42 @@
...
+      getUser(todo.userId)
+        .then(setTodoOwner)
+        .catch(error => {
+          alert(error);
+          setTodoOwner(null);
+        })
+        .finally(() => {
+          setIsLoading(false);
+        });
+    }
```
BudnikOleksii, line 28: Fetch todos flow should looks like this 👍 
- Only one moment, `useCallback` doesn't make any sense, as you don't pass this function to other components

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,42 @@
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { User } from '../../types/User';
+import { getUser } from '../../api';
+import { Todo } from '../../types/Todo';
+
+type Props = {
+  todo: Todo | null;
```
BudnikOleksii, line 8:
```suggestion
  todo: Todo;
```
You don't need to render modal without todo

BudnikOleksii (changes requested):
Looks good 👍 
Left some comments, check them and improve your solution

ArturErmolenko (approved):
Well done 🔥

PR #2525, opened 2024-08-20
Solution

AndGrachov:
[DEMO LINK](https://AndGrachov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,47 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { TodoState } from './types/TodoState';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [inputQuery, setInputQuery] = useState('');
+  const [selectQuery, setSelectQuery] = useState<string>(TodoState.ALL);
+
+  const showSelectedTodo = (todoId: number) => {
+    const receivedTodo = todos.find(todo => todo.id === todoId) || null;
+
+    setSelectedTodo(receivedTodo);
+  };
+
+  useEffect(() => {
+    getTodos().then(todosFromServer => {
+      let filteredTodos = todosFromServer;
+      const preparedInputQuery = inputQuery.trim().toLowerCase();
+
+      if (preparedInputQuery) {
+        filteredTodos = filteredTodos.filter(todo =>
+          todo.title.toLowerCase().includes(preparedInputQuery),
+        );
+      }
+
+      if (selectQuery !== TodoState.ALL) {
+        if (selectQuery === TodoState.COMPLETED) {
+          filteredTodos = filteredTodos.filter(todo => todo.completed);
+        }
+
+        if (selectQuery === TodoState.ACTIVE) {
+          filteredTodos = filteredTodos.filter(todo => !todo.completed);
+        }
+      }
+
+      setTodos(filteredTodos);
+    });
+  });
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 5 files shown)


vadimurbanovich (approved):
GJ!

src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+        }
+
+        if (selectQuery === TodoState.ACTIVE) {
+          filteredTodos = filteredTodos.filter(todo => !todo.completed);
+        }
+      }
+
+      setTodos(filteredTodos);
+    });
+  });
```
BudnikOleksii, line 49: - You should load todos only on the component first render.
- The dependencies array is missing, so this `useEffect` doesn't make any sesnse
- You don't need to filter them inside this `useEffect`, you should filter when the query or filter type change
- Create a separate helper function to filter todos, move it to utils/helpers file and use switch case instead

src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { TodoState } from './types/TodoState';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [inputQuery, setInputQuery] = useState('');
+  const [selectQuery, setSelectQuery] = useState<string>(TodoState.ALL);
```
BudnikOleksii, line 18: It's not a string, it should be enum type

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,11 +1,29 @@
-export const TodoFilter = () => (
+import { TodoState } from '../../types/TodoState';
+
+type Props = {
+  inputQuery: string;
+  setInputQuery: (query: string) => void;
+  selectQuery: string;
+  setSelectQuery: (query: string) => void;
```
BudnikOleksii, line 7: You should accept only valid values from your enum, it's not a string

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,15 +34,24 @@ export const TodoFilter = () => (
...
         <i className="fas fa-magnifying-glass" />
       </span>
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
+      {inputQuery && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
BudnikOleksii, line 46: - Avoid inline styles, use classes instead
- Don't disable linter

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,52 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+        return (
+          <tr key={todo.id} data-cy="todo" className="">
```
BudnikOleksii, line 36: Create a todo component

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,31 @@
...
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {!user ? (
         <Loader />
       ) : (
```
BudnikOleksii, line 30: The same, you should show loader only in case you actually load something

BudnikOleksii (changes requested):
Good work 👍 
Check the comments and improve your solution

src/App.tsx
```diff
@@ -1,14 +1,33 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+  const [inputQuery, setInputQuery] = useState('');
+  const [selectQuery, setSelectQuery] = useState<TodoState>(TodoState.ALL);
+
+  const preparedTodos = getPreparedTodos(todos, { inputQuery, selectQuery });
+
+  useEffect(() => {
+    getTodos().then(todosFromServer => {
+      setTodos(todosFromServer);
+      setIsLoading(false);
```
BudnikOleksii, line 27: You should remove loader even the request fails, so do it in the `finally` block

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  selectedTodo,
+  setSelectedTodo,
+}) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    getUser(selectedTodo.userId).then(userFromServer => {
+      setUser(userFromServer);
+      setIsLoading(false);
```
BudnikOleksii, line 22: The same

src/utils/todos.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  if (selectQuery !== TodoState.ALL) {
+    if (selectQuery === TodoState.COMPLETED) {
+      filteredTodos = filteredTodos.filter(todo => todo.completed);
+    }
+
+    if (selectQuery === TodoState.ACTIVE) {
+      filteredTodos = filteredTodos.filter(todo => !todo.completed);
+    }
```
BudnikOleksii, line 24: It's better to use a switch case

BudnikOleksii (approved):
Looks good 👍 
Approved, just left a few suggestions for the future

PR #2527, opened 2024-08-20
Solution;-)

YaroslavKolesnyk:
[DEMO LINK](https://YaroslavKolesnyk.github.io/react_dynamic-list-of-todos/)


piyopiyo0 (approved):
Great job! 🔥

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+      status === TodoStatus.completed ? todo.completed : !todo.completed,
+    );
+  }
+
+  return filterTodos;
+};
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(true);
```
BudnikOleksii, line 39: Boolean variables should start from `is` or `has`. And set loading state only when you start loading something

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+  const [query, setQuery] = useState('');
+  const [status, setStatus] = useState(TodoStatus.all);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos()
+      .then(setTodos)
+      .catch(() => {
+        return (
+          <p style={{ color: 'red' }}>Something went wrong, try again later</p>
```
BudnikOleksii, line 49: - It doesn't work this way, you can get this paragraph in the next `then` block only
- You can set an error state and depending on it render an error message
- Avoid inline styles, use classes instead
- Error message should be informative

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,76 @@
...
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setQuery(inputValue);
+    }, 1000);
+
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [inputValue, setQuery]);
```
BudnikOleksii, line 27: Use debounce function from some library or create it yourself instead

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,76 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={status}
+            onChange={handleStatusChange}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
BudnikOleksii, line 44: Use `TodoStatus` enum

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,82 @@
...
+          </th>
+          <th>Title</th>
+          <th> </th>
+        </tr>
+      </thead>
+      <tbody>
+        {todos.map(todo => (
+          <tr
+            key={todo.id}
+            data-cy="todo"
```
BudnikOleksii, line 38: Create a todo component

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,82 @@
...
+            className={classNames({
+              'has-background-info-light': todo.id % 2 === 0,
```
BudnikOleksii, line 40: Check the working version, you need highlight only the row that is selected

BudnikOleksii (changes requested):
Good work 👍 

BudnikOleksii (approved):
👍

PR #2535, opened 2024-08-21
*****

NIKaragu:
[DEMO LINK](https://nikaragu.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,54 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos, getUser } from './api';
+import { User } from './types/User';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedUser, setSelectedUser] = useState<User | null>(null);
+  const [initialTodos, setInitialTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos().then(todosList => {
+      setTodos(todosList);
+      setInitialTodos(todosList);
+    });
+  }, []);
+
+  const handleEyeClick = useCallback(
+    (selectedUserId: number, selectedTodoId: number) => {
+      getUser(selectedUserId).then(setSelectedUser);
+      setSelectedTodo(todos.find(todo => todo.id === selectedTodoId) || null);
+    },
+    [todos],
+  );
+
+  const handleFiltrationQueries = useCallback(
+    (finishQuery = 'all', searchQuery = '') => {
+      setTodos(
+        initialTodos.filter(todo => {
+          if (finishQuery === 'all') {
+            return todo.title.toLowerCase().includes(searchQuery.toLowerCase());
+          }
+
+          return (
+            todo.title.toLowerCase().includes(searchQuery.toLowerCase()) &&
+            todo.completed === (finishQuery === 'active' ? false : true)
+          );
+        }),
+      );
+    },
+    [initialTodos],
+  );
+
+  const handleCancelSelection = useCallback(() => {
+    setSelectedUser(null);
+    setSelectedTodo(null);
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -17,18 +63,28 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter handleFiltrationQueries={handleFiltrationQueries} />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0 && <Loader />}
+              <TodoList
+                todos={todos}
+                handleEyeClick={handleEyeClick}
+                selectedTodoId={selectedTodo?.id ?? null}
+              />
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedUser && (
+        <TodoModal
+          selectedUser={selectedUser}
+          userTodo={selectedTodo}
+          resetSelection={handleCancelSelection}
+        />
+      )}
     </>
   );
 };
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [initialTodos, setInitialTodos] = useState<Todo[]>([]);
```
BudnikOleksii, line 15: Why do you have two states for todos? There should be one source of truth

src/App.tsx
```diff
@@ -1,14 +1,56 @@
...
+            return todo.title.toLowerCase().includes(searchQuery.toLowerCase());
+          }
+
+          return (
+            todo.title.toLowerCase().includes(searchQuery.toLowerCase()) &&
+            todo.completed === (finishQuery === 'active' ? false : true)
+          );
+        }),
+      );
+    },
```
BudnikOleksii, line 46: - Filtered todos is a dynamic value that depends on selected filters so don't store it in the state, just calculate
- Create a helper function to filter and move it to utils/helpers file/folder
- Get rid of magic values such as "all"/"active"

Simple example:
```typescript
const filteredTodos = filterTodos(todos, { query, status });
```

src/App.tsx
```diff
@@ -17,18 +59,27 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter handleFiltrationQueries={handleFiltrationQueries} />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0 && <Loader />}
```
BudnikOleksii, line 66: You should show the loader only if you are loading something. If the API returns an empty array the user sees an infinite loader

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,64 @@
...
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
-    </p>
-  </form>
-);
+export const TodoFilter: React.FC<Props> = ({ handleFiltrationQueries }) => {
+  const [finishQuery, setFinishQuery] = useState('all');
```
BudnikOleksii, line 8: It's not a string, it should have only a few valid values

src/components/TodoList/TodoList.tsx
```diff
@@ -14,87 +26,48 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr data-cy="todo" className="" key={todo.id}>
```
BudnikOleksii, line 31: Create a todo component

src/components/TodoList/TodoList.tsx
```diff
@@ -14,87 +26,48 @@ export const TodoList: React.FC = () => (
...
+              className="button"
+              type="button"
+              onClick={() => {
+                handleEyeClick(todo.id);
+              }}
+            >
+              <span className="icon">
+                <i
+                  className={classNames('far', {
+                    'fa-eye': selectedTodoId !== todo.id,
```
BudnikOleksii, line 62: You compare ids twice, create a variable

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,30 @@
...
+
+export const TodoModal: React.FC<Props> = ({ userTodo, resetSelection }) => {
+  const [isLoaderShowing, setIsLoaderShowing] = useState(true);
+  const [selectedUser, setSelectedUser] = useState<User | null>(null);
+
+  useEffect(() => {
+    getUser(userTodo?.userId || 0).then(user => {
+      setSelectedUser(user);
+      setIsLoaderShowing(false);
+    });
```
BudnikOleksii, line 20:
```suggestion
  const [isLoaderShowing, setIsLoaderShowing] = useState(false);
  const [selectedUser, setSelectedUser] = useState<User | null>(null);

  useEffect(() => {
    setIsLoaderShowing(true);
    
    getUser(userTodo?.userId || 0)
      .then(setSelectedUser)
      .catch(error => // handler errors)
      .finally(() => {
        setIsLoaderShowing(false);
      });
```
You should show the loader not by default, but only in case you actually loading something
And as I mentioned before you should show the loader when you loading tods as well

BudnikOleksii (changes requested):
Good work 👍 
Check the comments and improve your solution

src/App.tsx
NIKaragu: 'cause we haven't to change todos state, otherwise after every filtration by query we will reduce our list of todos and in one moment no todo will be in todos list

src/components/TodoFilter/TodoFilter.tsx
NIKaragu: image

it's completely string

src/App.tsx
```diff
@@ -1,14 +1,77 @@
...
+
+  const handleFiltrationQueries = useCallback(
+    ({ finishQuery, searchQuery }: Query) => {
+      setTodos(
+        filterTodosByQuery(initialTodos.current, { finishQuery, searchQuery }),
+      );
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [initialTodos.current],
+  );
```
BudnikOleksii, line 69: As I said previously you don't need to store calculated values in the state, just in a variable. In state store all todos what you get from the API
Simple example:
```typescript
const filteredTodos = filterTodos(todos, { query, status });
```

src/components/Todo/TodoPost.tsx
```diff
@@ -0,0 +1,57 @@
...
+          type="button"
+          onClick={() => {
+            handleEyeClick(todo.id);
+          }}
+        >
+          <span className="icon">
+            <i
+              className={classNames('far', {
+                'fa-eye': selectedTodoId !== todo.id,
+                'fa-eye-slash': selectedTodoId === todo.id,
```
BudnikOleksii, line 49: You compare ids twice, create a variable

BudnikOleksii (approved):
Good work 👍 
Approved, but left a comment regarding two states for todos

src/App.tsx
NIKaragu: @BudnikOleksii 

okey look:

I receive `query `and `status` from handleFiltrationQueries, when user uses filtration fields, so fo relaying these parameters I'll have to create 2 new states in `App component` relay it to `TodoFilter component`, lift up from it and then relay `query `and `status` to `filterTodos`.
It looks not such flexible as I made it in my way. Moreover `filteredTodos `(in you suggestion) will be recreated every time and `filterTodos `will be recalculated every time, even if I don't change filtration fields.

Maybe I just haven't understood your solution. If that's the case, correct me.

PR #2592, opened 2024-10-10
add task solution

dokrod:
[DEMO LINK](https://dokrod.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,11 +19,11 @@ export const App: React.FC = () => {
   const [loading, setLoading] = useState<boolean>(false);
   const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
 
-  const onFilterChange = (event: ChangeEvent<HTMLSelectElement>) => {
+  const handleFilterChange = (event: ChangeEvent<HTMLSelectElement>) => {
     setFilter(event.target.value as FilterType);
   };
 
-  const onQueryChange = (event: ChangeEvent<HTMLInputElement>) => {
+  const handleQueryChange = (event: ChangeEvent<HTMLInputElement>) => {
     setQuery(event.target.value);
   };
 
@@ -31,11 +31,11 @@ export const App: React.FC = () => {
     setQuery('');
   };
 
-  const onTodoSelect = (todo: Todo) => {
+  const handleTodoSelect = (todo: Todo) => {
     setSelectedTodo(todo);
   };
 
-  const onSelectedTodoClear = () => {
+  const handleSelectedTodoClear = () => {
     setSelectedTodo(null);
   };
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -17,18 +67,30 @@ export const App: React.FC = () => {
...
+                setQuery={handleQueryChange}
+                cleaner={queryCleaner}
+              />
             </div>
 
+            {loading && <Loader />}
             <div className="block">
-              <Loader />
-              <TodoList />
+              <TodoList
```
Ann-Maksay:
```suggestion
              loading  ? <Loader /> : <TodoList />
```

It might not be worth rendering the list at all while the data is being loaded

src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
+
+    return filtered;
+  }, [filter, query, todos]);
+
+  useEffect(() => {
+    setLoading(true);
+
+    getTodos()
+      .then(setTodos)
+      .then(() => setLoading(false));
```
Ann-Maksay:
```suggestion
      .finally(() => setLoading(false));
```

It's better to use finally in this case.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,73 @@
...
+            <td className="is-vcentered">
+              {todo.completed && (
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              )}
+            </td>
+            <td className="is-vcentered is-expanded">
+              <p
+                className={
```
Ann-Maksay: You used classnames further down, why not use that here too? In general it's always a good idea if a class needs to be specified by a condition

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,73 @@
...
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map((todo: Todo) => (
+          <tr data-cy="todo" className="" key={todo.id}>
```
Ann-Maksay, line 32: We have to specify a class with a background for the selected todo

src/utils/getFilteredTodos.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Todo } from '../types/Todo';
+
+export const getFilteredTodos = (todos: Todo[], filterBy: FilterType) => {
+  if (filterBy === FilterType.completed) {
+    return todos.filter(todo => todo.completed === true);
+  } else if (filterBy === FilterType.active) {
+    return todos.filter(todo => todo.completed === false);
+  } else {
+    return todos;
+  }
```
Ann-Maksay, line 12: Seems like an ideal situation to use` switch` instead of `if else`

Ann-Maksay (approved):
Great job 💫

PR #2594, opened 2024-10-11
solution

YuliiaShatkovska:
[DEMO LINK](https://YuliiaShatkovska.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,27 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { Filters } from './types/Filters';
+import { getFilteredTodos } from './utils/getFilteredTodos';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [query, setQuery] = useState('');
+  const [selectedFilter, setSelectedFilter] = useState(Filters.ALL);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(setTodos)
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const visibleTodos = getFilteredTodos(todos, { query, selectedFilter });
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,33 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [query, setQuery] = useState('');
+  const [selectedFilter, setSelectedFilter] = useState(Filters.ALL);
+
+  useEffect(() => {
+    setIsLoading(true);
```
hma-3, line 23: you don't have to set loader to `true` here, just change the initial value in the state
```ts
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,66 @@
...
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value.trimStart());
+  };
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select data-cy="statusSelect" onChange={handleSelectFilter}>
```
hma-3: add `value` to `select`

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +27,46 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+        return (
+          <tr data-cy="todo" className="" key={id}>
```
hma-3: row needs to have class `'has-background-info-light'` when selected

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +27,46 @@ export const TodoList: React.FC = () => (
...
-      </tr>
+            {completed ? (
+              <td className="is-vcentered">
+                <span className="icon" data-cy="iconCompleted">
+                  <i className="fas fa-check" />
+                </span>
+              </td>
+            ) : (
+              <td className="is-vcentered" />
+            )}
```
hma-3: move condition inside `td` and then you won't have to write it twice

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,39 @@
...
+  selectedTodo,
+  setSelectedTodo,
+}) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [user, setUser] = useState<User | null>(null);
+
+  const { id, title, completed, userId } = selectedTodo;
+
+  useEffect(() => {
+    setIsLoading(true);
```
hma-3, line 22: > you don't have to set loader to true here, just change the initial value in the state
`const [isLoading, setIsLoading] = useState(true);`

same

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,25 +42,34 @@ export const TodoModal: React.FC = () => {
...
             </p>
 
             <p className="block" data-cy="modal-user">
               {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              {completed ? (
+                <strong className="has-text-success">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
```
hma-3, line 68: I have mixed feelings about this😅
it may be worth moving the condition inside `strong` to avoid duplication, but I think this is a moot point here

src/utils/getFilteredTodos.ts
```diff
@@ -0,0 +1,39 @@
...
+  filteredTodos = filteredTodos.filter(todo => {
+    switch (selectedFilter) {
+      case Filters.ACTIVE:
+        return !todo.completed;
+
+      case Filters.COMPLETED:
+        return todo.completed;
+
+      default:
+        return todo;
```
hma-3: filter callback should return boolean value. here it's still working, but it would be better to return just `true`

hma-3 (approved):
Good job!
Have just a few suggestions⬇️

PR #2595, opened 2024-10-11
Still in progress

Ostkreuzzz:
https://Ostkreuzzz.github.io/react_dynamic-list-of-todos/


src/App.tsx
```diff
@@ -8,7 +8,46 @@ import { TodoFilter } from './components/TodoFilter';
...
+        break;
+      case FILTER_STATUS.COMPLETED:
+        setIsLoading(true);
+
+        getCompleted()
+          .then(result => setTodos(handleQuery(result, query)))
+          .finally(() => setIsLoading(false));
+        break;
+    }
+  }, [filterBy, query]);
```
YuriiMykhailenko: You should make one get request here with an empty array of dependencies because in this way you did you make it each time you change state of app
```suggestion
  useEffect(() => {
    //set is loading true
    //make one request here 
    //set isloading false
    }
  }, []);
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,31 @@
...
+export const TodoModal: React.FC<Props> = ({ currentTodo, onCurrentTodo }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getUser(currentTodo.userId)
+      .then(setUser)
+      .finally(() => setIsLoading(false));
+  }, [currentTodo.userId]);
```
YuriiMykhailenko, line 22: you need to make something similar in app.tsx

src/utils/FILTER_STATUS.ts
```diff
@@ -0,0 +1,5 @@
+export const FILTER_STATUS = {
+  ALL: 'all',
+  ACTIVE: 'active',
+  COMPLETED: 'completed',
+};
```
YuriiMykhailenko, line 5: Use enums or types instead of this

YuriiMykhailenko (changes requested):
You did good job👍  Just consider my comments below and you don't need to extract functions in utils unless you use them in different parts of your app

PR #2596, opened 2024-10-11
Develop

Svidruk:
[DEMO LINK](Svidruk.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+
+      const matchesQuery = todo.title
+        .toLocaleLowerCase()
+        .includes(query.toLocaleLowerCase());
+
+      return matchesFilter && matchesQuery;
+    });
+
+    return newTodos;
+  })();
```
Hradivnyk, line 49: You can create folder "utils" and move this function there

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,59 @@
...
+      </p>
+
+      <p className="control is-expanded has-icons-left has-icons-right">
+        <input
+          data-cy="searchInput"
+          type="text"
+          className="input"
+          placeholder="Search..."
+          value={query}
+          onChange={event => setQuery(event.target.value.trimStart())}
```
Hradivnyk, line 39: You don't need trim this. In example, we can write with space, and this change result

Hradivnyk (approved):
Well done!

PR #2603, opened 2024-10-13
solution

EustasTheMonk:
DEMO LINK (https://eustasthemonk.github.io/react_dynamic-list-of-todos/)


PikalovAndrey (changes requested):
Looks good, but you have a bug that doesn't match the original working version. Once the filter is selected and you have characters in the input, after clicking the Clear button your filter should not become "All". Just check the working version.

PR #2606, opened 2024-10-14
Add task solution

WatarJoy:
https://WatarJoy.github.io/react_dynamic-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,17 +7,14 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
-import { getTodos, getUser } from './api';
+import { getTodos } from './api';
 import { Todo } from './types/Todo';
-import { User } from './types/User';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
   const [isLoadingTodos, setIsLoadingTodos] = useState(false);
-  const [isLoadingUser, setIsLoadingUser] = useState(false);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,71 @@
...
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [isLoadingTodos, setIsLoadingTodos] = useState(false);
+  const [isLoadingUser, setIsLoadingUser] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [userDetails, setUserDetails] = useState<User | null>(null);
+  const [filterStatus, setFilterStatus] = useState('all');
+  const [query, setQuery] = useState('');
```
mbulchak, line 19: Suggestion: states userDetails and isLoadingUser are needed only in TodoModal. You can create useEffect for them as you do for todos

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +27,49 @@ export const TodoList: React.FC = () => (
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          key={todo.id}
+          data-cy="todo"
+          className={todo.completed ? 'has-background-info-light' : ''}
```
mbulchak: This is causing a problem. You don't need to highlight completed todos. You should highlight when it is your selectedTodo. But in your case it SelectedTodoId

mbulchak (changes requested):
Good job. Just fix some moments.
Check your demo, your todos are highlighted at the beggining. Change it for SelectedTodoId, instead of todo.completed

PR #2627, opened 2024-11-03
Solution

ab3MN:
[DEMO LINK](https://ab3MN.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,15 +10,14 @@ import { Loader } from './components/Loader';
 import { ITodoFilter, Todo } from './types/Todo';
 import { getTodos } from './api';
 import { getFiltredTodo } from './utils/filterTodo';
+import { DEFAULT_FILTER_STATE } from './constants/TodoFilter';
 
 export const App: React.FC = () => {
   const [isLoading, setLoading] = useState(true);
   const [todos, setTodos] = useState<Todo[]>([]);
   const [selectedTodo, setselectedTodo] = useState<null | Todo>(null);
-  const [todosFilter, setTodosFilter] = useState<ITodoFilter>({
-    status: '',
-    filter: '',
-  });
+  const [todosFilter, setTodosFilter] =
+    useState<ITodoFilter>(DEFAULT_FILTER_STATE);
 
   const filtredTodos = useMemo(
     () => getFiltredTodo(todos, todosFilter),
@@ -35,7 +34,7 @@ export const App: React.FC = () => {
         // eslint-disable-next-line no-console
         console.log(e.message);
       });
-  }, [getTodos]);
+  }, []);
 
   const closeModal = () => setselectedTodo(null);
 
@@ -47,7 +46,10 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter setTodosFilter={setTodosFilter} />
+              <TodoFilter
+                setTodosFilter={setTodosFilter}
+                todosFilter={todosFilter}
+              />
             </div>
 
             <div className="block">
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useLayoutEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,14 +1,44 @@
...
+import { ITodoFilter, Todo } from './types/Todo';
+import { getTodos } from './api';
+import { getFiltredTodo } from './utils/filterTodo';
 
 export const App: React.FC = () => {
+  const [isLoading, setLoading] = useState(true);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [selectedTodo, setselectedTodo] = useState<null | Todo>(null);
+  const [todosFilter, setTodosFilter] = useState<ITodoFilter>({
+    status: '',
```
VolodymyrKirichenko: u can create `enum` for ur statuses

```suggestion
    status: '',
```
```
export enum Filter_Statuses {
  All = 'all',
  Active = 'active',
  Completed = 'completed',
}
```

src/utils/filterTodo.ts
```diff
@@ -0,0 +1,23 @@
+import { TODO_FILTER_STATUS } from '../constants/TodoFilter';
+import { ITodoFilter, Todo } from '../types/Todo';
+
+export const getTodoFilterStatus = (status = '') => {
+  if (status in TODO_FILTER_STATUS) {
+    return TODO_FILTER_STATUS[status as keyof typeof TODO_FILTER_STATUS];
+  }
+  return undefined;
```
VolodymyrKirichenko: do u need it?
I want to suggest a better option under this comment
with the help of the following u will avoid `undefined ` value

src/utils/filterTodo.ts
```diff
@@ -0,0 +1,23 @@
...
+};
+
+export const getFiltredTodo = (
+  todos: Todo[],
+  { status, filter }: ITodoFilter,
+) => {
+  const todoStatus = getTodoFilterStatus(status);
```
VolodymyrKirichenko:
```suggestion
   let filteredTodos = [...todos];

  filteredTodos = filteredTodos.filter(({ completed }) => {
    switch (status) {
      case Filter_Statuses.Active:
        return !completed;
      case Filter_Statuses.Completed:
        return completed;
      default:
        return true;
    }
  });
```
under this u can do an additional check on the query

src/utils/filterTodo.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  return todos.filter(todo =>
+    todoStatus === undefined
+      ? todo.title.toLowerCase().includes(filter)
```
VolodymyrKirichenko: add `trim` method to ignore spaces

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,69 @@
...
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            value={status}
+            onChange={handleChangeStatus}
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
VolodymyrKirichenko: use `map` method for it

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,69 @@
...
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
-    </p>
-  </form>
-);
+export const TodoFilter: FC<IProps> = ({ setTodosFilter }) => {
+  const [status, setStatus] = useState('');
+  const [filter, setFilter] = useState('');
```
VolodymyrKirichenko: for what?
u already have these states in the `App` component. u can pass it as a prop

VolodymyrKirichenko (changes requested):
Good job, but review my comments 
Please fix it or provide comments 🥹

src/utils/filterTodo.ts
ab3MN: yes,but i need to use two cycle,i thought that use one cycle is less expensive

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,76 @@
...
+            onChange={handleChangeStatus}
+          >
+            {TODO_FILTER_OPTIONS.map(({ value, title, id }) => (
+              <option value={value} key={id}>
+                {title}
+              </option>
+            ))}
+          </select>
+        </span>
+      </p>
```
VenomMB, line 49: add empty line

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +28,51 @@ export const TodoList: React.FC = () => (
...
+                {title}
+              </p>
+            </td>
+            <td className="has-text-right is-vcentered">
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => setSelectedUser(todo)}
+              >
```
VenomMB, line 60: add empty line between blocks

src/utils/filterTodo.ts
```diff
@@ -0,0 +1,24 @@
...
+  filteredTodos = filteredTodos.filter(({ completed }) => {
+    switch (status) {
+      case FilterStatuses.Active:
+        return !completed;
+      case FilterStatuses.Completed:
+        return completed;
+      default:
+        return true;
+    }
+  });
```
VenomMB, line 21: I think you can combine it into one filter, and split it into two functions in the middle.
like: 

```suggestion
  
  const isTitleMatchedToFilter = (title) => title.toLowerCase().trim().includes(filter);
  
  const filterByStatus = (status, completed) => {
    switch (status) {
      case FilterStatuses.Active:
        return !completed;
      case FilterStatuses.Completed:
        return completed;
      default:
        return true;
    }
  }

  return todos.filter(({ completed, title }) => {
    if(!isTitleMatchedFilter(title) {
      return;
    }
    
    return filterByStatus(status, completed); 
   
  });
```

VenomMB (approved):
Greate job!!! Leave small comments)

src/utils/filterTodo.ts
ab3MN: I understand,so if i have 10 filters it would be better to create more function to check and for filtered than create a huge ternary operate or many if in filter

PR #2642, opened 2024-11-18
solution

Krykunov:
[DEMO LINK](https://Krykunov.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
@@ -14,23 +14,11 @@ import { getTodos } from './api';
 export const App: React.FC = () => {
   const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
   const [todos, setTodos] = useState<Todo[]>([]);
-  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  // const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
   const [search, setSearch] = useState<string>('');
   const [status, setStatus] = useState<string>('all');
   const [errorMessage, setErrorMessage] = useState<string>('');
   const [updatedAt, setUpdatedAt] = useState(new Date());
-
-  const onSearch = (value: string) => {
-    setSearch(value);
-  };
-
-  const onClearSearch = () => {
-    setSearch('');
-  };
-
-  const onStatusChange = (value: string) => {
-    setStatus(value);
-  };
 
   useEffect(() => {
     getTodos()
@@ -40,8 +28,8 @@ export const App: React.FC = () => {
       });
   }, [updatedAt]);
 
-  useEffect(() => {
-    const filtered = todos.filter(todo => {
+  const filteredTodos = useMemo(() => {
+    return todos.filter(todo => {
       if (status === 'all') {
         return todo.title.toLowerCase().includes(search.toLowerCase());
       }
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
 export const App: React.FC = () => {
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
+  const [search, setSearch] = useState<string>('');
+  const [status, setStatus] = useState<string>('all');
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const [updatedAt, setUpdatedAt] = useState(new Date());
+
+  const onSearch = (value: string) => {
```
ab3MN: There is no need to create this function,u can use directly setSearch

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+    setSearch(value);
+  };
+
+  const onClearSearch = () => {
```
ab3MN: Also no need to create this function,u can use directly setSearch

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+    setSearch('');
+  };
+
+  const onStatusChange = (value: string) => {
+    setStatus(value);
```
ab3MN: Also no need to create this function,this function do nothing,u can use directly setStatus

src/App.tsx
```diff
@@ -17,18 +72,38 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                onSearch={onSearch}
+                onStatusChange={onStatusChange}
+                onClearSearch={onClearSearch}
```
ab3MN: u can set props directly setSearch and setStatus

src/App.tsx
```diff
@@ -17,18 +72,38 @@ export const App: React.FC = () => {
...
+                />
+              )}
+              {errorMessage && todos.length === 0 && (
+                <div>
+                  <p>{errorMessage}</p>
+                  <button className="button is-danger" onClick={reload}>
+                    Reload
+                  </button>
+                </div>
+              )}
```
ab3MN: Will be better to create function like getTodosListView and do all conditions there

src/App.tsx
```diff
@@ -17,18 +72,38 @@ export const App: React.FC = () => {
...
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal todo={selectedTodo} onClose={() => handleSelectTodo(null)} />
+      )}
```
ab3MN: Why u dont set props directly setSelectedTodo?

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+        todo.title.toLowerCase().includes(search.toLowerCase()) &&
+        todo.completed === (status === 'completed')
+      );
+    });
+
+    setFilteredTodos(filtered);
+  }, [todos, search, status]);
+
+  const handleSelectTodo = (todo: Todo | null) => {
+    setSelectedTodo(todo);
```
ab3MN: There is no need to create this function only for setting selectedTodo,u have setSelectedTodo for it

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>([]);
```
ab3MN: there is no need to create state for filtredTodos, u can use useMemo

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+    const filtered = todos.filter(todo => {
+      if (status === 'all') {
+        return todo.title.toLowerCase().includes(search.toLowerCase());
+      }
+
+      return (
+        todo.title.toLowerCase().includes(search.toLowerCase()) &&
+        todo.completed === (status === 'completed')
+      );
+    });
```
ab3MN: u can use useMemo for filterdTodos

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,84 @@
...
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map(todo => (
+          <tr
```
ab3MN: It would be better to create separate component for it.like TodoItem

ab3MN (changes requested):
Good work👍, but need to fix a few comments

PR #2656, opened 2024-12-09
Solution

OkMoroz:
- [DEMO LINK](https://OkMoroz.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,6 +1,17 @@
 import React from 'react';
+import { Todo } from '../../types/Todo';
 
-export const TodoList: React.FC = () => (
+type Props = {
+  todos: Todo[];
+  selectedTodo: Todo | null;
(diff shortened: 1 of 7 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +27,51 @@ export const TodoList: React.FC = () => (
...
-        <td className="has-text-right is-vcentered">
-          <button data-cy="selectButton" className="button" type="button">
-            <span className="icon">
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+            <td className="is-vcentered is-expanded">
+              <p className={completed ? 'has-text-success' : 'has-text-danger'}>
```
StasSokolov1: will be better to use classnames library in this case but it's not critical just my recomendation

StasSokolov1 (approved):
Great work, it looks really good to me🔥

PR #2661, opened 2024-12-10
Solution added

CamilleHaus:
[DEMO LINK](https://CamilleHaus.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,30 +1,79 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
-    <p className="control">
-      <span className="select">
-        <select data-cy="statusSelect">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+import { ChangeEvent, useEffect, useState } from 'react';
+import { useTodo } from '../../context/TodoContext';
 
-    <p className="control is-expanded has-icons-left has-icons-right">
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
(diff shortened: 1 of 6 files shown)


src/context/TodoContext.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const [modalOpen, setModalOpen] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo>();
+  const [filteredTodos, setFilteredTodos] = useState<Todo[]>(todos);
+
+  const [users, setUsers] = useState<User[]>([]);
+
+  useEffect(() => {
+    const fetchTodos = async () => {
+      setLoading(true);
+      const allTodos = await getTodos();
```
joaorpereira, line 46: add try catch block here:

```
setLoading(true);
try {
const allTodos = await getTodos();
setTodos(allTodos);
} catch {

} finally {
setLoading(false);
}

```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,79 @@
...
-      </span>
-    </p>
-  </form>
-);
+import { ChangeEvent, useEffect, useState } from 'react';
+import { useTodo } from '../../context/TodoContext';
+
+export const TodoFilter = () => {
+  const {  todos, setFilteredTodos } = useTodo();
+  const [filter, setFilter] = useState<'all' | 'active' | 'completed'>('all');
```
joaorpereira, line 6: you can create a enum type:

```
enum FilterState  {
ALL = "all"
ACTIVE = "active"
COMPLETED = "completed"
}
```

and use like `FilterState.ADD`

joaorpereira (approved):

joaorpereira:
@CamilleHaus just another thing that I noted is that when I click to show an ITEM, on the eye icon, the table is loading.

For UI perspective this is not required, I believe that will be better if the table maintains the same while the modal content is loading.

Greate work, your code is very clean and when done. Continue on this path!

PR #2678, opened 2024-12-12
solution

Fridrif27:
[DEMO LINK](https://fridrif27.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


schyrva (approved):
Good Job! 👍 

You can try to use `classnames` for cleaner conditional rendering :)

src/App.tsx
```diff
@@ -1,34 +1,100 @@
...
+                filterOptions={filters}
+                onFilterChange={updateFilters}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todos.length === 0 ? (
+                <Loader />
```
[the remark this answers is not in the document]
BudnikOleksii: > The condition `todos.length === 0` is used to determine when to show the `Loader`. However, this will show the loader even if the todos have been fetched but the list is empty. Consider using a separate state to track the loading status, such as `isLoading`, to ensure the loader is only shown while data is being fetched.

Good point
If you get an empty array from the API, a user will see an infinite loader. You should show the loader only when you actually load something

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,68 @@
...
+          className="input"
+          placeholder="Search..."
+          value={filterOptions.searchByText}
+          onChange={handleSearchChange}
+        />
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+        {filterOptions.searchByText && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
BudnikOleksii, line 56: Avoid inline styles, use classes instead

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,76 @@
...
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+          return (
+            <tr
+              data-cy="todo"
+              className={isActive ? 'has-background-info-light' : ''}
```
BudnikOleksii, line 44: To set classes conditionally use [classnames](https://www.npmjs.com/package/classnames) or [clsx](https://www.npmjs.com/package/clsx) library

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,76 @@
...
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+          return (
+            <tr
+              data-cy="todo"
```
BudnikOleksii, line 43: Create a todo component

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,44 @@
...
+        .then(setFetchedUser)
+        .catch(err => {
+          setError(`Error fetching user: ${err.message}`);
+          setFetchedUser(null);
+        });
+    }
+  }, [todo]);
+
+  const handleCloseModal = () => {
+    setModalOpen(false);
```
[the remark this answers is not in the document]
BudnikOleksii: > The `setModalOpen` function is expected to take a boolean value, but the current implementation passes `false`. Ensure that the function signature and usage are consistent across the application. If `setModalOpen` is intended to close the modal, consider renaming it to `closeModal` or similar for clarity.

Good point

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,44 @@
...
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {fetchedUser === null ? (
         <Loader />
       ) : (
```
BudnikOleksii, line 43: The same thing with the loader, fix it

BudnikOleksii (changes requested):
Good work 👍

PR #2681, opened 2024-12-13
task solution

AsyaDev14:
[demo](https://asyadev14.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,57 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
-import { TodoFilter } from './components/TodoFilter';
+import { STATUSES, TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<string>(STATUSES.ALL);
+  const [query, setQuery] = useState<string>('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  // const [isLoading, setIsLoading] = useState(true);
+console.log('selectedTodo', selectedTodo);
+
+  useEffect(() => {
+    getTodos().then(res => setTodoList(res));
+  }, []);
+
+  const filteredTodoListByStatus = (): Todo[] => {
+    switch (status) {
+      case STATUSES.COMPLETED:
+        return todoList.filter(todo => todo.completed);
+
+      case STATUSES.ACTIVE:
+        return todoList.filter(todo => !todo.completed);
+
+      case STATUSES.ALL:
+      default:
+        return todoList;
+    }
+  };
+
+  const getTodoList = () => {
+    const list = filteredTodoListByStatus();
+    if (query) {
+      return list.filter((item) => item.title.toLocaleLowerCase().includes(query.toLocaleLowerCase()));
+    } else {
+      return list;
+    }
+  };
+
+  const handleSelectTodo = (todo: Todo) => {
+    setSelectedTodo(todo);
+  };
+
+  const closeModal = () => {
+    setSelectedTodo(null);
+  };
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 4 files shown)


save-v (approved):
looks good!

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
-import { TodoFilter } from './components/TodoFilter';
+import { STATUSES, TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const [todoList, setTodoList] = useState<Todo[]>([]);
+  const [status, setStatus] = useState<string>(STATUSES.ALL);
```
BudnikOleksii, line 14:
```suggestion
  const [status, setStatus] = useState(STATUSES.ALL);
```
It's not a string, it should have the corresponding type

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+  const [query, setQuery] = useState<string>('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    getTodos().then(res => setTodoList(res));
+  }, []);
+
+  const filteredTodoListByStatus = (): Todo[] => {
```
BudnikOleksii, line 22: Move this function outside the component

src/App.tsx
```diff
@@ -1,14 +1,58 @@
...
+  const getTodoList = () => {
+    const list = filteredTodoListByStatus();
+
+    if (query) {
+      return list.filter(item =>
+        item.title.toLocaleLowerCase().includes(query.toLocaleLowerCase()),
+      );
+    } else {
+      return list;
+    }
```
BudnikOleksii, line 45:
```suggestion
      return list.filter(item =>
        item.title.toLocaleLowerCase().includes(query.toLocaleLowerCase()),
      );
    }
    
    return list;
```
- Move this function as well
- You don't need `else`

src/App.tsx
```diff
@@ -17,18 +61,29 @@ export const App: React.FC = () => {
...
+                setStatus={setStatus}
+                query={query}
+                setQuery={setQuery}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todoList.length ? (
```
BudnikOleksii, line 72: If you get an empty array from the API, a user will see an infinite loader. You should show loader only when you actually load something

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,61 @@
...
-          <option value="active">Active</option>
-          <option value="completed">Completed</option>
-        </select>
-      </span>
-    </p>
+export enum STATUSES {
+  COMPLETED = 'completed',
+  ACTIVE = 'active',
+  ALL = 'all',
+}
```
BudnikOleksii, line 5: You don't need to write an enum name in upper case, check the example https://mate-academy.github.io/fe-program/js/extra/typescript
image

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,61 @@
...
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+interface FilterProps {
+  setStatus: (value: string) => void;
```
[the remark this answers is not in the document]
BudnikOleksii: > The `setStatus` function expects a string, but it would be more type-safe to use the `STATUSES` enum. Consider changing the type of `setStatus` to `(value: STATUSES) => void` to ensure that only valid status values are passed.

Good point

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,61 @@
...
+  };
+
+  return (
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select data-cy="statusSelect" onChange={handleSelect}>
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
BudnikOleksii, line 26: Use your enum for values

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,61 @@
...
+          onChange={event => setQuery(event.target.value)}
+        />
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span
+            className="icon is-right"
+            style={{ pointerEvents: 'all' }}
```
BudnikOleksii, line 47: Avoid inline styles, use classes instead

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -12,19 +1,61 @@
...
+            onClick={() => setQuery('')}
+          >
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
```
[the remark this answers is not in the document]
BudnikOleksii: > Consider adding an accessible label to the button to improve accessibility. This can be done by using the `aria-label` attribute.

Good point, so you can avoid linter disabling

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,73 @@
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+          return (
+            <tr data-cy="todo" className="" key={id}>
```
BudnikOleksii, line 33: Create a todo component

BudnikOleksii (changes requested):
Good work 👍

PR #2684, opened 2024-12-14
add solution

TargoniyAnatoliy:
[DEMO LINK](https://TargoniyAnatoliy.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,11 +24,11 @@ export const App: React.FC = () => {
     .filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()))
     .filter(todo => {
       switch (filterParameter) {
-        case 'completed':
+        case Filters.Completed:
           return todo.completed;
-        case 'active':
+        case Filters.Active:
           return !todo.completed;
-        case 'all':
+        case Filters.All:
         default:
           return true;
       }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,45 @@
...
+  const [query, setQuery] = useState('');
+  const [filterParameter, setFilterParameter] = useState<Filters>(Filters.All);
+
+  const selectedTodo = todos.find(todo => todo.id === selectedTodoId);
+
+  const filteredTodos: Todo[] = todos
+    .filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()))
+    .filter(todo => {
+      switch (filterParameter) {
+        case 'completed':
```
StasSokolov1: let's move `completed, active, all` to the enum

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,77 @@
...
-  </table>
-);
+      <tbody>
+        {todos.length === 0 ? (
+          <tr>
+            <td colSpan={4}>No data</td>
+          </tr>
+        ) : (
+          todos.map(todo => (
+            <tr data-cy="todo" className="" key={todo.id}>
```
StasSokolov1:
```suggestion
            <tr data-cy="todo" key={todo.id}>
```

let's remove redundant class name

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,25 +33,33 @@ export const TodoModal: React.FC = () => {
...
             </p>
 
             <p className="block" data-cy="modal-user">
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              {todo.completed ? (
+                <strong className="has-text-success">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
```
StasSokolov1, line 58: you can use one `strong` tag and just change class name and content

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,57 @@
...
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={e => setFilter(e.target.value as Filters)}
+          >
+            <option value={Filters.All}>All</option>
+            <option value={Filters.Active}>Active</option>
+            <option value={Filters.Completed}>Completed</option>
```
StasSokolov1, line 23: we can show these options using `.map()`

StasSokolov1 (changes requested):
Almost done

src/App.tsx
TargoniyAnatoliy: enum Filters already exists in types folder:  enum Filters {
  All = 'all',
  Active = 'active',
  Completed = 'completed',
}

src/components/TodoList/TodoList.tsx
TargoniyAnatoliy: forgot to use 'has-background-info-light' class if todo is selected

src/components/TodoModal/TodoModal.tsx
TargoniyAnatoliy: Desided to use these aproach, because it's more obvious and simplest than using many conditions in className and content

src/components/TodoFilter/TodoFilter.tsx
TargoniyAnatoliy: It's good idea, but I don't know how to use map to the enum object Filters: enum Filters {
  All = 'all',
  Active = 'active',
  Completed = 'completed',
}

BudnikOleksii: > It's good idea, but I don't know how to use map to the enum object Filters: enum Filters { All = 'all', Active = 'active', Completed = 'completed', }

You can use `Object.values/key/entries` to convert enum into an array

BudnikOleksii (approved):
👍

PR #2689, opened 2024-12-16
solution

SemenVodolazskij:
[DEMO LINK](https://SemenVodolazskij.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,38 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+import { Filters } from './types/Filters';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+
+  const [selectedTodoId, setSelectedTodoId] = useState<number | null>(null);
+
+  const [query, setQuery] = useState('');
+  const [status, setStatus] = useState<Filters>(Filters.ALL);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(data => setTodos(data))
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const selectedTodo = todos.find(todo => todo.id === selectedTodoId);
+
+  const filteredTodos = todos
+    .filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()))
+    .filter(todo => {
+      if (status === Filters.ALL) {
+        return true;
+      }
+
+      return status === Filters.COMPLETED ? todo.completed : !todo.completed;
+    });
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,14 +1,44 @@
...
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(data => setTodos(data))
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  const selectedTodo = todos.find(todo => todo.id === selectedTodoId);
+
+  const filteredTodos = todos
```
AnderZerfall, line 32: I'd suggest moving this function to a separate file (like into the folder 'utils') just for clarity. However, that's optional here

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,56 @@
...
-      <input
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+type Props = {
```
AnderZerfall, line 4: Actually, you don't use STATUS prop anywhere in that component, so consider removing it from the type.

src/App.tsx
```diff
@@ -17,18 +47,34 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
```
AnderZerfall, line 50: You can remove STATUS prop, 'cause you don't use it in that component

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,33 @@
...
+  todo: Todo;
+  onClose: () => void;
+};
+
+export const TodoModal: React.FC<Props> = props => {
+  const { todo, onClose } = props;
+  const [isLoading, setIsLoading] = useState(true);
+  const [user, setUser] = useState<User | null>(null);
+
+  useEffect(() => {
```
AnderZerfall, line 17: I'd rather use a 'catch' block here instead of allowing types to use NULL values by default.
That way, you don't need to worry about missing users, because that will be automatically handled by CATCH block. (for example, you can hide component and show an error message if you don't find any user by the given id).

The same is for Todos.

AnderZerfall (approved):
Nice job!

I left some comments below. Those are just my suggestions - nothing super important

src/App.tsx
BudnikOleksii: > I'd suggest moving this function to a separate file (like into the folder 'utils') just for clarity. However, that's optional here

Good point

src/App.tsx
```diff
@@ -1,14 +1,44 @@
...
+  }, []);
+
+  const selectedTodo = todos.find(todo => todo.id === selectedTodoId);
+
+  const filteredTodos = todos
+    .filter(todo => todo.title.toLowerCase().includes(query.toLowerCase()))
+    .filter(todo => {
+      if (status === Filters.ALL) {
+        return true;
+      }
```
BudnikOleksii, line 37: If `status === Filters.ALL` you don't need to filter at all

src/components/TodoFilter/TodoFilter.tsx
BudnikOleksii: > Actually, you don't use STATUS prop anywhere in that component, so consider removing it from the type.

But actually, you should use in your select

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,56 @@
...
+          className="input"
+          placeholder="Search..."
+          value={query}
+          onChange={event => setQuery(event.target.value)}
+        />
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
BudnikOleksii, line 42: Avoid inline styles, use classes instead

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,77 @@
...
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map(todo => (
+          <tr
+            key={todo.id}
+            data-cy="todo"
```
BudnikOleksii, line 33: Create a todo component

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,77 @@
...
+            className={
+              todo.id === selectedTodoId ? 'has-background-info-light' : ''
```
BudnikOleksii, line 35: To set classes conditionally use [classnames](https://www.npmjs.com/package/classnames) or [clsx](https://www.npmjs.com/package/clsx) library

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,77 @@
...
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => setSelectedTodoId(todo.id)}
+              >
+                <span className="icon">
+                  <i
+                    className={classNames(
+                      'far',
+                      todo.id === selectedTodoId ? 'fa-eye-slash' : 'fa-eye',
```
BudnikOleksii, line 66: You compare `todo.id === selectedTodoId` twice, create a variable

BudnikOleksii (approved):
Good work 👍 
Approved, but left comments to improve your solution

PR #2695, opened 2024-12-16
solution

VitaliiShc:
[DEMO LINK](https://<your_account>.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,53 @@
...
+      filteredTodos = todos.filter(todo => !todo.completed);
+    }
+
+    return filteredTodos.filter(todo =>
+      todo.title.toLowerCase().includes(query.toLowerCase()),
+    );
+  };
+
+  useEffect(() => {
+    setIsloading(true);
```
StasSokolov1: we can remove this line as we set isLoading true initially

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,55 @@
...
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            data-cy="statusSelect"
+            onChange={evt => setStatus(evt.target.value as Filters)}
+          >
+            <option value={Filters.All}>All</option>
+            <option value={Filters.Active}>Active</option>
+            <option value={Filters.Completed}>Completed</option>
```
StasSokolov1: let's iterate these options using .map()

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,80 @@
...
-      </tr>
-    </tbody>
-  </table>
-);
+      <tbody>
+        {todos.map(todo => (
+          <tr
+            data-cy="todo"
+            className={classNames(
+              todo === selectedTodo ? 'has-background-info-light' : '',
```
StasSokolov1: let's create `isSelectedTodo` const and reuse

StasSokolov1 (changes requested):
Let's fix a few comments before approving!

StasSokolov1 (approved):
Looks good to me, I'm approving it 🔥

PR #2711, opened 2024-12-26
solution

Yevhen-Filimonov:
[DEMO LINK](https://yevhen-filimonov.github.io/react_dynamic-list-of-todos/)


Krrampuss (changes requested):
4 tests fail, so unfortunately, I have to reject the solution(((

PR #2756, opened 2025-02-04
solution

djtyrf312:
[DEMO LINK](https://djtyrf312.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,40 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+export enum Status {
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
+}
 
 export const App: React.FC = () => {
+  const [status, setStatus] = React.useState<Status>(Status.all);
+  const [query, setQuery] = React.useState<string>('');
+  const [todosFromServer, setTodosFromServer] = React.useState<Todo[]>([]);
+  const [isOpenedTodo, setIsOpenedTodo] = React.useState<Todo | null>(null);
+  let displayedTodos = [...todosFromServer];
+
+  if (status === Status.active) {
+    displayedTodos = todosFromServer.filter(todo => !todo.completed);
+  }
+
+  if (status === Status.completed) {
+    displayedTodos = todosFromServer.filter(todo => todo.completed);
+  }
+
+  if (query) {
+    displayedTodos = displayedTodos.filter(todo =>
+      todo.title.toLowerCase().includes(query.toLowerCase()));
+  }
+
+  useEffect(() => {
+    getTodos().then(todos => {
+      setTodosFromServer(todos);
+    });
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+export enum Status {
```
naziamoff, line 12: would be nice to have a separate file with type definitions, typedefs.tsx
So we have related types in a single place

src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+
+  if (status === Status.completed) {
+    displayedTodos = todosFromServer.filter(todo => todo.completed);
+  }
+
+  if (query) {
+    displayedTodos = displayedTodos.filter(todo =>
+      todo.title.toLowerCase().includes(query.toLowerCase()),
+    );
+  }
```
naziamoff, line 37: what if we introduce a separate state for displayedTodos + util function?

It can look like this:

```suggestion
  const [displayedTodos, setDisplayedTodos] = useState(
    getDisplayedTodos(status, query)
  );
```

And the function will handle the stuff inside instead of keeping this low level logic in the component

src/App.tsx
```diff
@@ -17,18 +50,30 @@ export const App: React.FC = () => {
...
+                onStatusChange={setStatus}
+                onQueryChange={setQuery}
+                query={query}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {todosFromServer.length === 0 && <Loader />}
```
naziamoff, line 61: looks cleaner this way to me. No strong opinion though
```suggestion
              {!todosFromServer.length && <Loader />}
```

src/App.tsx
```diff
@@ -17,18 +50,30 @@ export const App: React.FC = () => {
...
+              {todosFromServer.length !== 0 && (
```
naziamoff, line 62: would be nice to have one line indent to make neighbouring condition statements more readable

src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+  all = 'all',
+  active = 'active',
+  completed = 'completed',
+}
+
+export const App: FC = () => {
+  const [status, setStatus] = useState<Status>(Status.all);
+  const [query, setQuery] = useState<string>('');
+  const [todosFromServer, setTodosFromServer] = useState<Todo[]>([]);
+  const [isOpenedTodo, setIsOpenedTodo] = useState<Todo | null>(null);
```
naziamoff, line 22: we start the variable name with `is/can/should/etc.` only for boolean values. I thought it is a boolean at first. What about `activeTodo`?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,63 @@
...
+            onQueryChange(event.target.value);
+          }}
+        />
+        <span className="icon is-left">
+          <i className="fas fa-magnifying-glass" />
+        </span>
+
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
naziamoff, line 49: it is better to fix linting errors when possible instead of disabling them

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,79 @@
...
+                  onClick={() => {
+                    onClickedPreview(todo);
+                  }}
+                >
+                  <span className="icon">
+                    {isOpenedTodo?.id === id ? (
+                      <i className="far fa-eye-slash" />
+                    ) : (
+                      <i className="far fa-eye" />
+                    )}
```
naziamoff, line 69: looks cleaner to me this way
```suggestion
                    {isOpenedTodo?.id === id 
                      ?  <i className="far fa-eye-slash" />
                      :  <i className="far fa-eye" />
                    }
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,25 +34,39 @@ export const TodoModal: React.FC = () => {
...
             <p className="block" data-cy="modal-user">
               {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              <strong
+                className={classNames({
+                  'has-text-success': completed,
+                  'has-text-danger': !completed,
+                })}
+              >
+                {completed ? 'Done' : 'Planned'}
```
naziamoff, line 64: ternary conditions are usually kept on different lines, so they are better readable

Vinicius-Ferraz (changes requested):
Please review the solutions

PR #2781, opened 2025-02-16
init commit

nex1994:
[DEMO LINK](https://nex1994.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,6 +18,7 @@ export const App: React.FC = () => {
   const [isLoaded, setIsLoaded] = useState<boolean>(false);
   const [completeStatus, setCompleteStatus] = useState<string>('all');
   const [searchInput, setSearchInput] = useState<string>('');
+  const [error, setError] = useState('');
 
   const todo = todos?.find(t => t.id === selectedTask);
   let filteredTodos = todos;
@@ -54,12 +55,12 @@ export const App: React.FC = () => {
           setIsLoaded(true);
         })
         .catch(() => {
-          throw new Error('No todos found');
+          setError('Todos cant be loaded. Try again later');
         });
     };
 
     loadData();
-  }, [todo?.userId]);
+  }, []);
 
   return (
     <>
@@ -67,7 +68,7 @@ export const App: React.FC = () => {
         <div className="container">
           <div className="box">
             <h1 className="title">Todos:</h1>
-
+            {error !== '' && <p>{error}</p>}
             <div className="block">
               <TodoFilter
                 searchInput={searchInput}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+export type FilterByCompletion = 'all' | 'active' | 'completed';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null | undefined>(null);
+  const [selectedTask, setSelectedTask] = useState<number | null>(null);
+  const [isLoaded, setIsLoaded] = useState<boolean>(false);
+  const [completeStatus, setCompleteStatus] = useState<string>('all');
```
zibi-dev: use enum or obj signed as constant for this kind of things.

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
+      break;
+    case 'completed':
+      filteredTodos = todos?.filter(task => task.completed === true);
+      break;
+    case 'all':
+      filteredTodos = todos;
+      break;
+    default:
+      break;
+  }
```
zibi-dev: You can simply use `useMemo` here
```suggestion
const filteredTodos = useMemo(() => {
  const normalizedSearch = searchInput.toLowerCase().trim();
  
  let filtered = todos;
  
  switch (completeStatus) {
    case 'active':
      filtered = filtered?.filter(task => !task.completed);
      break;
    case 'completed':
      filtered = filtered?.filter(task => task.completed);
      break;
    case 'all':
      break;
    default:
      throw new Error("Invalid completeStatus");
  }
  
  if (!normalizedSearch) {
    return filtered;
  }
    
  return filtered?.filter(todo => 
    todo.title?.toLowerCase().trim().includes(normalizedSearch)
  );
}, [todos, completeStatus, searchInput]);
```

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+export type FilterByCompletion = 'all' | 'active' | 'completed';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null | undefined>(null);
+  const [selectedTask, setSelectedTask] = useState<number | null>(null);
+  const [isLoaded, setIsLoaded] = useState<boolean>(false);
```
zibi-dev: Im fan of having status instead of booleans isLoaded etc. Here is article about that

https://kentcdodds.com/blog/stop-using-isloading-booleans

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
+      : null;
+  });
+
+  useEffect(() => {
+    setIsLoaded(false);
+    const loadData = () => {
+      getTodos()
+        .then(data => {
+          setTodos(data);
+          setIsLoaded(true);
```
zibi-dev: There is finally block for isLoaded status

src/App.tsx
```diff
@@ -1,34 +1,97 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { getTodos } from './api';
+
+export type FilterByCompletion = 'all' | 'active' | 'completed';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[] | null | undefined>(null);
+  const [selectedTask, setSelectedTask] = useState<number | null>(null);
```
zibi-dev: Always look for good name for variables. Here name indicates that you keep whole todo item but its only id

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,54 @@
...
+    <form className="field has-addons">
+      <p className="control">
+        <span className="select">
+          <select
+            onChange={e => setCompleteStatus(e.target.value)}
+            data-cy="statusSelect"
+          >
+            <option value="all">All</option>
+            <option value="active">Active</option>
+            <option value="completed">Completed</option>
```
zibi-dev: it should come from some source object or Enum

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,75 @@
...
+                  className="button"
+                  type="button"
+                >
+                  <span className="icon">
+                    <i
+                      className={
+                        todo.id === selectedTask
+                          ? 'far fa-eye-slash'
+                          : 'far fa-eye'
+                      }
```
zibi-dev: You have installed package `classnames` that is great tool for conditional class assignments

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,43 +1,84 @@
...
+                }
+              >
+                {todo?.completed ? 'Done' : 'Planned'}
+              </strong>
 
               {' by '}
 
-              <a href="mailto:Sincere@april.biz">Leanne Graham</a>
+              {error === '' ? (
+                <a href="mailto:Sincere@april.biz">{user?.name}</a>
```
zibi-dev: what about user email and what if email is not specified or not valid?

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,43 +1,84 @@
...
               className="modal-card-title has-text-weight-medium"
               data-cy="modal-header"
             >
-              Todo #2
+              Todo #{todo?.id}
             </div>
 
             {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-            <button type="button" className="delete" data-cy="modal-close" />
+            <button
```
zibi-dev, line 59: As extra implement functionality that closes modal when you click outside of modal window

Zibi95 (changes requested):

src/components/TodoModal/TodoModal.tsx
```diff
@@ -82,9 +82,9 @@ export const TodoModal: React.FC<Props> = ({ setSelectedTask, todo }) => {
               {' by '}
 
               {error === '' ? (
-                <a href="mailto:Sincere@april.biz">{user?.name}</a>
+                <a href={user?.email}>{user?.name}</a>
```
zibi-dev: It should still use `mailto:${user.email}` and if there is no user email i mean its empty string then it should be nomrla text

Zibi95 (changes requested):

nex1994:
Hi. I just send changed version of email link. Can you give me a feedback if i use object as const correctly in this example and if i understood it correctly is it better practice to use objects as const instead of enums? Also useState type should be `typeof OBJNAME` or just `string` is enought?

Zibi95:
> Hi. I just send changed version of email link. Can you give me a feedback if i use object as const correctly in this example and if i understood it correctly is it better practice to use objects as const instead of enums? Also useState type should be `typeof OBJNAME` or just `string` is enought?

It depends from you. I more like objects. Here are some problems with enums https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh . Yep it should be typeof for more strict values

Zibi95 (approved):

PR #2838, opened 2025-03-24
solution

dmytroKolesnikov52:
[DEMO LINK](https://Mazafakerqq.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { getTodos } from '../../api';
 import { Todo } from '../../types/Todo';
 import { Loader } from '../Loader';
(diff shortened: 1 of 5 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,111 @@
...
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+export const TodoList: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(true);
+  const [filter, setFilter] = useState('');
```
AnOvramenko: Boolean variables should start with 'is' or 'has'

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,111 @@
...
-            </span>
-          </button>
-        </td>
-      </tr>
+  useEffect(() => {
+    getTodos().then(fetchedTodos => {
+      setTodos(fetchedTodos);
+      setLoading(false);
+    });
+  }, []);
```
AnOvramenko, line 20: You should use catch() to avoid error

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,111 @@
...
+  const filteredTodos = todos.filter(todoItem => {
+    const matchesQuery = query
+      ? todoItem.title.toLowerCase().includes(query.toLowerCase().trim())
+      : true;
+    const matchesFilter =
+      filter === 'active'
+        ? !todoItem.completed
+        : filter === 'completed'
+          ? todoItem.completed
+          : true;
```
AnOvramenko: Not a good idea to use ternary like this, I suggest you move this to a separate function, and then you can use useMemo hook for optimization

AnOvramenko (approved):
Good job! 👍 
Check comments above
It is a little bit of strange logic to move all components into TodoList. Now it looks like you don't need the App file and I think the idea wasn't in it.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,122 @@
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-);
+      switch (hasFilter) {
+        case 'active':
```
ArturErmolenko, line 31: magic string

ArturErmolenko (approved):

PR #2861, opened 2025-03-29
Solution

Olehgoon093:
[DEMO LINK](https://Olehgoon093.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,14 +1,74 @@
...
+  const closeModal = () => {
+    setSelectTodo(null);
+  };
+
+  useEffect(() => {
+    setIsTodoloading(true);
+    getTodos().then(todos => {
+      setTodoList(todos);
+      setIsTodoloading(false);
+    });
```
maxbilfed, line 35: Would be good to add error catching, if received result was empty for example

src/TodoStatus.ts
maxbilfed, line 1: For better organization, type files are better to placed into the "types" folder, alongside others

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,27 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
+import { TodoStatus } from '../../TodoStatus';
+
+type Props = {
+  onSelectStatus: (status: TodoStatus) => void;
+  selectedStatus: string;
```
maxbilfed: Probably better to use **TodoStatus** type instead of **string**

src/components/TodoList/TodoList.tsx
```diff
@@ -16,85 +21,39 @@ export const TodoList: React.FC = () => (
...
+          <td className="has-text-right is-vcentered">
+            <button
+              data-cy="selectButton"
+              className="button"
+              type="button"
+              onClick={() => onSelect(todo)}
+            >
+              <span className="icon">
+                <i className="far fa-eye" />
+              </span>
```
maxbilfed, line 52: I think when selected, the icon is supposed to be different on the selected one, with this class: "'fas fa-eye-slash'"

maxbilfed (approved):
Good work, but there are some improvements that could be done

PR #2863, opened 2025-03-30
Develop

VladRoz95:
[DEMO LINK](https://VladRoz95.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,50 +19,42 @@ export const App: React.FC = () => {
   );
   const [searchByName, setSearchByName] = useState<string>('');
   const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [todosData, setTodosData] = useState<Todo[]>([]);
 
   useEffect(() => {
     setIsTodosLoading(true);
     getTodos().then(todos => {
       // eslint-disable-next-line no-console
-      console.log(todos);
+      setTodosData(todos);
       setTodoList(todos);
       setIsTodosLoading(false);
     });
   }, []);
 
   useEffect(() => {
-    setIsTodosLoading(true);
-    getTodos()
-      .then(todos => {
-        if (filterByStatus === TodoStatus.ALL) {
-          return todos;
-        }
+    let filteredTodos = [...todosData];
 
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+          todo.title.toLocaleLowerCase().includes(searchByName.toLowerCase()),
+        );
+      })
+      .then(todos => {
+        // eslint-disable-next-line no-console
+        console.log(todos);
+        setTodoList(todos);
+        setIsTodosLoading(false);
+      });
+  }, [filterByStatus, searchByName]);
```
ArturErmolenko: not the best move, so every time filter change or query change you make API request, imagine you have 10 000 users platform .... your API will die

you should make request just once on component mount and put data in state
and than working with state without extra request

src/App.tsx
```diff
@@ -1,14 +1,69 @@
...
+    TodoStatus.ALL,
+  );
+  const [searchByName, setSearchByName] = useState<string>('');
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    setIsTodosLoading(true);
+    getTodos().then(todos => {
+      // eslint-disable-next-line no-console
+      console.log(todos);
```
ArturErmolenko: code on GH should be clear, remove console
```suggestion
```

ArturErmolenko (changes requested):
Good job 👍 
But let's improve performance and kill second useEffect

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+    getTodos().then(todos => {
+      // eslint-disable-next-line no-console
+      setTodosData(todos);
+      setTodoList(todos);
+      setIsTodosLoading(false);
+    });
+  }, []);
+
+  useEffect(() => {
+    let filteredTodos = [...todosData];
```
alexander-ignatow: You are doing an extra copy here

```suggestion
    let filteredTodos = todosData;
```

alexander-ignatow (approved):
Good job! 👍

PR #2936, opened 2025-06-01
Develop

14lucas-mendes:
[DEMO LINK](https://14lucas-mendes.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
 /* eslint-disable max-len */
-import React from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +6,30 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+
+const API_TODOS = 'http://localhost:5173/api/todos.json';
+const API_USER = 'http://localhost:5173/api/users.json';
+
+fetch(API_USER).then(response => response.json());
 
 export const App: React.FC = () => {
+  function getAll(): Promise<Todo[]> {
+    return fetch(API_TODOS).then(response => response.json());
+  }
+
+  function getAllActive() {
+    return getAll().then(todos =>
+      todos.filter(todo => todo.completed === false),
+    );
+  }
+
+  function getAllCompleted() {
+    return getAll().then(todos =>
+      todos.filter(todo => todo.completed === true),
+    );
+  }
+
   return (
     <>
       <div className="section">
@@ -17,7 +38,11 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                onClickAll={() => getAll}
+                onClickHandlerActive={() => getAllActive}
+                onClickHandlerCompleted={() => getAllCompleted}
+              />
             </div>
 
             <div className="block">
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,83 +7,120 @@ import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 import { getTodos } from './api';
-import { useEffect, useState } from 'react';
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -17,18 +59,24 @@ export const App: React.FC = () => {
...
+                filterByCompleted={filterByCompleted}
+                resetFilters={resetFilters}
+                query={query}
+                setQuery={handleQueryChange}
+                closeByQuery={closeByFilter}
+              />
             </div>
 
             <div className="block">
               <Loader />
```
pedro-ruas, line 111: Why is this fixed? Your page seems like it is Loading all the time

pedro-ruas (changes requested):
Your app has a fixed Loading in front of the screen, and I see it did not pass all tests

src/App.tsx
```diff
@@ -17,18 +66,24 @@ export const App: React.FC = () => {
...
+                query={query}
+                setQuery={handleQueryChange}
+                closeByQuery={closeByFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!isLoading && <Loader />}
```
pedro-ruas: If it is NOT loading, you are showing the Loader?
Makes no sense...

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+import { getTodos } from './api';
+import { useEffect, useState } from 'react';
+import { Todo } from './types/Todo';
 
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [originalTodos, setOriginalTodos] = useState<Todo[]>([]); // guarda a lista original
+  const [query, setQuery] = useState('');
+  const [open, setOpen] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
```
pedro-ruas, line 17: If your component will always request the Todos when it first starts, and when requesting it loads, why not already start the state with `true` ?

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+        setTimeout(() => {
+          setTodos(response);
+          setOriginalTodos(response);
+          setIsLoading(false);
+        }, 1000);
+      });
+    }
+
+    const normalizedQuery = query.trim().toLowerCase();
+    const filtered = originalTodos.filter(todo =>
```
pedro-ruas: This is not good. You are performing multiple requests, one for each time the query changes.
So if I type `Some todo name`, your app will do 14 unnecessary requests.

You can handle the filtered **outside** the state and useEffect.
**Inside** the useEffect, you handle the first request to get the todos, when your app doesn't have then. And ignore the query.
Outside the useEffect, just before the component's `return`, you declare filteredTodos...
You don't need the "todos" state, just the one you are calling "originalTodos"

src/App.tsx
```diff
@@ -1,14 +1,63 @@
...
+  }, [query, originalTodos]);
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+  };
+
+  const closeByFilter = () => {
+    setQuery('');
+  };
+
```
pedro-ruas, line 83: Declare the filteredTodos here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,58 @@
...
+            <p
+              className={
+                todo.completed ? 'has-text-success' : 'has-text-danger'
+              }
+            >
+              {todo.title}
+            </p>
+          </td>
+          <td className="has-text-right is-vcentered">
+            <button data-cy="selectButton" className="button" type="button">
```
pedro-ruas: This button isn't doing anything. You need to pass `setOpen` to it too

src/App.tsx
```diff
@@ -17,18 +66,24 @@ export const App: React.FC = () => {
...
+                setQuery={handleQueryChange}
+                closeByQuery={closeByFilter}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!isLoading && <Loader />}
+              <TodoList todos={todos} />
```
pedro-ruas: Aren't you missing any prop here? Maybe some way for the Todo to open the TodoModal?

pedro-ruas (changes requested):
Several improvements needed. There is still another Loading that is always present, styles are changing when selecting todos, modal isn't working....
And tests are failing too.

PR #2943, opened 2025-06-05
add solution

zbvika:
[DEMO LINK](https://zbvika.github.io/react_dynamic-list-of-todos/)


ystyslo (approved):
Good job!

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,39 +1,75 @@
...
           <div className="modal-card-body">
             <p className="block" data-cy="modal-title">
-              quis ut nam facilis et officia qui
+              {todo.title}
             </p>
 
             <p className="block" data-cy="modal-user">
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              {todo.completed ? (
```
ystyslo, line 64: You can use conditional rendering only for text
{todo?.completed ? `Done` : `Planned`}

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,39 +1,75 @@
...
+  onSelected: (todo: Todo | null) => void;
+  setShowModal: (value: boolean) => void;
+  showModal: boolean;
+}
+
+export const TodoModal: React.FC<Props> = ({
+  todo,
+  onSelected,
+  setShowModal,
+  showModal,
```
ystyslo, line 18: no need to use state showModal. You can show it when todo is selected

zbvika: Thank you for the comment) showModal is also used to correctly display the loader in TodoModal. The logic is that this state prevents the loader from appearing if no user is selected, and allows it to be triggered only when we are waiting for a response from the server.

src/components/TodoModal/TodoModal.tsx
zbvika: Thank you for the comment :) If I understood you correctly, then if we use conditional rendering only for the text, we would also need to apply styles conditionally. To simplify things, I added the condition directly to the tags.

ystyslo: It's your will☺️ I just suggested👌🏻

PR #2962, opened 2025-06-09
task solution

YarRudenko:
[DEMO LINK](https://YarRudenko.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,100 +1,85 @@
 import React from 'react';
 
-export const TodoList: React.FC = () => (
+export interface Todo {
+  id: number;
+  title: string;
+  completed: boolean;
+  userId: number;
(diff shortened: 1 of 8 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,85 @@
...
-              <i className="far fa-eye" />
-            </span>
-          </button>
-        </td>
-      </tr>
+      {todos.map(todo => (
+        <tr
+          key={todo.id}
+          data-cy="todo"
+          className={todo.completed ? 'has-background-info-light' : ''}
```
koliiis, line 37: I think 'has-background-info-light' should be on selected todo, not on completed

koliiis (approved):
Great job!!
But in TodoList might be a mistake, or not🤷‍♀️

PR #2987, opened 2025-06-13
added task solution

torry455:


deadslowright (approved):
Keep it up^^

cagataycali:
## 🐛 Critical Bug Fix Pushed: Attribute Assignment Ordering

**Root Cause:** `_infer_config_name()` was called on line 1110 before `self.robot_variant` and `self.output_resolution` were assigned (lines 1118-1119), causing `AttributeError` on every constructor call.

**Impact:** 57 of 69 tests failed — the class was completely non-functional.

**Fix:** Moved `self.output_resolution` and `self.robot_variant` assignments to **before** `self.config_name = config_name or self._infer_config_name()`. This is a 2-line reorder.

**Result:** ✅ 69/69 tests passing in 0.17s

*Note: This same pattern exists safely in `CosmosTrainer.__init__()` because `_infer_config_name()` there only uses `self.suite` which IS assigned before the call. The difference is that `CosmosTransferTrainer._infer_config_name()` additionally depends on `self.robot_variant` and `self.output_resolution`.*

PR #3069, opened 2025-08-16
dynamic-todos solution

NicssLucik02:


src/App.tsx
```diff
@@ -1,14 +1,87 @@
...
+            .trim()
+            .includes(textInput.toLowerCase().trim()),
+        );
+
+        setPreparedTodos(filtered);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, [filterStatus, textInput]);
```
ArturErmolenko: AI style
why we need useEffect? can we just load todo in useEffect and have preparedTodos function?
```suggestion
  useEffect(() => {
    todos
      .then(data => {
        const filtered = data.filter(todo =>
          todo.title
            .toLowerCase()
            .trim()
            .includes(textInput.toLowerCase().trim()),
        );

        setPreparedTodos(filtered);
      })
      .finally(() => {
        setIsLoading(false);
      });
  }, [filterStatus, textInput]);
```

src/App.tsx
```diff
@@ -1,14 +1,87 @@
...
+
+  useEffect(() => {
+    if (!selectedTodo) {
+      return;
+    }
+
+    setPreparedUser(null);
+    getUser(selectedTodo.userId).then(user => setPreparedUser(user));
+  }, [selectedTodo]);
```
ArturErmolenko: this should handle with handlers not with useEffect we should get rid of this

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,144 @@
...
+                {todo.completed === true && (
+                  <span className="icon" data-cy="iconCompleted">
+                    <i className="fas fa-check" />
+                  </span>
+                )}
+              </td>
+              <td className="is-vcentered is-expanded">
+                <p
+                  className={
+                    todo.completed === true
```
ArturErmolenko:
```suggestion
                    todo.completed
```
???

ArturErmolenko: use classname lib

src/components/TodoModal/TodoModal.tsx
```diff
@@ -15,25 +32,36 @@ export const TodoModal: React.FC = () => {
...
 
             <p className="block" data-cy="modal-user">
-              {/* <strong className="has-text-success">Done</strong> */}
-              <strong className="has-text-danger">Planned</strong>
+              <strong
+                className={
+                  !selectedTodo.completed
+                    ? 'has-text-danger'
+                    : 'has-text-success'
+                }
```
ArturErmolenko, line 68: classneme lib

src/App.tsx
ArturErmolenko: do we need load user in App? where we need user?
i believe we should load user in modal

src/App.tsx
```diff
@@ -1,14 +1,87 @@
...
 export const App: React.FC = () => {
+  const [preparedTodos, setPreparedTodos] = useState<Todo[]>([]);
+  const [filterStatus, setFilterStatus] = useState<FilterTypes>(
+    FilterTypes.All,
+  );
+  const [textInput, setTextInput] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [todoModalStatus, setTodoModalStatus] = useState(false);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+  const [preparedUser, setPreparedUser] = useState<User | null>(null);
```
ArturErmolenko:
```suggestion
```
we don't need this state

src/App.tsx
```diff
@@ -1,14 +1,87 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getActiveTodos, getcompletedTodos, getTodos, getUser } from './api';
+import { Todo, FilterTypes } from './types/Todo';
+import { User } from './types/User';
 
 export const App: React.FC = () => {
+  const [preparedTodos, setPreparedTodos] = useState<Todo[]>([]);
```
ArturErmolenko:
```suggestion
```
should be const

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,144 @@
...
-          </button>
-        </td>
-      </tr>
+      <tbody>
+        {preparedTodos.map((todo: Todo) => {
+          return (
+            <tr data-cy="todo" className="" key={todo.id}>
+              <td className="is-vcentered">{todo.id}</td>
+              <td className="is-vcentered">
+                {todo.completed === true && (
```
ArturErmolenko:
```suggestion
                {todo.completed && (
```

ArturErmolenko (changes requested):
Good start Let's improve your code

mkyloa (approved):
👍

PR #3075, opened 2025-08-21
added task solution

horidor:
Link: [DEMO LINK](https://horidor.github.io/react_dynamic-list-of-todos/)


AARON-FOX (approved):
Good job🔥

horidor:
@mateacademy-ai-mentor 
2. As per example, it is not required to actually clear the select options, only the search filter one. Therefore there is no such requirement.
4. TodoModal blocks any input on the website, and Loader blocks any input (since it doesn't actually show detailed information before todo appears). Therefore, race condition can never occur, and there is no need to guard against it.

PR #3079, opened 2025-08-21
add solution

YuliaProkipchuk:
[DEMO LINK](https://YuliaProkipchuk.github.io/react_dynamic-list-of-todos/)


jeolinna (approved):
Well done!

YuliaProkipchuk:
@mateacademy-ai-mentor There is no need to use AbortController here because when user selects todo, modal window opens immediatly with a loader so it's impossible for user to choose another todo before the previous one loades. As for classNames library it is installed and imported correctly and all tests pass.

YuliaProkipchuk:
@mateacademy-ai-mentor There is no need to use AbortController here because when user selects todo, modal window opens immediatly with a loader before fetching data so it's impossible for user to choose another todo before the previous one loades.

PR #3085, opened 2025-08-22
Add: Solution

bebelekele:
[DEMO LINK](https://bebelekele.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,58 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
+import { Todo } from './types/Todo';
+
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { TodoModal } from './components/TodoModal';
+
+export enum Filter {
+  All = 'all',
+  Active = 'active',
+  Completed = 'completed',
+}
+
+function filterTodo(selectedFilter: Filter, todo: Todo) {
+  switch (selectedFilter) {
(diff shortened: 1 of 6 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -47,7 +47,9 @@ export const TodoList: React.FC<Props> = ({
                   </p>
                 )}
 
-                {!todo.completed && <p>{todo.title}</p>}
+                {!todo.completed && (
+                  <p className="has-text-danger">{todo.title}</p>
+                )}
```
derpedcatto: You forgot to add green text color for completed tasks. Just use classnames for this:
```tsx
          <td className="is-vcentered is-expanded">
            <p
              className={cn({
                'has-text-danger': !todo.completed,
                'has-text-success': todo.completed,
              })}
            >
              {todo.title}
            </p>
          </td>

File: src/App.tsx [hunk 2]
```diff
@@ -1,14 +1,58 @@
...
+      setAllTodos(todosFromServer);
+      setLoader(false);
+    });
+  }, []);
+
+  const todos = allTodos.filter(
+    todo =>
+      filterTodo(selectedFilter, todo) &&
+      todo.title.toLowerCase().includes(inputValue.toLowerCase()),
+  );
```
derpedcatto, line 54: It is better to filter/prepare everything in one function `filterTodo`, because it can be confusing on why you have different filters in different places. Here is an example of how you can filter everything in bounds of one function:
```ts
const prepareTodos = (
  todos: Todo[],
  filterOption: FilterOptions,
  query: string,
) => {
  return todos
    .filter(todo => {
      switch (filterOption) {
        case FilterOptions.Completed:
          return todo.completed;

        case FilterOptions.Active:
          return todo.completed === false;

        case FilterOptions.All:
          return true;
      }
    })
    .filter(todo => {
      return todo.title
        .toLocaleLowerCase()
        .includes(query.trim().toLocaleLowerCase());
    });
};
```

derpedcatto (approved):

PR #3090, opened 2025-08-22
Add solution2

yanaovcharyk:
[DEMO LINK](https://yanaovcharik.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -8,7 +8,78 @@ import { TodoFilter } from './components/TodoFilter';
...
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    loadTodos();
+  }, []);
+
+  const handleOpenMod = (todo: Todo) => {
+    if (todo) {
```
alexander-ignatow: why do you need this check?
This is redundant as Todo object is always truthly

src/App.tsx
```diff
@@ -8,7 +8,78 @@ import { TodoFilter } from './components/TodoFilter';
...
+      setSelectedTodo(todo);
+      setIsModalOpen(true);
```
alexander-ignatow: You don't need a separate state for `isModalOpen`

remove it and use something like
```
const isModalOpen = !!selectedTodo
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,61 @@
...
-        data-cy="searchInput"
-        type="text"
-        className="input"
-        placeholder="Search..."
-      />
-      <span className="icon is-left">
-        <i className="fas fa-magnifying-glass" />
-      </span>
+interface Props {
+  enableTitle: () => void;
```
alexander-ignatow: Please rename this prop and aal related handlers

1. It is very hard to guess, what it does
2. Event handler props commonly start on `on`. 
3. For instance: `onTitleClick`, `onTitleEnable`
4. If you are in doubt with the variable name you can use copilot or any AI chat  ;)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,30 +1,61 @@
...
+  setQuery: Dispatch<SetStateAction<string>>;
+  query: string;
+  chousedFilter: string;
+  setChousedFilter: Dispatch<SetStateAction<string>>;
```
alexander-ignatow: Mind the grammar 😉
choose - chose - chosen

Even more - I strongly recommend you to use Code Spell Checker VSCode extention

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,71 @@
...
-            <i className="fas fa-check" />
-          </span>
-        </th>
-        <th>Title</th>
-        <th> </th>
-      </tr>
-    </thead>
+interface Props {
+  todos: Todo[];
+  openMod: (todo: Todo) => void;
```
alexander-ignatow: I would rename it

It is a callback, so starts with "on"
and the name should sound like an action (what has happened) - not like reaction request (what I want you to do)

For instance - `onTodoClicked`, `onTodoSelected`, etc

alexander-ignatow (changes requested):
Good job, but some issues are needed to be fixed

src/components/TodoFilter/TodoFilter.tsx
yanaovcharyk: thank you ! 😊 
i added the changes

DanilWeda (approved):
Good job. 
Approved. 
Many thanks!

PR #3098, opened 2025-08-25
fix

IlchenkoDima:
https://IlchenkoDima.github.io/react_dynamic-list-of-todos/


pbahrynovskyi (approved):
Good job! Everything works!
I have one little notice, there is a class has-background-info-light. I think (as it was in example) that you should add it only when your item in todo is checked. If you have time you might try to fix it.

PR #3151, opened 2025-10-03
add solution

Aiiyuu:
[DEMO LINK](https://aiiyuu.github.io/react_dynamic-list-of-todos/)


Aiiyuu:
I'm still struggling with `useCallback` and `React.memo` I think I'm overusing them. Life was great until I learned about memoization - now I feel paranoid every time I create a function or component because I'm never sure whether it's a good idea to use memoization in that case)

src/types/Todo.ts
```diff
@@ -4,3 +4,10 @@ export interface Todo {
   completed: boolean;
   userId: number;
 }
+
+export enum TodoStatus {
+  All = 'All',
+  Active = 'Active',
+  Completed = 'Completed',
+  Default = All,
```
781calmstate, line 12: It's not a common practice, so in my opinion we shouldn't provide default options which use this enum's  other options

781calmstate (approved):
Your code is fine even with memoization.
But I'd like to share with you a few important notes about react memoization to help you to get rid from your confusion and paranoia 😅 

The memoization approaches usually used when you have some heavy JS functionality that takes a lot of time and makes it really difficult for the client to process.
 And you can guess - it's a pretty rare scenario and usually not our case. JS usually fast enough to re-computate the whole component in a milliseconds.

So let's review pros and cons of react memoization:

✅ Pros:
• Performance boost for expensive operations:
When a function does heavy computation (e.g., sorting large data, running complex math, or rendering heavy UI), memoization stores the result for given inputs.
Next time the function is called with the same inputs, it returns the cached result instantly instead of recomputing.

• Helps React rendering efficiency:
In React, memoizing components (React.memo) or computed values (useMemo, useCallback) prevents unnecessary re-renders when inputs (props or dependencies) haven’t changed.
This can make the UI more responsive and reduce wasted rendering cycles.

⚠️ Cons:
• Higher memory usage:
Since we are caching the values, those values should be stored somewhere. Cached values live in memory until explicitly cleared or garbage-collected.
For frequently changing or large input sets, this can quickly bloat memory and even hurt performance.

• Complexity in cache invalidation:
When inputs are dynamic or derived from external state (like API responses or user input), you must ensure the cache updates correctly.
Otherwise, the app may show stale data because it’s returning an outdated cached result.
This creates unpleasant dev experience and forces you to track every dependency of each memoized value. If you use React.memo(Component) - it forces you to make sure that every passed props are memozied as well otherwise the memoization will not work. You would have to memoize every non-primitive prop - the callbacks, objects, arrays. **And you have to remember that every time you use memoization in react, you are forcing react to perform additional computation for each memoized value to make sure if dependency changed or not, those additional computation don't come for free, they also take some resources.**

• Not always worth the overhead:
For simple or fast computations (like adding two numbers), memoization adds unnecessary overhead—checking the cache can be slower than just re-running the function.

• Can hide deeper performance issues:
Sometimes, developers use memoization to “bandage” slow code rather than fixing the real cause (inefficient loops, large re-renders, poor data structure choices).
That can make debugging and optimization harder later.

**My advice is - use memoization only when profiling shows real performance bottlenecks and you are really sure that here you have the performance problem**

PR #3165, opened 2025-10-17
feat(app): implement full dynamic todos flow with loader, modal, filt…

S-Khraban:
[DEMO LINK](https://S-Khraban.github.io/react_dynamic-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,16 +11,14 @@ import { Loader } from './components/Loader';
 
 import { getTodos } from './api';
 import type { Todo } from './types/Todo';
-
-export const FILTERS = ['all', 'completed', 'active'] as const;
-type Filter = typeof FILTERS[number];
+import { Filter } from './types/Filter';
 
 export const App: React.FC = () => {
   const [todos, setTodos] = useState<Todo[]>([]);
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const [selectedId, setSelectedId] = useState<number | null>(null);
-  const [filter, setFilter] = useState<Filter>('all');
+  const [filter, setFilter] = useState<Filter>(Filter.All);
   const [query, setQuery] = useState('');
 
   const handleSelectTodo = useCallback((id: number | null) => {
@@ -47,28 +45,28 @@ export const App: React.FC = () => {
   };
 
   useEffect(() => {
-    void handleLoadTodos();
+    handleLoadTodos();
   }, []);
 
   const normalizedQuery = query.trim().toLowerCase();
 
-  const matchesStatus = (t: Todo, current: Filter) => {
-    if (current === 'all') return true;
-    if (current === 'completed') return t.completed;
+  const isStatusMatch = (t: Todo, current: Filter) => {
+    if (current === Filter.All) return true;
+    if (current === Filter.Completed) return t.completed;
     return !t.completed;
   };
 
-  const matchesQuery = (t: Todo, q: string) =>
+  const doesTitleMatchQuery = (t: Todo, q: string) =>
     t.title.toLowerCase().includes(q);
 
   const filteredTodos = todos.filter(
-    t => matchesStatus(t, filter) && matchesQuery(t, normalizedQuery),
+    t => isStatusMatch(t, filter) && doesTitleMatchQuery(t, normalizedQuery),
   );
 
   const selectedTodo =
     selectedId == null
       ? null
-      : (filteredTodos.find(t => t.id === selectedId) ?? null);
+      : filteredTodos.find(t => t.id === selectedId) ?? null;
 
   return (
     <>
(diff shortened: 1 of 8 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,15 +38,24 @@ export const TodoFilter = () => (
...
       <span className="icon is-left">
         <i className="fas fa-magnifying-glass" />
       </span>
 
-      <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-        {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
+      {query && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
781calmstate: why wouldn't you implement it within class instead of `style`?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,30 @@
...
       <span className="select">
-        <select data-cy="statusSelect">
+        <select
+          data-cy="statusSelect"
+          value={value}
+          onChange={e => onChange(e.target.value as Filter)}
+        >
           <option value="all">All</option>
           <option value="active">Active</option>
           <option value="completed">Completed</option>
```
781calmstate: if you use values from your strict type, you should use enum

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,82 @@
...
+                >
+                  <span className="icon">
+                    <i
+                      className={`far ${isSelected ? 'fa-eye-slash' : 'fa-eye'}`}
+                    />
+                  </span>
+                </button>
+              </td>
+            </tr>
+          );
```
781calmstate, line 80: you should've moved this logic into separated component

src/components/TodoList/TodoList.tsx
```diff
@@ -1,100 +1,82 @@
...
+              <td className="has-text-right is-vcentered">
+                <button
+                  data-cy="selectButton"
+                  className="button is-small"
+                  type="button"
+                  onClick={() => handleToggleSelect(todo.id)}
+                >
+                  <span className="icon">
+                    <i
+                      className={`far ${isSelected ? 'fa-eye-slash' : 'fa-eye'}`}
```
781calmstate: `classnames`

src/App.tsx
```diff
@@ -1,34 +1,128 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
+import { getTodos } from './api';
+import type { Todo } from './types/Todo';
+
+export const FILTERS = ['all', 'completed', 'active'] as const;
+type Filter = typeof FILTERS[number];
```
781calmstate: You have different declaration of you `Filter` type in your solution even though they are identical.

consider moving it into separated file and reuse.

src/App.tsx
```diff
@@ -1,34 +1,128 @@
...
+
 export const App: React.FC = () => {
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+  const [selectedId, setSelectedId] = useState<number | null>(null);
+  const [filter, setFilter] = useState<Filter>('all');
```
781calmstate: should be value from enum

src/App.tsx
```diff
@@ -1,34 +1,128 @@
...
+
+  useEffect(() => {
+    void handleLoadTodos();
+  }, []);
+
+  const normalizedQuery = query.trim().toLowerCase();
+
+  const matchesStatus = (t: Todo, current: Filter) => {
+    if (current === 'all') return true;
+    if (current === 'completed') return t.completed;
```
781calmstate: should be value from enum

src/App.tsx
```diff
@@ -1,34 +1,128 @@
...
+    } catch (e) {
+      const message = e instanceof Error ? e.message : 'Failed to load todos';
+      setError(message);
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    void handleLoadTodos();
```
781calmstate: why do you need the `void` here?

src/App.tsx
```diff
@@ -1,34 +1,128 @@
...
+  const normalizedQuery = query.trim().toLowerCase();
+
+  const matchesStatus = (t: Todo, current: Filter) => {
+    if (current === 'all') return true;
+    if (current === 'completed') return t.completed;
+    return !t.completed;
+  };
+
+  const matchesQuery = (t: Todo, q: string) =>
+    t.title.toLowerCase().includes(q);
```
781calmstate: this is a helper function.

the functions should be named so it's clear they are functions, the name should represent some kind of action. you named it like a boolean value

781calmstate (changes requested):
Good commits' names!

src/components/TodoList/TodoList.tsx
781calmstate: still classnames

781calmstate (changes requested):
You didn't fix all of the comments

mkyloa (approved):
fantastic!

PR #3173, opened 2025-10-22
add task solution

Julia-Hertseva:
[DEMO LINK](https://Julia-Hertseva.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,79 @@
...
+  const [status, setStatus] = useState<Status>('all');
+  const [user, setUser] = useState<User | null>(null);
+  const [selectedTodo, setSelectedTodo] = useState<Todo | null>(null);
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos().then(receivedTodos => {
+      setTodos(receivedTodos);
+      setIsLoading(false);
```
DmytroLymar: It's better to move it to the `finally` block
`getTodos()
      .then(setTodos )
      .finally(() => {
          setIsLoading(false);
      });
 `

src/App.tsx
```diff
@@ -1,14 +1,79 @@
...
+  };
+
+  const handleTodoSelect = (todo: Todo) => {
+    setIsUserLoading(true);
+    setSelectedTodo(todo);
+
+    getUser(todo.userId).then(receivedUser => {
+      setUser(receivedUser);
+      setIsUserLoading(false);
+    });
```
DmytroLymar: It is better to move the user loading logic to a modal window.

src/App.tsx
```diff
@@ -17,18 +82,35 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                status={status}
+                query={query}
+                onChangeQuery={onChangeQuery}
+                onClearQuery={onClearQuery}
```
DmytroLymar, line 97: you could use onChange Query inside Todo Filter

src/App.tsx
```diff
@@ -17,18 +82,35 @@ export const App: React.FC = () => {
...
+                onChangeStatus={onChangeStatus}
+              />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              <Loader isLoading={isLoading} />
```
DmytroLymar, line 103: It was possible not to change the Loader logic, but to use the condition rendering

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,39 +1,63 @@
...
+}) => {
   return (
     <div className="modal is-active" data-cy="modal">
-      <div className="modal-background" />
+      <div className="modal-background" onClick={onClose} />
 
-      {true ? (
-        <Loader />
+      {isLoading ? (
+        <Loader isLoading={isLoading} />
```
DmytroLymar, line 24: Again about changing the Loader component.
Here, both the component and conditional rendering work double duty

DmytroLymar (approved):
Overall, a working solution, but there are some points that AI might explain during verification.

PR #3270, opened 2026-01-23
added solution

Demtar23:
[DEMO LINK](https://Demtar23.github.io/react_dynamic-list-of-todos/)


Br0No (approved):
Good job, you can use lifting state up for input, but it's not necessary

PR #3282, opened 2026-01-24
add task solution

SergiyKusiy:
[DEMO LINK](https://SergiyKusiy.github.io/react_dynamic-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+      return;
+    }
+
+    setIsUserLoading(true);
+    setUser(null);
+
+    getUser(selectedTodo.userId)
+      .then(setUser)
+      .finally(() => setIsUserLoading(false));
+  }, [selectedTodo]);
```
de-vasta, line 44: It’s better to put it inside **TodoModal.tsx**, as was introduced in this morning’s class.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,29 @@
-export const TodoFilter = () => (
-  <form className="field has-addons">
+import React from 'react';
+type Status = 'all' | 'active' | 'completed';
```
de-vasta, line 2: we can export it and use whenever we need to (e.g. App.tsx [line 22](https://github.com/mate-academy/react_dynamic-list-of-todos/pull/3282/changes/12d7a034194a3bcb3a83b16af7b6020baa7f6d48#diff-26ad4b834941d9b19ebf9db8082bd202aaf72ea0ddea85f5a8a0cb3c729cc6f2R22))

src/App.tsx
```diff
@@ -1,14 +1,60 @@
...
+
+  const visibleTodos = todos.filter(todo => {
+    if (status === 'active' && todo.completed) {
+      return false;
+    }
+
+    if (status === 'completed' && !todo.completed) {
+      return false;
+    }
```
de-vasta, line 53: if you want to stick with that approach, why not to combine them into one?

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,29 @@
...
+
+type Props = {
+  status: Status;
+  query: string;
+  onStatusChange: (value: 'all' | 'active' | 'completed') => void;
```
de-vasta, line 7: that's a DRY violation. It's more convenient and easier to maintain our code when it's not hardcoded. Type `Status` isn't used even though it's defined in the same file

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -16,15 +37,24 @@ export const TodoFilter = () => (
...
-        <button data-cy="clearSearchButton" type="button" className="delete" />
-      </span>
+      {query && (
+        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="clearSearchButton"
+            type="button"
+            className="delete"
+            onClick={onClearQuery}
```
de-vasta, line 54: I guess, there's a little reason to pass additional callback which uses `setQuery` anyway, but makes our props even more bloated

src/components/TodoList/TodoList.tsx
```diff
@@ -1,6 +1,18 @@
...
+type Props = {
+  todos: Todo[];
+  onSelectTodo: (todo: Todo) => void;
+  selectedTodoId: number | null;
+};
+
+export const TodoList: React.FC<Props> = ({
+  todos,
+  onSelectTodo,
+  selectedTodoId,
```
de-vasta, line 14: We could just get `id` from `selectedTodo` without enlarging our props

de-vasta (approved):
Well, good job! Generally, I appove 👍🏼

de-vasta (approved):

PR #3426, opened 2026-05-15
add task solution

DmitriyS2k:
[DEMO LINK](https://DmitriyS2k.github.io/react_dynamic-list-of-todos/)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,26 @@
-export const TodoFilter = () => (
+type Props = {
+  searchQuery: string;
+  onSearchChange: (newQuery: string) => void;
+  selectedFilter: 'all' | 'completed' | 'active';
+  onFilterChange: (newFilter: 'all' | 'completed' | 'active') => void;
```
linska: you could extract the filter statuses into a separate type (e.g. FilterStatus) to avoid duplicating 'all' | 'completed' | 'active' in multiple places.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,12 +1,32 @@
...
+      .finally(() => setIsLoading(false));
+  }, [todo.userId]);
 
-export const TodoModal: React.FC = () => {
   return (
     <div className="modal is-active" data-cy="modal">
       <div className="modal-background" />
 
-      {true ? (
+      {isLoading || !user ? (
```
linska: What if the todo exists, but the related user is not found? In this case, `!user` will keep showing the loader even though loading has already finished. It might be better to handle loading and “user not found” as separate states.

linska (approved):
Nice work 👍

src/components/TodoModal/TodoModal.tsx
DmitriyS2k: Thanks. I updated the logic to handle the “user not found” state separately from loading.

src/components/TodoFilter/TodoFilter.tsx
DmitriyS2k: Done

PR #3428, opened 2026-05-15
Sol

sergeisvidlo4:
[DEMO](https://sergeisvidlo4.github.io/react_dynamic-list-of-todos/)


whomngmnt (approved):
everything works well!!!

src/App.tsx
```diff
@@ -8,7 +8,40 @@ import { TodoFilter } from './components/TodoFilter';
...
+
+  let visibleTodos = todos;
+
+  if (status === 'completed') {
+    visibleTodos = todos.filter(todo => todo.completed);
+  }
+
+  if (status === 'active') {
+    visibleTodos = todos.filter(todo => !todo.completed);
+  }
```
dothearts, line 29: You don't use ts in this case. 
All status need to be typified or you can use ENUM in this case.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,8 +1,26 @@
-export const TodoFilter = () => (
+import React from 'react';
+
+type Props = {
+  status: string;
+  onStatusChange: (status: string) => void;
```
dothearts, line 5: Need add types in all cases where you use statuses. 
Create separate type or ENUM for it.

dothearts (changes requested):
Many thanks for the solution, please check the comments and fix the status types.

PR #3502, opened 2026-07-24
Develop

aleks-pashkevych:


etojeDenys (changes requested):
1. after closing the todo, the icon still looks like the todo is open


Image

2. make sure that the search is not case-sensitive

Image

3. also try to pass all the tests

etojeDenys (approved):
lgtm

